PR #10, opened 2015-11-03
scatterplot: scale dot radius according to zoom level

karenpeng:
Hi @vicapow , in the scatterplot example, the radius of dots remain the same regardless of map zoom level. 
I find it useful to scale it so as to keep pace with the map, especially if the radius represents some value, for example doing data visualization.

What do you think?

before
before

after
after

### DIFF
--- a/src/overlays/scatterplot.react.js
+++ b/src/overlays/scatterplot.react.js
@@ -40,7 +40,9 @@ var ScatterplotOverlay = React.createClass({
     globalOpacity: React.PropTypes.number,
     dotRadius: React.PropTypes.number,
     dotFill: React.PropTypes.string,
-    compositeOperation: React.PropTypes.oneOf(COMPOSITE_TYPES)
+    compositeOperation: React.PropTypes.oneOf(COMPOSITE_TYPES),
+    defaultZoom: React.PropTypes.number,
+    zoom: React.PropTypes.number
   },
 
   getDefaultProps: function getDefaultProps() {
@@ -70,7 +72,8 @@ var ScatterplotOverlay = React.createClass({
     var canvas = this.getDOMNode();
     var ctx = canvas.getContext('2d');
     var props = this.props;
-    var radius = this.props.dotRadius;
+    var scale = Math.pow(this.props.zoom / this.props.defaultZoom, 6);
+    var radius = this.props.dotRadius * scale;
     var fill = this.props.dotFill;
     ctx.save();
     ctx.scale(pixelRatio, pixelRatio);
--- a/src/overlays/scatterplot.react.js
+++ b/src/overlays/scatterplot.react.js
@@ -73,6 +73,9 @@ var ScatterplotOverlay = React.createClass({
     var ctx = canvas.getContext('2d');
     var props = this.props;
     var scale = Math.pow(this.props.zoom / this.props.defaultZoom, 6);
+    //dot doesn't scale according to the map zoom level:
+    //var radius = this.props.dotRadius;    
+    //dot scales according to zoom level:
     var radius = this.props.dotRadius * scale;
     var fill = this.props.dotFill;
     ctx.save();
--- a/example/examples/scatterplot.react.js
+++ b/example/examples/scatterplot.react.js
@@ -85,7 +85,9 @@ var ScatterplotOverlayExample = React.createClass({
         locations: locations,
         dotRadius: 2,
         globalOpacity: 1,
-        compositeOperation: 'screen'
+        compositeOperation: 'screen',
+        defaultZoom: 11,
+        zoom: this.state.zoom
       })
     ]);
   }


vicapow:
Thanks for the PR!

1. Can I ask what the use case is? I think this would probably be useful but would like to make sure there's at least one concrete use case before adding it.

2. I'm noticing that the relative size of the circles doesn't seem to be maintained.

scatter-plot-scale

We probably also want to make sure the size of the circles is maintained dependent on the latitude. Since the project is web mercator: http://blogs.esri.com/esri/arcgis/2011/03/24/tissot-s-indicatrix-helps-illustrate-map-projection-distortion/ this should be possible.

Would also be nice to specify the radius of the circles in meters or kilometers instead of just "zoom". That also eliminates the need for `defaultZoom`.

karenpeng:
1. I realize the importance of dots keeping pace with the map when doing a data visualization project. In the project, the radius of dots represents the amount of bikes, and it varies over time. So from the map, users could see the trend of the amount changes. However if the users zoom out, they lose sight of the trend because it's too crowded, and if they zoom in, the amount appeals to be much less because the radius feels less.
I believe this applies to other visualizations as long as the radius represents some amount on the map. The radius scaling should be an option in this API.

2. Yeah the scaling in the code now is not scientifically correct, the pow number 6 is totally up for adjustment. I agree with you that if we could make solid ratio connection between radius and length unit for example latitude of the map, the experience will be much better. Could you reference some possible mapbox-gl API to look at?

vicapow:
> I realize the importance of dots keeping pace with the map when doing a data visualization project. In the project, the radius of dots represents the amount of bikes, and it varies over time. So from the map, users could see the trend of the amount changes. However if the users zoom out, they lose sight of the trend because it's too crowded, and if they zoom in, the amount appeals to be much less because the radius feels less.
I believe this applies to other visualizations as long as the radius represents some amount on the map. It'll be nice if the radius scaling is an option in this API.

That sounds like a solid use case. Thanks for sharing!

> Yeah the scaling in the code now is not scientifically correct, the pow number 6 is totally up for adjustment. I agree with you that if we could make solid ratio connection between radius and length unit, and maybe also considering latitude, the experience will be much better. Could you reference some possible mapbox-gl API to look into?

I wanna take back what I said earlier about scaling with regard to latitude. For most use cases (yours included!) we probably want to keep the circles constant with respect to the tile coordinate system instead of keeping them constant with respect to the projected geographical distance. Most data overlay use cases I can think of tie together non-geographical data to geography so it shouldn't matter if the circles stay the same size over greenland as they do Ecuador.

I'm still undecided about expanding the api of the scatterplot overlay (and other overlays for that matter). I'd most rather see this implemented as a separate component in a separate module that wraps the scatterplot component or just as a separate utility function. If it helps in the meantime, here's a diff that adds [geometric zooming](http://bl.ocks.org/mbostock/3680999) to the scatterplot example.

```diff
diff --git a/example/examples/scatterplot.react.js b/example/examples/scatterplot.react.js
index 2ccea93..766ffe6 100644
--- a/example/examples/scatterplot.react.js
+++ b/example/examples/scatterplot.react.js
@@ -28,6 +28,12 @@ var d3 = require('d3');
 var MapGL = require('../../src/index.js');
 var ScatterplotOverlay = require('../../src/overlays/scatterplot.react');
 
+function scaleRelativeToZoom(radius, relativeToZoom, zoom) {
+  return radius * Math.pow(2, zoom) / Math.pow(2, relativeToZoom);
+}
+
+var INITIAL_ZOOM = 11;
+
 // San Francisco
 var location = require('./../data/cities.json')[0];
 
@@ -54,7 +60,7 @@ var ScatterplotOverlayExample = React.createClass({
     return {
       latitude: location.latitude,
       longitude: location.longitude,
-      zoom: 11,
+      zoom: INITIAL_ZOOM,
       startDragLatLng: null,
       isDragging: false
     };
@@ -83,7 +89,7 @@ var ScatterplotOverlayExample = React.createClass({
     }, this.props), [
       r(ScatterplotOverlay, {
         locations: locations,
-        dotRadius: 2,
+        dotRadius: scaleRelativeToZoom(1, INITIAL_ZOOM, this.state.zoom),
         globalOpacity: 1,
         compositeOperation: 'screen'
       })
```

map-gl-geometric-zoom

karenpeng:
The experience is great! Curious about the math. Thanks for sharing the geometric zooming example :smile:

Just some thoughts when dealing with radius:

1.
What if we want fine control with radius for each dot? 
It happens in dataviz project when radius of each dot represents some value for that location specificly.

That requires a binding of radius value with location. I think if this api grows to be more extensible, more abstraction on the data->view side should be done.
Maybe we could provide a structure for each dot as:
```js
var obj = {
  loc: [37.7749, -122.4194],
  value: 7
}
```
And for all the dots, we could have an array of obj, or an object of obj ( O(1) to access a specific dot )

2.
keep the circles constant with respect to the tile coordinate system?
You mentioned some possible ways to do it, I'm more apt to making it a separate util functionality. 

For one thing, it's a general function that might be used by other examples.
For another, separating it enables us to treat it as an option, to scale/not scale according to different projects's needs.

vicapow:
> The experience is great! Curious about the math. Thanks for sharing the geometric zooming example :smile:

We can calculate the size of the tile map in pixels using `2 ^ mapScale * sizeOfASingleTile`. (mapbox-gl uses a [web mercator project](https://en.wikipedia.org/wiki/Web_Mercator) and works just like a [slippy map](https://en.wikipedia.org/wiki/Tiled_web_map).) If we want our circles to always have the same radius, we can compute the scale factor by dividing the two map sizes:

```
scaleFactor = (2 ^ currentZoom * sizeOfASingleTile) / (2 ^ relativeToZoom * sizeOfASingleTile)
```

which reduces to:

```
scaleFactor = 2 ^ currentZoom / 2 ^ relativeToZoom
```

> What if we want fine control with radius for each dot? It happens in dataviz project when radius of each dot represents some value for that location specifically.

Good point. I like the idea of using d3 style accessors. This makes it so the component isn't concerned with what particular property names you use but would have common defaults.

````html
<Overlay locations={locations} radius={function(location) { return location.radius; } />
````

> You mentioned some possible ways to do it, I'm more apt to making it a separate util functionality.

I'm leaning that way, too. Speaking more philosophically, I like the unix mentality of ["do one thing well"](http://blog.izs.me/post/48281998870/unix-philosophy-and-nodejs). I'd like to eventually remove the overlays completely and have them be entirely separate modules. Same for any utility functions. I've begun to do this with the [heatmap overlay](https://github.com/vicapow/react-map-gl-heatmap-overlay).

playing-with-heatmaps

And I've also created a template module to make it easy to fork, create and publish new overlays: https://github.com/vicapow/react-map-gl-example-overlay

I'll just link to any overlays others create in the readme

**Comment by starhillUber** [comment]
You can potentially save a bit of computation by further simplifying 

<code>scaleFactor = 2 ^ currentZoom / 2 ^ relativeToZoom</code>

to

<code>scaleFactor = 2 ^ (currentZoom - relativeToZoom)</code>

Depending on the implementation details of the language, it can be a lot more efficient to implement the exponent by a bitshift operation.

**Comment by vicapow** [comment]
@starhillUber oh, good call!

File: src/overlays/scatterplot.react.js [hunk 1]
```diff
@@ -25,6 +25,11 @@ var d3 = require('d3');
 var r = require('r-dom');
 var Immutable = require('immutable');
 var COMPOSITE_TYPES = require('canvas-composite-types');
+var scaleRelativeToZoom = require('./../utils').scaleRelativeToZoom;
+
+function scaleRelativeToZoom(radius, relativeToZoom, zoom) {
```
contra: duplicated?

abmai:
@vicapow @karenpeng is this still valid?

contra:
@abmai Don't think so, all possible now with the min/max options + the meter radius setting.

abmai:
Closing since this looks to be outdated.

PR #15, opened 2015-11-09
Move images from repository

vinkla:
This way we make the repository smaller to download and the file structure looks a lot cleaner.

Maybe we should also look at the [files](https://docs.npmjs.com/files/package.json#files) attributes to make the package size even smaller. I guess we only need the src folder?


vicapow:
I think it's safe to assume most people will be using `npm install` to download and install the project. How about we just add `react-map-gl-screenshots.png` to the `.npmignore` file? Might also make sense to move the images to an `img` folder.

Thanks for the PR!

vinkla:
Thanks for getting back! In my opinion images shouldn't exists within repositories since they don't make any difference on the code. In this case we want to showcase examples of what this package can do and that is where GitHub's CDN would be great. This is of course totally up to you, I'm just sharing my thoughts :)

In my opinion we should add the `files` attribute instead since we only want the `src` folder anyway. The `.npmignore` will be one more file to keep track of.

vicapow:
I'm not sure what you mean `files attribute`. Can you elaborate?
Github seems to do a fair job of supporting images in diffs. Other than keeping the repo file size small, are there any other arguments to keeping images outside of the repo? I kind of like using the git to version control the screenshots to keep track of their changes.

vicapow:
I maturely submitted that last comment. let me try that again:

I'm not sure how github's CDN works. How would we go about updating the screenshots?

I'm not sure what you mean `files attribute`. Can you elaborate?

Github seems to do a fair job of supporting images in diffs. Other than keeping the repo file size small, are there any other arguments to keeping images outside of the repo? I kind of like using the git to version control the screenshots to keep track of their changes.

vinkla:
[Here is an article on utilising GitHub's CDN.](https://vinkla.com/posts/upload-images-github/) You basically create an empty issue and the drag n drop the images onto it. Then copy the link GitHub will provide and then just leave the page (not saving the issue). 

My main argument is just to make it look less cluttered. Since GitHub will automatically upload all images within the repository to their CDN anyway. You will also still be able to keep track of what is changing in the images since the urls will still exist within the history.

The `files` attributes basically is telling NPM which files and folder that it actually needs to download. 

> The "files" field is an array of files to include in your project. If you name a folder in the array, then it will also include the files inside that folder. (Unless they would be ignored by another rule.)
https://docs.npmjs.com/files/package.json#files

It works in the as the `.npmignore` file but the other way around. It includes only the necessary files.

vinkla:
Any update on this?

vicapow:
Hey @vinkla, sounds gtm! merged. :) 

Feel free to submit a separate diff for npm `files` attribute, if you'd like

vinkla:
Thanks!

PR #16, opened 2015-11-11
Rebase to react-map-gl 0.3.0 and pin mapbox-gl to 0.11.1

gnavvy:


vicapow:
We need to hold off on this just yet. Still some work that needs to happen in some of our downstream dependencies.

PR #24, opened 2015-11-17
Add a radius to mouse move callback

baileyid:
Adding a radius parameter in mouse move callback so that it supports point features.


vicapow:
:+1:

PR #27, opened 2015-11-24
Remove ReactDOM.findDOMNode calls from src

drewbo:
Per the react docs, this use case is discouraged and can be
replaced with the ref itself. It also allows the removal of
ReactDOM from the src require statements

https://facebook.github.io/react/docs/top-level-api.html#reactdom.finddomnode

### DIFF
--- a/src/map-interactions.react.js
+++ b/src/map-interactions.react.js
@@ -20,7 +20,6 @@
 'use strict';
 
 var React = require('react');
-var ReactDOM = require('react-dom');
 var MapboxGL = require('mapbox-gl');
 var Point = MapboxGL.Point;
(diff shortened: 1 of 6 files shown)


vicapow:
Awesome. Thanks, @drewbo!

drewbo:
Np, great project @vicapow. I think this also allows the removal of `ReactDOM` from the `peerDependencies` (it's only used in the example) but I wasn't positive

vicapow:
Oh, good catch! Added in: https://github.com/uber/react-map-gl/commit/3021cd697d081d4adac0f1e219399801f6f7b5b8

PR #33, opened 2015-12-13
remove the debounce on MapGL._resizeMap

wwwtyro:


src/map.react.js
```diff
@@ -282,10 +282,10 @@ var MapGL = React.createClass({
     }
   },
 
-  _resizeMap: debounce(function _resizeMap() {
```
vicapow: sweet! If this is the only use of `debounce`, can we remove it as a dep?

wwwtyro:
@vicapow It was, thanks! :+1:

vicapow:
awesome! lgtm

PR #37, opened 2015-12-15
rebuild gh-pages bundle

wwwtyro:
Whoops! I'll try to remember this, next time.


vicapow:
it's probably better that it's in a separate diff, anyway.

PR #44, opened 2016-01-18
Update budo to latest version and get example working again

mattdesl:
- Ran `npm install budo@latest --save-dev` (this adds a bit of commit noise since it re-orders deps)
- Removed `bistre` since budo includes its own built-in pretty-printer which tends to look nicer
- Removed the `require('process')` line since it seems to break with latest browserify; should instead just assume it exists globally like in Node.

Budo has lots of improvements/fixes since v4, you can see them all [here](https://github.com/mattdesl/budo/blob/master/CHANGELOG.md). Cheers! :tada:


vicapow:
Woo! Thanks @mattdesl. This is awesome. 👍

PR #45, opened 2016-01-23
Feature/efficient updating of geojson sources

Que3216:
Currently GeoJSON sources get updated by removing the old source, and adding the new source: but this is many times slower than calling setData. This PR updates it to use setData for GeoJSON sources (as long as they haven't changed type), and the old method for all other sources.

### DIFF
--- a/src/map.react.js
+++ b/src/map.react.js
@@ -63,7 +63,7 @@ function updateSource(map, update) {
 
     if (newSource.type === "geojson") {
         var oldSource = map.getSource(update.id);
-        if (oldSource.type === "geojson") {
+        if (oldSource.setData != null) {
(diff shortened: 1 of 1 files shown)


Que3216:
I've tested it locally, and it works.

vicapow:
Hey @Que3216, thanks for the PR and apologies for the delay in getting back to you.

I like the idea here but I think you need to also account for the case where a prop other than `data` has changed for the source. A geoJSON source can have other props like `cluster` or `clusterRadius`. see: https://www.mapbox.com/mapbox-gl-js/api/#GeoJSONSource

Apercu:
Closing in favor of #124.

PR #48, opened 2016-02-04
Fix bug in batched source/layer updates

AndyMoreland:
_setDiffStyle called map.batch but didn't use the "style_batch" argument
that it receives in order to batch updates to layers and sources. This
caused one "_broadcastLayers" call in mapbox-gl per changed source or layer,
which is very slow in some cases.

Example usage from mapbox-gl docs: https://www.mapbox.com/mapbox-gl-js/api/#Map.batch

### DIFF
--- a/src/map.react.js
+++ b/src/map.react.js
@@ -414,29 +414,29 @@ var MapGL = React.createClass({
       return;
     }
 
-    map.batch(function batchStyleUpdates() {
+    map.batch(function batchStyleUpdates(batch) {
       sourcesDiff.enter.forEach(function each(enter) {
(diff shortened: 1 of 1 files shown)


vicapow:
@AndyMoreland apologies for not getting back to you sooner. This PR looks good! Before I can merge, could you sign our Contributor License Agreement (CLA)? You can find the [link here](https://docs.google.com/a/uber.com/forms/d/1pAwS_-dA1KhPlfxzYLBqK6rsSWwRwH95OCCZrcsY5rk/viewform).

AndyMoreland:
I signed it. Thanks!

PR #54, opened 2016-02-25
Add require example directly in readme

mdamien:
makes the copy paste easier

also, looks like you need an API token, maybe should put it in the readme/doc too ?

### DIFF
--- a/README.md
+++ b/README.md
@@ -22,6 +22,8 @@ npm install react-map-gl --save
 ### Usage
 
 ````js
+var MapGL = require('react-map-gl');
+
 <MapGL width={400} height={400} latitude={37.7577} longitude={-122.4376}
   zoom={8} onChangeViewport={(viewport) => {
     var {latitude, longitude, zoom} = viewport;


vicapow:
Hey @mdamien, thanks for the PR! Before I can merge, could you sign our Contributor License Agreement (CLA)? You can find the [link here](https://docs.google.com/a/uber.com/forms/d/1pAwS_-dA1KhPlfxzYLBqK6rsSWwRwH95OCCZrcsY5rk/viewform). You only need to sign it once for all Uber projects.

I think the README already mentioned the Mapbox API token (for map styles that use Mapbox's API's). See: https://github.com/uber/react-map-gl/blame/master/README.md#L110 Do you think we should go into more detail or call it out more?

mdamien:
For the API token, it's just required my mapbox gl by default: https://github.com/mapbox/mapbox-gl-js/blob/b7032636206448e16eb681836f97e79f0587fd7e/js/util/config.js

Enforced here: https://github.com/mapbox/mapbox-gl-js/blob/master/js/util/mapbox.js#L9

But you make the props no required to indicate that it's only for custom styles: https://github.com/uber/react-map-gl/blob/master/src/map.react.js#L105

So, for now, mounting the component without a token throws an exception.

Maybe have another PR to make it required or something.

Signed the CLA, should be ok.

vicapow:
@mdamien I see your point. It is a little confusing that we use a Mapbox style by default but don't require the access token from within MapGL. With that said, there are lots of use cases where a user doesn't need to have a Mapbox API token since they're not using Mapbox's services. One example of this is when using Stamen's base tiles: https://github.com/vicapow/react-map-gl-heatmap-overlay/blob/master/example/main.js#L15. In the case where we're using a map style that relies on Mapbox services, I think we should continue to rely on MapboxGL-js to throw an error.

mdamien:
ok, looks good to me :smile_cat:

PR #55, opened 2016-03-02
Free up resource on unmount

chrisirhc:


vicapow:
Thanks for the PR, @chrisirhc!

Issue #34, opened 2015-12-13
Investigate using individual d3 4.0 modules


vicapow:
instead of the d3 3.0 kitchen sink module.

samhogg:
@vicapow are you looking for help here?

Thought using the entirety of d3 was overkill and then came across this issue. Happy to jump in with a PR.

vicapow:
Hi @samhogg! A PR that tackles this would be much appreciated.

ibgreen:
This is implemented on latest master

PR #57, opened 2016-03-09
Upgrade D3 to version 4

samhogg:
Here's an initial cut at #34

* Used d3 modules where appropriate
* Added round function to utils, d3@4 doesn't have it yet
* TODO - find a replacement for d3.geo in choropleth.react.js
  This might be coming soon - although https://github.com/d3/d3-geo
  is currently empty.
* TODO - use alphaify from `npm` rather than my fork [(relies on this PR being merged)](https://github.com/vicapow/alphaify/pull/1)

Obviously the d3.geo part is a blocker at the moment, but the other changes are pretty straightforward.

### DIFF
--- a/example/examples/custom.react.js
+++ b/example/examples/custom.react.js
@@ -21,7 +21,8 @@
 
 var assign = require('object-assign');
 var React = require('react');
-var d3 = require('d3');
+var d3Array = require('d3-array');
+var d3Random = require('d3-random');
 var window = require('global/window');
 var alphaify = require('alphaify');
 var transform = require('svg-transform');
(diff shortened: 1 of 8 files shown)


vicapow:
This is lookin great, @samhogg. I guess we should wait on the geo stuff before landing this

mcnuttandrew:
Looks like the geo stuff is up and running, i think the commented out lines can be replaced with with

`      const transform = geoTransform({point: projectPoint});
      const path = geoPath().projection(transform).context(ctx);`

samhogg:
@vicapow - I have some bandwidth to have another go at this later in the week if that's helpful

vicapow:
@samhogg that would be awesome! Might need to rebase from master. I think there's been a few changes

samhogg:
@vicapow yeah the d3 module dependencies have moved significantly too. Will have a shot and get back to you later in the week.

ibgreen:
This has been resolved separately, see latest master. Apologies for not using this PR, this became part of a larger dependency cleanup effort.

PR #58, opened 2016-03-14
expose a prop to allow onHoverFeatures callback to be called when the…

abmai:
…re are no features under x,y

let me know if you want the opposite approach instead (and variable naming).

### DIFF
--- a/src/map.react.js
+++ b/src/map.react.js
@@ -182,7 +182,7 @@ var MapGL = React.createClass({
       mapboxApiAccessToken: config.DEFAULTS.MAPBOX_API_ACCESS_TOKEN,
       preserveDrawingBuffer: false,
       attributionControl: true,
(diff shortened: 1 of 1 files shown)


src/map.react.js
```diff
@@ -174,7 +181,8 @@ var MapGL = React.createClass({
       onChangeViewport: null,
       mapboxApiAccessToken: config.DEFAULTS.MAPBOX_API_ACCESS_TOKEN,
       preserveDrawingBuffer: false,
-      attributionControl: true
+      attributionControl: true,
+      ignoreEmptyFeatures: false
```
vicapow: should be `true`, right?

abmai: yes, good catch. seems like I might be dyslexic, thanks!

PR #59, opened 2016-03-16
Cleaner exposition of the library

alex3165:
Multiple things :
- No more node_modules in react-map-gl library
- Don't expose the source code directly but compile it into one file
- Clean unnecessary dependencies

It require the user to install mapbox-gl as a peer dependency, so that the fix for the shaders file work when using mapbox-gl with webpack.
see : https://github.com/alex3165/react-map-gl-poc

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "name": "react-map-gl",
   "description": "A React wrapper for MapboxGL-js and overlay API.",
-  "version": "0.6.5",
+  "version": "0.6.6",
   "main": "dist/react-mapbox-gl.js",
   "keywords": [
     "mapbox",
@@ -18,6 +18,9 @@
   "devDependencies": {
     "bowser": "^1.0.0",
     "global": "^4.3.0",
+    "keymirror": "^0.1.1",
+    "mapbox-gl": "0.15.0",
+    "nop": "^1.0.0",
     "object-assign": "^4.0.1",
     "svg-transform": "0.0.3",
     "viewport-mercator-project": "^2.0.1",


vicapow:
Hey @alex3165, thanks for the PR! I like what you did here. I am concerned this will break the API for existing consumers. I can't decide if it's worth the change now or if we should group this in with a future version bump. @ibgreen or @abmai, any thoughts?

alex3165:
Hey @vicapow It will indeed break the API when it come to import stuff but the change is worth i think so if peoples use es2015 they can use destruct import like this :
```
import { MapGL, ChoroplethOverlay } from "react-map-gl";
```

And another important point is that it allow peoples that use Webpack to use the recommended fix when it come to import shaders : https://mikewilliamson.wordpress.com/2016/02/24/using-mapbox-gl-and-webpack-together/

We can maybe add it clearly in the readme file so peoples are aware of this change.

kiRach:
:+1: for this PR

ibgreen:
> @ibgreen or @abmai, any thoughts?
@vicapov my 2 cents

- Adding webpack support is fine, however I am uncertain about having the package.json main entry pointing to a webpack generated bundle. We are a lot of folks using react-map-gl via browserify and AFAIK we have no experience with this setup. So, as a first step, we could still generate the webpack bundle, but leave package.json `main` unchanged and have webpack folks consume the new bundle with 
```
import MapGL from 'react-map-gl/dist/webpack';
```

- A minor thing: adding all the layers to the main index.js is fine, although I'd personally prefer for MapGL to be the default export, enabling:
```
import MapGL, {SVGLayer, CanvasLayer, ...} from 'react-map-gl';
```
Since this is an ES5 module, It would require attaching the layers as "static" props on the MapGL constructor but that's not too bad IMHO. Just a thought...

joewood:
Isn't it more usual practice to provide a webpack entry point to the source rather than a prebuilt distribution? This would also maintain browserify compatibility.

alex3165:
@ibgreen I am not sure but the webpack bundled file is i think compatible with browserify and for your second point, i did `exports.default = require('./map.react');` so we can do something like that :
```
import MapGL, {SVGLayer, CanvasLayer, ...} from 'react-map-gl';
```

@joewood Do you have an example of the settings ?

joewood:
Sorry, I didn't mean directly to the source, I meant to the non-bundled JavaScript. That *could* be the actual source if a transpiler is not being used, but more commonly an output **lib** directory that Babel output is written to. The most common practice is to build separate JS files and point the **package.json** `main` to the main entry point.

Example - [React Big Calendar](https://github.com/intljusticemission/react-big-calendar/blob/master/package.json#L8) points to **lib/index.js**. This is not a bundled distribution it's just one of 20 or so built JS files which are used alongside the dependencies in **node_modules**.

alex3165:
Ah ok i see, in this case we don't need to compile files with babel we would just need to move sources files from `src` to `dist` and yes everybody would be happy :)

Apercu:
Thanks @alex3165 and sorry for the late follow-up.

I will close your PR in favor of https://github.com/uber/react-map-gl/commit/8ae5e20c7854e25940b53c1abc326998611b1953 as it resolves most of the issues you raised at the time. Also, you should never push a builded version on git.

PR #66, opened 2016-03-31
Fix onClickFeatures to use the new `map.queryRenderedFeatures`.

vicapow:
cc: @shortsleeves

### DIFF
--- a/src/map.react.js
+++ b/src/map.react.js
@@ -509,17 +509,11 @@ var MapGL = React.createClass({
     if (!this.props.onHoverFeatures) {
       return;
     }
-    map.featuresAt([pos.x, pos.y], {
-      radius: 1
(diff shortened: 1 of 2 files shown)


src/map.react.js
```diff
@@ -540,17 +534,13 @@ var MapGL = React.createClass({
...
-    }, function callback(error, features) {
-      if (error) {
-        throw error;
-      }
-      if (!features.length) {
-        return;
-      }
-      this.props.onClickFeatures(features);
-    }.bind(this));
+    var size = 5;
```
shortsleeves: 5 is going to be a little bit less than 15 (box extent of 10). Was 15 too large a value?

shortsleeves:
:+1:

PR #67, opened 2016-04-01
Fix tests on Travis-CI by using testron.

vicapow:


ibgreen:
@AshleyTravers @vicapow 

I've integrated this into dev (1.0.0) 

- it still has issues (not all tests pass on travis, perhaps due to lack of mapbox access token?) 
- and testron gives a positive return code even when tests fail.

But at least the tests cases are run and results are logged on travis, and we don't get broken builds for valid changes.

PR #69, opened 2016-04-11
Initial ES6 port

ibgreen:
Examples run. Initial build scripts in place. 

Keeping on es6 branch since additional testing and script polish needed.


ibgreen:
> there may be a few other places were `opts` is used. We could replace these with spread syntax. ie., `({fill, stroke})`.
Done

vicapow:
@ibgreen looking good so far. What do you think about adding `dist/` to `.gitignore` ? I think it would cut down on the diff noise here and in the future.

Qix-:
:+1: on ignoring `dist/` - they'll cause desynchronization and noisy diffs otherwise since they're generated.

ibgreen:
> What do you think about adding dist/ to .gitignore
@vicapow @Qix - Yes good idea. 
- added `dist` to `.gitgnore`
-  added `.npmignore` file to override `.gitignore` for npm package publishing (which uses gitignore by default when selecting which files to publish). 
- Can't be completely sure until we publish a version but it should work.

All examples work per my testing. 
I cannot get unit tests to work (npm issues?), but pushed a new commit anyway per discussion with @vicapow.

example/examples/custom.react.js
```diff
@@ -66,85 +69,97 @@ var OverlayExample = React.createClass({
...
+      <MapGL
+        { ...viewport }
+        onChangeViewport={ this._onChangeViewport }>
+
+        <CanvasOverlay
+          { ...viewport }
+          redraw={ this._redrawCanvasOverlay }/>
+
+        {
+          /* We use invisible SVG elements to support interactivity. */
```
vicapow: can you move this comment back to the relevant code?

example/examples/route.react.js
```diff
@@ -53,70 +56,77 @@ var RouteOverlayExample = React.createClass({
...
-    var ctx = opt.ctx;
-    var width = opt.width;
-    var height = opt.height;
+      }
+      </g>
+    );
+  }
+
+  @autobind
+  _redrawCanvasOverlay(opt) {
```
vicapow: good use of `({ctx, width, height})` instead of `(opt)`

example/examples/style-diffing.react.js
```diff
@@ -17,61 +17,61 @@
...
+};
 
-  displayName: 'StyleDiffingExample',
+export default class StyleDiffingExample extends Component {
 
-  PropTypes: {
-    width: React.PropTypes.number.isRequired,
-    height: React.PropTypes.number.isRequired
-  },
+  static get propTypes() {
```
vicapow: Is there any benefit to `static get propTypes() {` over...

````js
StyleDiffingExample.propTypes = {...};
````
If not, the latter seems more inline with the React docs.

File: src/config.js [hunk 4]
```diff
@@ -17,22 +17,19 @@
...
   '';
 
-var config = {
+export default {
   DEFAULTS: {},
   MAPBOX_API_ACCESS_TOKEN: null,
   CURSOR: {
-    GRABBING: prefix + 'grabbing',
-    GRAB: prefix + 'grab'
+    GRABBING: `${PREFIX} grabbing`,
```
vicapow: You've got an extra space here. should be `${PREFIX}grabbing`. ditto for the next line. otherwise, the "grabber" cursor wont be present.

src/map-interactions.react.js
```diff
@@ -187,36 +195,39 @@ var MapInteractions = React.createClass({
...
     // Scale by sigmoid of scroll wheel delta.
-    var scale = 2 / (1 + Math.exp(-Math.abs(delta / 100)));
+    let scale = 2 / (1 + Math.exp(-Math.abs(delta / 100)));
     if (delta < 0 && scale !== 0) {
       scale = 1 / scale;
     }
-    this.props.onZoom({pos: pos, scale: scale});
+    this.props.onZoom({pos, scale});
     window.clearTimeout(this._zoomEndTimeout);
     this._zoomEndTimeout = window.setTimeout(function _setTimeout() {
```
vicapow: the closure can be `() => this.props.onZoomEnd()`

src/map.react.js
```diff
@@ -531,22 +506,22 @@ var MapGL = React.createClass({
...
     }
     this.props.onClickFeatures(features);
-  },
+  }
 
-  _onZoom: function _onZoom(opt) {
-    var map = this._getMap();
-    var transform = cloneTransform(map.transform);
-    var around = unprojectFromTransform(transform, opt.pos);
+  @autobind _onZoom(opt) {
```
vicapow: `_onZoom({scale, pos})`

src/map.react.js
```diff
@@ -555,81 +530,112 @@ var MapGL = React.createClass({
...
     this._onChangeViewport({isDragging: false});
-  },
+  }
 
-  render: function render() {
-    var props = this.props;
-    var style = assign({}, props.style, {
+  render() {
+    const {className} = this.props;
+    const props = this.props;
```
vicapow:
````js
const {props} = this;
const {className} = props;
````

File: test/index.js [hunk 8]
```diff
@@ -1,11 +1,12 @@
-'use strict';
+require('babel/register');
```
vicapow: I think this is missing from `package.json`

vicapow:
@ibgreen almost there! I left some comments inline. Can you also squash these commits into one to clear the history with the `dist` commits?

example/examples/style-diffing.react.js
ibgreen: > Is there any benefit to `static get propTypes() {` over...`StyleDiffingExample.propTypes = {...};`
I feel it is important for readability that the public api (i.e. the props) comes first, just like in an ES5 react component. Whether to compensate for the lack of static properties in ES6 with a static getter at the beginning of the class, or an assignment after the end is a matter of taste. I dropped the static accessors per your suggestion, but left the PROP_TYPES constants at the top, and did a `StyleDiffingExample.propTypes = PROP_TYPES;` at the bottom. Does that seem reasonable?

src/map-interactions.react.js
ibgreen: > the closure can be () => this.props.onZoomEnd()

I've had problems when using window.setTimeout with fat arrow functions in the past, and since they are asynchronous they can be a little tricky to test. So I intentionally left these out for now. Easy enough to update them when everything else is confirmed to be working.

ibgreen:
> almost there! I left some comments inline. Can you also squash these commits into one to clear the history with the dist commits?
@vicapow - Squashed to single commit, dist removed. I believe that I have either resolved or responded to all your comments, let me know if I missed something. If not, it might be good time to pull this into your es6 branch so that you can do some testing on your side, and I can open new PRs as needed.

example/examples/style-diffing.react.js
chrisirhc: 👍  to `PROP_TYPES`

vicapow:
@ibgreen I was about to merge but then noticed some errors while testing. Specifically, I don't think this will work for non-babel consumers. I'd like to still support ES5 consumers but if others disagree, I'm happy to move forward with this.

Specifically, looking at dist/index.js, I see...

````js
Object.defineProperty(exports, "__esModule", {
  value: true
});

var _map = require('./map.react');

Object.defineProperty(exports, 'default', {
  enumerable: true,
  get: function get() {
    return _interopRequireDefault(_map).default;
  }
});

function _interopRequireDefault(obj) { return obj && obj.__esModule ? obj : { default: obj }; }
````

Which for commonJS / es5 consumers exposes the following object when requiring.

````js
var MapGL = require('../');
console.log(MapGL); // { __esModule: true, get default: function() { ... }, __proto__: ...}
````

This is causing the tests to fail for me. I know I originally said to forget the tests, but now I think it's important to make sure the tests are running for you before this lands, to make sure the `require` in ES5 works properly. We can sync up offline.

> If not, it might be good time to pull this into your es6 branch so that you can do some testing on your side, and I can open new PRs as needed.

Sure. Feel free to merge into the es6 branch for now. If you do, can you create a new issue for the above?

ibgreen:
@vicapow - Top level file uses now module.exports. I will merge this into main repo.

PR #71, opened 2016-04-16
Include bounds in onChangeViewport

Willyham:
@vicapow

### DIFF
--- a/src/map.react.js
+++ b/src/map.react.js
@@ -244,7 +244,8 @@ var MapGL = React.createClass({
       longitude: center.lng,
       zoom: map.getZoom(),
       isDragging: this.props.isDragging,
-      startDragLngLat: this.state.startDragLngLat
+      startDragLngLat: this.state.startDragLngLat,
+      bounds: map.getBounds()
     }, _changes);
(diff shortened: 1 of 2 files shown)


Willyham:
Tests are passing locally, I think that there's an issue with running the tests on Travis that you're already aware of. If you want to land that first I'll rebase.

vicapow:
Hi @Willyham. Thanks for the PR. This should be documented better but this is already supported with the `viewport-mercator-project` module. see: https://github.com/uber-common/viewport-mercator-project

````js
> var ViewportMercator = require('viewport-mercator-project');
> var mercator = ViewportMercator(viewport);
> [mercator.unproject([0, 0]), mercator.unproject([viewport.width, viewport.height])]
[ [ -0.20599365234374462,
    0.17166112013818108 ],
  [ 0.20599365234374462,
    -0.17166112013818108 ] ]
````

Maybe a good first step would be to better explain the relationship between these two modules in the docs for react-map-gl. Since the bounds can be derived from the rest of the viewport state, if we were going to support this in react-map-gl, I think it should be in the form of a utility method.

Willyham:
Ah, thanks! Makes sense to have a separate module, wasn't aware of that.

Issue #21, opened 2015-11-13
Building in webpack throws error


KyleAMathews:
```
ERROR in ./~/glify/index.js
Module not found: Error: Cannot resolve module 'fs' in /root/RelateRocket/webapp/node_modules/glify
 @ ./~/glify/index.js 5:9-22
```

KyleAMathews:
The answer apparently is to add some configuration to my webpack config file. It'd be nice if this was documented.

gnavvy:
Known issue: https://github.com/mapbox/mapbox-gl-js/issues/1649

shanecav:
@KyleAMathews What configuration did you add to webpack to make it work? I'm having the same issue.

The fix suggested at mapbox/mapbox-gl-js#1649 (adding `node: { fs: "empty" }` to my webpack config) allowed it to be bundled without errors, but when I view the app in browser I get "Uncaught TypeError: fs.readFileSync is not a function".

KyleAMathews:
Shane, I didn't ever get it to work actually :-)

imgettingdeported:
has anyone been able to build successfully with webpack?

lafin:
I was able.
Need add to webpack config
```
{
  test: /\.json$/,
  loader: 'json-loader'
},
{
  test: /\.js$/,
  include: path.resolve(__dirname, 'node_modules/mapbox-gl/js/render/shaders.js'),
  loader: 'transform/cacheable?brfs'
},
{
  test: /\.js$/,
  include: path.resolve(__dirname, 'node_modules/webworkify/index.js'),
  loader: 'worker'
}
```
and add to package.json 
```
"brfs": "^1.4.3",
"json-loader": "^0.5.4",
"transform-loader": "^0.2.3",
 "worker-loader": "^0.7.0"
```

yuvadm:
@lafin do you have a complete example for that? I've added the config, but I'm still stuck with the `fs.readFileSync` problem.

lafin:
@yuvadm you can see example https://github.com/lafin/frontend-boilerplate all changes https://github.com/lafin/frontend-boilerplate/commit/c75a3f43b0aa322fddce9a8cee90e758db8b8b04 (react-map-gl + heatmap overlay)
For heatmap use forked webgl-heatmap because pull request not apply yet https://github.com/vicapow/webgl-heatmap/pull/1

yuvadm:
[...]

PR #73, opened 2016-04-27
Add props for bearing and pitch

jwarning:
Just a simple addition of optional props to set the bearing and pitch in the viewport.

### DIFF
--- a/src/map.react.js
+++ b/src/map.react.js
@@ -172,7 +172,17 @@ var MapGL = React.createClass({
       * There are still known issues with style diffing. As a temporary stopgap,
       * add the option to prevent style diffing.
       */
-    preventStyleDiffing: React.PropTypes.bool
+    preventStyleDiffing: React.PropTypes.bool,
+
+    /**
+      * Specify the bearing of the viewport
+      */
+    bearing: React.PropTypes.number,
+
+    /**
+      * Specify the pitch of the viewport
+      */
+    pitch: React.PropTypes.number
   },
 
   getDefaultProps: function getDefaultProps() {
@@ -293,8 +303,8 @@ var MapGL = React.createClass({
       this._getMap().jumpTo({
         center: [state.longitude, state.latitude],
         zoom: state.zoom,
-        bearing: 0,
-        pitch: 0
+        bearing: this.props.bearing || 0,
+        pitch: this.props.pitch || 0
       });
     }
     if (state.width !== state.prevWidth || state.height !== state.prevHeight) {


ibgreen:
@jwarning 
This is one of three pieces I am looking at for supporting "tilted view".
1) Adding support for these props and forwarding them to mapbox-gl (exactly as you have done in this PR!)
2) Add support for 3D navigation (event handling) and add the pitch/bearing props to the `onChangeViewport` callback. Haven't made final decision but perhaps a variant of `map-interactions.react.js`?
3) Calculate a projection transformation matrix from bearing and pitch (that matches the transformation matrix mapbox-gl calculates) and supply it to `onChangeViewport`, so that overlays can apply the same 3D transform as MapBoxGL does. 
About 3) I am not sure how practical setting 3D transforms is for the overlays that are included in react-map-gl. My use case is the WebGL based map overlays in [deck.gl](https://github.com/uber/deck.gl)

@vicapow - This covers #1 nicely. Happy to pull this.

ibgreen:
@jwarning 
* For completeness, it would be nice to have a mention of these in the docs, maybe a note towards the end of README. 
* Also I think we will want to add an example that showcases these. It would become more elaborate as we do 2 and 3 on the list, but you could get it started, e.g. just hard coding bearing and pitch to non-zero constants for now.
Let me know what you think?

vicapow:
I don't recommend merging these changes now as this will break how the viewport transformation code works during interaction when they're present.

jwarning:
Fair enough. I realise that there are a couple other things which need to be done to make this more complete. Happy to potentially help on getting that sorted if need be.

ibgreen:
>Fair enough. I realise that there are a couple other things which need to be done to make this more complete. Happy to potentially help on getting that sorted if need be.

@jwarning 
If you'd like to help you could try installing version 1.0.0-beta-5 (or pull the es6 branch) and let me know if any issues.
- MapGL now takes `pitch` and `bearing` props (per your PR)
- You can now enable tilt controls with `perspectiveMode={true}`. (using Command-drag)
- There is a new example with tilt so do `npm start` first after you pull and try Command-drag in the top example.
- In your own app, you do need to make sure that you pass the entire `mapState` from `onUpdateViewport` back into the component (there are new perspective related props in addition to `lat/lon/zoom/isDragging`)

jwarning:
I'm having some issue compiling the new version for use in my project. I'm using a couple hacks to get it to build with webpack, however the new build seems to have new issues to work around. I'll try and see if I can figure a way to get it working though.

ibgreen:
> Supporting webpack/browserify properly might be something worth investigating, as most people will consume this library with a tool like that.

@jwarning 
Thanks for the update. While I've never used webpack, I am consuming this internally in a browserify based production environment and I can confirm that it works well with browserify.

Would be interested to hear more about the `gl-matrix` issue. My first thought is that since this is an indirect dependency (`mapbox-gl-js` requires it), if webpack has issues with that it should probably be reported directly in either the `gl-matrix` or `mapbox-gl-js` repos, not sure `react-map-gl` can do anything except change the version of its `mapbox-gl-js` dependency if there is a version that fixes it.

Otherwise, I'd have thought that as long as webpack was directed to `react-map-gl`'s `dist/index.js` it would simply walk the require tree from there. Since all of `react-map-gl`'s ES6 code has been pre-transpiled to ES5 in the package.json prepublish hook, it should pose no problems for a mature JS bundler like webpack.

jwarning:
Even despite using the transpiled files in dist/index.js, there are a couple other issues which arise from including some of react-map-gl's dependencies. The main one seems to be that the `shaders.js` file in `mapbox-gl` calls `'fs'` which cannot be included into a browser bundle. One can get around this with certain webpack loaders. See some of the solutions in [https://github.com/uber/react-map-gl/issues/21](https://github.com/uber/react-map-gl/issues/21).

PR #75, opened 2016-04-28
Es6

ibgreen:
- Perspective mode support
- Light refactoring of map react component to reduce state.


src/map-interactions.react.js
```diff
@@ -210,6 +223,7 @@ export default class MapInteractions extends Component {
         ref="container"
         onMouseMove={ this._onMouseMove }
         onMouseDown={ this._onMouseDown }
+        onContextMenu={ this._onMouseDown }
```
chrisirhc: Does this mean that right clicking now triggers the mousedown?

PR #76, opened 2016-05-02
Upgrade to mapbox-18

vicapow:
cc: @ibgreen

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 node_modules
 npm-debug.log
 coverage
+dist
--- a/package.json
+++ b/package.json
@@ -18,7 +18,6 @@
   "dependencies": {
     "alphaify": "^1.0.0",
     "bowser": "^1.0.0",
-    "budo": "^8.2.2",
     "canvas-composite-types": "^1.0.4",
     "d3": "^3.5.16",
     "global": "^4.3.0",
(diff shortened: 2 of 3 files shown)


vicapow:
Hey @ibgreen what do you think about merging this?

ibgreen:
> Hey @ibgreen what do you think about merging this?

@vicapow Absolutely. I just realized that I have made essentially the same changes in dev / 1.0.0-beta releases (i.e. bumped to mapbox 0.18 and dropped the style batching) and that has been working well in apps for some time now. 

Will rebase dev when this lands.

vicapow:
> @vicapow Absolutely. I just realized that I have made essentially the same changes in dev / 1.0.0-beta releases (i.e. bumped to mapbox 0.18 and dropped the style batching) and that has been working well in apps for some time now.
Cool but why drop style batching? Without it updates to the map style that only involve changing map data / geoJSON will cause the underlying basemap styles to reload unnecessarily.

In the meantime, can we land this now?

ibgreen:
> Cool but why drop style batching? Without it updates to the map style that only involve changing map data / geoJSON will cause the underlying basemap styles to reload unnecessarily.

> We are probably talking about different things. All I meant is that `map.batch` is no longer supported in the mapbox 0.18 API. Your diff drops "batching" [too](https://github.com/uber/react-map-gl/pull/76/files#diff-837ea5eff7048b52938b3703eb97b692L431), and my change is identical to yours.

> In the meantime, can we land this now?

Again yes, please go ahead and land and publish. Or ping me separately if you would like me to do it.

PR #77, opened 2016-05-02
Check is style loading

vicapow:
Wait for the current style to load before trying to apply the next one to prevent `Style not done loading` errors.


vicapow:
I think there's a bug with this diff in the case when the map gets several style updates and looses reference to the correct previous style. Will look into this more when I get the time but for now, don't merge.

PR #84, opened 2016-05-17
Fix overlay export

nrabinowitz:
Copy/pasta error broke all of the overlay exports.


ibgreen:
@nrabinowitz Great catch - we should also update the examples to do a proper import of these, as that would have caught the problem.

ibgreen:
@nrabinowitz Fixed examples, plus a mapbox-gl v 0.18 issue and published as 1.0.0-beta-10

PR #88, opened 2016-06-07
Advice for using with redux

Willyham:
@ibgreen @vicapow


vicapow:
I think this is sweet!

Issue #80, opened 2016-05-04
Use HTTPS on interactive docs


iamvdo:
There are mixed content errors in Interactive docs, because of HTTP AJAX calls on tiles providers.

> Mixed Content: The page at 'https://uber.github.io/react-map-gl/#/' was loaded over HTTPS, but requested an insecure XMLHttpRequest endpoint 'http://tile.stamen.com/toner/12/656/1582.png'. This request has been blocked; the content must be served over HTTPS.

winkler1:
Yes -- means the examples don't work. Really great docs & examples otherwise!

![image](https://cloud.githubusercontent.com/assets/883896/15290024/dfae9e64-1b43-11e6-82b0-3cc41bf452e2.png)

kachkaev:
This is still the case. When you are on https://uber.github.io/, not http://uber.github.io/, the tiles from `stamen.com` do not load. 

Looks like something is also wrong with the SSL cert for stamen maps:
https://tile.stamen.com/toner/12/653/1582.png

![stamen maps privacy error](https://cloud.githubusercontent.com/assets/608862/16655705/94b9b374-4453-11e6-85bc-1a3e70b15f17.png)

How about just switching the map style into something more reliable?

wandergis:
The url i think can change to `//tile.stamen.com/toner/z/x/y.png`

dnordberg:
+1 haven't been able to try the maps

Apercu:
Resolved

PR #106, opened 2016-08-08
Un-protical-ify tile.stamen.com to fix mixed content errors

nbergseng:
Fix for https://github.com/uber/react-map-gl/issues/80

Sorry the bundle change is massive... adding shrinkwrapping would likely prevent it in the future


Apercu:
Do you mind signing the CLA? I can fallback to the other one @alukach if it's not possible.

Apercu:
And if you can remove the bundle from the diff it would be nice too 😄

Apercu:
Closing in favor of #153.

PR #113, opened 2016-08-22
Introduce d3 v4 micro modules. Move utilities to utils directory

ibgreen:
The complete d3 module is a rather big dependency. Moving to d3 v4 allows us to require only the submodules we need which should help make app executables smaller.

@vp @mcnuttandrew @bulyonov


vicapow:
lgtm. in the future, I recommend separating out PRs do they only address one issue at a time. This makes it easier for review, track and revert, if necessary.

Issue #111, opened 2016-08-20
fitBounds returns a zoom level of NaN


esamelson:
Hello,

I am running into an issue in that `fitBounds` is returning an object whose `zoom` property is NaN. However, the `latitude` and `longitude` properties seem to be working correctly, as they are at a location where I would expect the center of the viewport to be. I have looked into the source code a bit, but I am unfamiliar with the Transform library so I am unsure as to where the issue might be. I don't think the issue is with my direct input to `fitBounds`, as the `latitude` and `longitude` are correct, but I could be wrong. Any ideas?

Thanks very much.

ibgreen:
@esamelson You could probably single step through the [fitBounds](https://github.com/uber/react-map-gl/blob/master/src/fit-bounds.js) code and see where the zoom calculations go wrong and the NaN appears, that would be helpful. If not, it would probably be easiest to debug this if we could write a test case for it.

Either you can open a PR with a new test case, or provide the failing input/outputs to the function so that we can write one.

If you want to go the PR route, please add a test case in test directory, something like:

```
import {fitBounds} from '../src';

test('fitBounds', t => {
  const result = fitBounds(<Your failing parameters go here>);
  t.equal(result.latitude, ..., 'got expected latitude');
  t.equal(result.longitude, ..., 'got expected latitude');
  t.equal(result.zoom, ..., 'got expected zoom');  
  t.end();
});
```

Note, you would need to sign our open source contributor form as well.

esamelson:
Well, I stepped through the fitBounds code and discovered the error was on my end after all. So sorry for wasting your time @ibgreen .

PR #114, opened 2016-08-22
Add fitBounds test

ibgreen:
Even though it turned out to be a false alarm, #111 was a good reminder that we need more unit tests. This PR adds an initial basic sanity test for fitBounds.

@vp @esamelson

### DIFF
--- a/src/utils/fit-bounds.js
+++ b/src/utils/fit-bounds.js
@@ -23,16 +23,29 @@
 import Transform from './transform';
 import {LngLatBounds, Point} from 'mapbox-gl';
 
-export default function fitBounds(width, height, _bounds, options) {
-  const bounds = new LngLatBounds([_bounds[0].reverse(), _bounds[1].reverse()]);
-  options = options || {};
-  const padding = typeof options.padding === 'undefined' ? 0 : options.padding;
+/**
+ * Returns map settings {latitude, longitude, zoom}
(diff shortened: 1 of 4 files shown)


src/fit-bounds.js
```diff
@@ -3,10 +3,13 @@
 import Transform from 'mapbox-gl/js/geo/transform';
 import {LngLatBounds, Point} from 'mapbox-gl';
 
-export default function fitBounds(width, height, _bounds, options) {
-  const bounds = new LngLatBounds([_bounds[0].reverse(), _bounds[1].reverse()]);
-  options = options || {};
-  const padding = typeof options.padding === 'undefined' ? 0 : options.padding;
+export default function fitBounds(width, height, _bounds, {
+  padding = 0
```
vicapow: not a big deal but I would have put this in a seperate PR since it changes the functionality of `fitBounds`.

vicapow:
👍

src/fit-bounds.js
ibgreen: @vicapow - Good catch - I assume you are referring to the dropping of the `reverse` calls. Yes that was not intentional. 

Since there was no JSDoc it took me a while to figure out that this function only works on arrays and not Mapbox.LngLats, which is what I initially assumed, and the reason I removed `reverse` (which only works on arrays).

I'll make a new PR to fix this.

PR #115, opened 2016-08-22
Add a stateful wrapper to react map gl

ibgreen:
An optional wrapper.

Minimizes the amount of callback handling and state-keeping required by the application.

Use case: optimizes rendering in some of our applications, especially during startup, where eliminating just a couple of rerenders (map + overlay stack) can make a noticeable difference.

@vicapow

### DIFF
--- a/src/stateful-map.react.js
+++ b/src/stateful-map.react.js
@@ -1,0 +1,117 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
+
+// Permission is hereby granted, free of charge, to any person obtaining a copy
+// of this software and associated documentation files (the "Software"), to deal
+// in the Software without restriction, including without limitation the rights
+// to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+// copies of the Software, and to permit persons to whom the Software is
+// furnished to do so, subject to the following conditions:
+
+// The above copyright notice and this permission notice shall be included in
+// all copies or substantial portions of the Software.
+
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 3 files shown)


src/stateful-map.react.js
```diff
@@ -0,0 +1,117 @@
...
+// THE SOFTWARE.
+import React, {PropTypes, Component} from 'react';
+import autobind from 'autobind-decorator';
+import pureRender from 'pure-render-decorator';
+import MapGL from './map.react';
+
+function noop() {}
+
+const PUBLIC_MAP_STATE_DEFAULTS = {
+  latitude: 0,
```
Apercu: Maybe defaults to a user-friendly location rather than in the Atlantic?

src/stateful-map.react.js
```diff
@@ -0,0 +1,117 @@
...
+  longitude: 0,
+  zoom: 11,
+  pitch: 0,
+  bearing: 0,
+  altitude: 1.5
+};
+
+function compareProps(refObject, newObject) {
```
Apercu: We should maybe create an util for this.

src/stateful-map.react.js
```diff
@@ -0,0 +1,117 @@
...
+    const publicMapState = this._getPublicMapState(mapState);
+    if (!compareProps(publicMapState, this.props)) {
+      this.props.onChangeViewport(publicMapState);
+    }
+  }
+
+  render() {
+    return (
+      <MapGL
+        { ...this.props }
```
Apercu: Maybe spread state first, this way the props can override these values easily.

abmai:
@ibgreen should we abandon this one since there's a pretty big refactor going into 2.0.0 with the interactions separation?

PR #117, opened 2016-08-23
Remove alphaify

ibgreen:
Peer dependency error from alphaify is causing issues for apps. 

Inlined code and removed dependency since - alphaify is just a two line function.


vicapow:
👍

PR #124, opened 2016-09-19
update data if no other GeoJSONSource options changed - prevents flicker

tsemerad:
Currently if I update a GeoJSON source in my style, the rendered data flickers every time it changes, and also the map freezes up a bit. This PR uses `setData` instead of `removeSource`, `addSource` when updating a GeoJSON feature whose other options haven't changed. This fixes the issue. This PR is the same as #45 except it handles the cases where other GeoJSON source options have changed.

### DIFF
--- a/src/map.react.js
+++ b/src/map.react.js
@@ -257,6 +257,36 @@ export default class MapGL extends Component {
     });
   }
 
+  _updateSource(map, update) {
+    const newSource = update.source.toJS();
+    if (newSource.type === 'geojson') {
+      const oldSource = map.getSource(update.id);
+      if (oldSource instanceof mapboxgl.GeoJSONSource) {
+        // update data if no other GeoJSONSource options were changed
+        if (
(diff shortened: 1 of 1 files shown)


cammanderson:
+1 for this one. This is a goody. This is one of most commonly reported through to us in testing from users that the flickr while modifying geojson data. :-)

ibgreen:
@tsemerad Thanks for the PR, changes look good. If you are willing to register as an [Uber open source contributor](https://docs.google.com/a/uber.com/forms/d/e/1FAIpQLSfu-DmpJD5wYYUOxY50noHF9BWJwDbV0d2W1w45oEIbIx72RQ/viewform), I will merge the PR and publish a new version.

tsemerad:
You're welcome. I just registered as an Uber open source contributor. Glad I could help out!

ibgreen:
Published as v1.6.0

PR #125, opened 2016-09-20
Handle unsupported browsers

Pessimistress:
Issue: in old browsers that do not support webGL, `new mapboxgl.Map()` returns `undefined` and results in errors in subsequent references to `this._map`.

- Exposes mapboxgl's `supported()` API 
- Do not creating a map instance unless mapbox is supported. 
- Disable interaction in unsupported browsers to prevent further errors.

### DIFF
--- a/src/map.react.js
+++ b/src/map.react.js
@@ -64,6 +64,11 @@ const PROP_TYPES = {
     */
   mapboxApiAccessToken: PropTypes.string,
   /**
+    * `onUnsupported` callback is fired when the component is mounted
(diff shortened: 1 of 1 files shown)


src/map.react.js
```diff
@@ -187,6 +193,13 @@ export default class MapGL extends Component {
   }
 
   componentDidMount() {
+    if (!this.state.isSupported) {
+      if (this.props.onUnsupported) {
```
ibgreen: Set the prop's default to "noop" and drop the if statement here.

PR #126, opened 2016-09-20
Change cursor to pointer when hovering over interactive features

Pessimistress:
- Allow changing the cursor style over clickable features
- Correct prefix for default cursor style in Chrome >= 28

### DIFF
--- a/src/config.js
+++ b/src/config.js
@@ -30,6 +30,7 @@ export default {
   MAPBOX_API_ACCESS_TOKEN: null,
   CURSOR: {
     GRABBING: `${PREFIX}grabbing`,
-    GRAB: `${PREFIX}grab`
+    GRAB: `${PREFIX}grab`,
+    POINTER: 'pointer'
   }
 };
--- a/src/config.js
+++ b/src/config.js
@@ -21,7 +21,7 @@
 import browser from 'bowser';
 
 const PREFIX =
-  browser.webkit ? '-webkit-' :
+  browser.webkit || browser.blink ? '-webkit-' :
   browser.gecko ? '-moz-' :
   '';
 
--- a/src/map.react.js
+++ b/src/map.react.js
@@ -562,7 +562,7 @@ export default class MapGL extends Component {
   }
 
   render() {
-    const {className, width, height, cursor, style} = this.props;
+    const {className, width, height, style} = this.props;
     const mapStyle = {
       ...style,
       width,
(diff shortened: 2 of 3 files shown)


src/config.js
```diff
@@ -21,7 +21,7 @@
 import browser from 'bowser';
 
 const PREFIX =
-  browser.webkit ? '-webkit-' :
+  browser.webkit || browser.blink ? '-webkit-' :
```
ibgreen: What is browser.blink? A comment perhaps?

src/map.react.js
```diff
@@ -78,6 +78,10 @@ const PROP_TYPES = {
     */
   height: PropTypes.number.isRequired,
   /**
+    * Override the default cursor.
+    */
+  cursor: PropTypes.string,
```
ibgreen: If possible, add default value in DEFAULT_PROPS. Looking at the cursor method, don't we need two cursor props (GRAB/GRABBING)?

src/map.react.js
```diff
@@ -559,12 +563,12 @@ export default class MapGL extends Component {
...
 
   render() {
-    const {className, width, height, style} = this.props;
+    const {className, width, height, cursor, style} = this.props;
     const mapStyle = {
       ...style,
       width,
       height,
-      cursor: this._cursor()
+      cursor: cursor || this._cursor()
```
ibgreen: Can this be moved to DEFAULT_PROPS?

src/config.js
Pessimistress: blink is the fork of webkit used in newer Chrome. The webkit flag is off for Chrome >= 28. https://github.com/ded/bowser#rendering-engine-flags

chrisirhc: `.blink` was added in bowser 1.2.0, so you might want to raise the package.json dependency for it.

Ref:
https://github.com/ded/bowser/releases/tag/1.2.0

ibgreen: Thanks, my point was it would be nice to add this link as a comment in the source code.
```
// See https://github.com/ded/bowser#rendering-engine-flags
```

jordymeow:
Hello guys,

Sorry but is this working? If yes, how? I tried to do it with a ref, but it doesn't work:
`mapRef.current.getCanvas().style.cursor = 'pointer';`

Also, I really wonder why there are no mentions to this anywhere for this library; did I miss something or nobody wants that pointer to change when something is actionable?

PR #128, opened 2016-09-21
fixed GeoJSONSource updating, broken when upgrading to mapbox-gl v0.24.0

tsemerad:
The change applied in #124 has broken when updating to the latest Mapbox GL JS. I just tried out v1.6.0, and immediately noticed that it throws an error when updating GeoJSON sources. The cause of this is that when upgrading from [Mapbox GL JS v0.21.0 to v0.22.0](https://github.com/mapbox/mapbox-gl-js/blob/master/CHANGELOG.md#0220-august-11-2016), they made `GeoJSONSource` private, along with some other refactoring, so we can no longer check for an `instanceof` it. I have it fixed now against Mapbox GL JS v0.24.0.

### DIFF
--- a/src/map.react.js
+++ b/src/map.react.js
@@ -276,21 +276,22 @@ export default class MapGL extends Component {
     const newSource = update.source.toJS();
     if (newSource.type === 'geojson') {
       const oldSource = map.getSource(update.id);
-      if (oldSource instanceof mapboxgl.GeoJSONSource) {
+      if (oldSource.type === 'geojson') {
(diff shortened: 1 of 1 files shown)


ibgreen:
Merged, we'll have to fix the linter warning in next diff.

ibgreen:
Published as 1.6.1, thanks for the quick catch.

tsemerad:
You're welcome. Sorry about that semicolon. I just fixed it in a PR (#129).

PR #131, opened 2016-09-27
Provide a way to control per-layer interactivity

Pessimistress:
https://www.mapbox.com/mapbox-gl-js/api/#Map#queryRenderedFeatures

ReactMapGL was calling mapbox's queryRenderedFeatures API without passing any filtering parameters. Features on all layers are returned as a result.

Without asking the user to provide a separate prop, we may only query layers whose interactive property is set to true, as suggested by the documentation for **onHoverFeatures** and **onClickFeatures**. In the case where no layers have this property specified (e.g. using Mapbox's default styles), we fall back to querying all layers. 

@vicapow @ibgreen

### DIFF
--- a/src/map.react.js
+++ b/src/map.react.js
@@ -93,11 +93,16 @@ const PROP_TYPES = {
     */
   startDragLngLat: PropTypes.array,
   /**
-    * Called when a feature is hovered over. Features must set the
-    * `interactive` property to `true` for this to work properly. see the
-    * Mapbox example: https://www.mapbox.com/mapbox-gl-js/example/featuresat/
-    * The first argument of the callback will be the array of feature the
-    * mouse is over. This is the same response returned from `featuresAt`.
+    * Called when a feature is hovered over. Uses Mapbox's
+    * queryRenderedFeatures API to find features under the pointer:
+    * https://www.mapbox.com/mapbox-gl-js/api/#Map#queryRenderedFeatures
+    * To query only some of the layers, set the `interactive` property in the
+    * layer style to `true`. See Mapbox's style spec
+    * https://www.mapbox.com/mapbox-gl-style-spec/#layer-interactive
+    * If no interactive layers are found (e.g. using Mapbox's default styles),
+    * will fall back to query all layers.
+    * @callback
+    * @param {array} features - The array of features the mouse is over.
     */
   onHoverFeatures: PropTypes.func,
   /**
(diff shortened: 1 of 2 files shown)


src/map.react.js
```diff
@@ -303,6 +305,20 @@ export default class MapGL extends Component {
     map.addSource(update.id, newSource);
   }
 
+  // Hover and click only query layers whose interactive property is true
+  // If no interactivity is specified, query all layers
+  _updateQueryParams(mapStyle) {
+    if (mapStyle instanceof Immutable.Map) {
```
chrisirhc: Might want to use `Map#isMap` since instanceof checks can be slow unless you specifically want to support classes that inherit from Immutable.Map.

http://facebook.github.io/immutable-js/docs/#/Map/isMap

vicapow:
I would wait for @ibgreen to review but looks 👍 to me.

src/map.react.js
```diff
@@ -303,6 +305,20 @@ export default class MapGL extends Component {
...
+      mapStyle = mapStyle.toJS();
+    }
+
+    const interactiveLayerIds = Array.isArray(mapStyle.layers) ? mapStyle.layers
+      .filter(l => l.interactive)
+      .map(l => l.id) : [];
+    this._queryParams = interactiveLayerIds.length === 0 ? {} :
```
ibgreen: What happens if all layers are non-interactive? Doesn't this drop the filter so they all become interactive?

src/map.react.js
```diff
@@ -303,6 +305,20 @@ export default class MapGL extends Component {
     map.addSource(update.id, newSource);
   }
 
+  // Hover and click only query layers whose interactive property is true
+  // If no interactivity is specified, query all layers
+  _updateQueryParams(mapStyle) {
+    if (mapStyle instanceof Immutable.Map) {
+      mapStyle = mapStyle.toJS();
```
ibgreen: Getting a little worried that as we add style manipulating features, we risk converting the entire mapStyle blob `toJS()` more than once. That would be the big performance hit here. Can we just extract the layers and convert just those?

src/map.react.js
```diff
@@ -303,6 +305,20 @@ export default class MapGL extends Component {
     map.addSource(update.id, newSource);
   }
 
+  // Hover and click only query layers whose interactive property is true
```
ibgreen: I think more documentation here would be very helpful for casual maintainers/readers. 
1) You could turn this into a full JSDoc section.
2) You could add explanatory text by rewriting the great summary you did for this PR. Maybe add the link to the API doc

```
https://www.mapbox.com/mapbox-gl-js/api/#Map#queryRenderedFeatures
```

Pessimistress: Mapbox's documentation for layer.interactive: 
https://www.mapbox.com/mapbox-gl-style-spec/#layer-interactive

src/map.react.js
Pessimistress: We only perform the query if props.onHoverFeatures or props.onClickFeatures is defined. So maybe it's not unreasonable to assume the user wants something to be interactive...?

PR #133, opened 2016-10-05
Fix bug where onClickFeatures is fired after dragging

Pessimistress:
MapGL was checking for clicked features upon the onMouseUp event. When the user releases the mouse after panning or rotating, it should not be considered a click.

- Adding events onMouseClick and onTouchTap to MapInteractions
- Move firing of onClickFeatures in MapGL to handlers for onMouseClick and onTouchTap

### DIFF
--- a/src/map-interactions.react.js
+++ b/src/map-interactions.react.js
@@ -64,10 +64,12 @@ const PROP_TYPES = {
   onMouseRotate: PropTypes.func,
   onMouseUp: PropTypes.func,
   onMouseMove: PropTypes.func,
+  onMouseClick: PropTypes.func,
   onTouchStart: PropTypes.func,
   onTouchDrag: PropTypes.func,
   onTouchRotate: PropTypes.func,
   onTouchEnd: PropTypes.func,
+  onTouchTap: PropTypes.func,
   onZoom: PropTypes.func,
   onZoomEnd: PropTypes.func
 };
@@ -78,10 +80,12 @@ const DEFAULT_PROPS = {
   onMouseRotate: noop,
   onMouseUp: noop,
   onMouseMove: noop,
+  onMouseClick: noop,
   onTouchStart: noop,
   onTouchDrag: noop,
   onTouchRotate: noop,
   onTouchEnd: noop,
+  onTouchTap: noop,
   onZoom: noop,
   onZoomEnd: noop
 };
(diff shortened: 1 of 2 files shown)


src/map-interactions.react.js
```diff
@@ -167,6 +174,9 @@ export default class MapInteractions extends Component {
     const pos = this._getMousePos(event);
     this.setState({pos});
     this.props.onMouseUp({pos});
+    if (!this.state.dragged) {
+      this.props.onMouseClick({pos});
```
ibgreen: Does it make sense to set `dragged` to false here?

src/map-interactions.react.js
```diff
@@ -176,6 +186,9 @@ export default class MapInteractions extends Component {
     const pos = this._getTouchPos(event);
     this.setState({pos});
     this.props.onTouchEnd({pos});
+    if (!this.state.dragged) {
```
ibgreen: Set dragged to false?

Pessimistress:
@ibgreen the `dragged` state is internal to MapInteraction, instead of the map itself, for the sole purpose of determining when click/tap events should be fired. MapGL sets `isDragging` to true as soon as the mouse is down, so it will not help in this case.

src/map-interactions.react.js
vicapow: is it not possible to read this from the `prop.isDragging` prop?

Pessimistress: `isDragging` is set to true on mouse down, so it will not help telling whether it's been a click on mouse up. I don't want to change this behavior, in case it breaks other things.

src/map-interactions.react.js
Pessimistress: We're not monitoring onMouseLeave, so there's a chance onMouseUp doesn't get hit when the mouse button is released. It's safer to reset it onMouseDown. Unless we want to do it in both places for clarity.

src/map-interactions.react.js
```diff
@@ -138,7 +145,7 @@ export default class MapInteractions extends Component {
   @autobind
   _onMouseDrag(event) {
     const pos = this._getMousePos(event);
-    this.setState({pos});
+    this.setState({pos, dragged: true});
```
vicapow: Since this is a flag to indicate if the mouse moved while being "down", how about `didDrag`?

vicapow:
minus the minor nick pick on the name, I'm 👍 on this.

PR #136, opened 2016-10-13
update mapbox-gl to v0.25.1

stepankuzmin:
Hi there! Mapbox GL v0.25.1 out on September 30 (https://github.com/mapbox/mapbox-gl-js/blob/master/CHANGELOG.md), all tests are passed, so maybe we can update to this version?

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "d3-scale": "^1.0.3",
     "d3-selection": "^1.0.2",
     "global": "^4.3.0",
-    "mapbox-gl": "0.24.0",
+    "mapbox-gl": "0.25.1",
     "pure-render-decorator": "^1.1.0",
     "svg-transform": "0.0.3",
     "viewport-mercator-project": "^2.0.1"


cammanderson:
I think we have some issues merging this directly. Version 0.25 dropped support for webpack due to the ongoing webworkify issues. You now directly read in the dist files. This wouldn't be an issue for people using the mapbox API directly, but the utils/transform module in react-map-gl needs access to the mapbox GP class not exposed using the dist.

stepankuzmin:
I see. So there is no way to use mapbox-gl 0.25 with react-gl-map if mapbox-gl is not exposing Transform class?

cammanderson:
You could potentially do some trick with webpack I guess; alias the transform class import to the source? I'm a little stuck myself at the moment.

cammanderson:
I think I managed to do it by running 1.6.0 react map gl, and setting up my webpack to call in the webpack dist, with another alias to the transform file directly. Works and compiles, testing across chrome etc. Yet to confirm if IE11 etc run..

contra:
Should be closed, lib already updated to 0.26.0 https://github.com/uber/react-map-gl/blob/master/package.json#L29

Apercu:
Thanks, closing as @contra mentioned.

PR #140, opened 2016-10-22
Adds `onClick` prop handler

forrert:
This PR adds a new prop to the component that allows to subscribe to click events on the map. The handler is called with the map coordinates (longitude / latitude) as well as with the screen coordinates (x / y).

A new example demonstrating the `onClick` property as well as the existing `onClickFeatures` was added.

### DIFF
--- a/example/examples/click.react.js
+++ b/example/examples/click.react.js
@@ -56,10 +56,15 @@ export default class ClickExample extends Component {
 
   @autobind
   _onClickFeatures(features) {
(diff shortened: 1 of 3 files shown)


toabey:
+1

tsemerad:
This is great! I was just about to make a PR for this myself - I have a fork with the exact same new `onClick` prop written nearly the same way. +1 for this.

abmai:
#49

PR #145, opened 2016-10-28
update webpack configuration documentation

davidhampgonsalves:
Update readme to reflect that Mapbox-gl-js has [dropped webpack support](https://github.com/mapbox/mapbox-gl-js/#using-mapbox-gl-js-with-other-module-systems). As such the official mapbox webpack config file linked to no longer exists.

The mention of historical issues might still have value to some users but have gotten pretty hairy (in my experience) and no longer reflect the mapbox-gl-js official approch.  As such I removed mention of them and believe the RFH represents the current best practice.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,12 +19,11 @@ See the interactive docs at: https://uber.github.io/react-map-gl
 npm install react-map-gl --save
 ```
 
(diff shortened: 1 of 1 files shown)


Apercu:
Thanks, but closing in favor of #155.

PR #146, opened 2016-10-29
use command | shift | ctrl | alt to enable perspective mode

### LINKED ISSUE #103: Custom modifier keys in map interactions
It seems that meta key is hard-coded as a trigger for perspective changing inside `<MapInteractions>`. It would be nice if this key could be changed to `ctrl` or `alt` via a prop of `<MapGL>`, for instance. Current behaviour triggers windows start menu time and again when a user attempts to see their data from an unusual angle :–)

Or is it possible to replace metaKey with another one already? I could not figure this out from the source code.

### LINKED ISSUE #138: Modifier keys for map interaction on Linux non-functional
While getting a demo of the new deck.gl example site discovered that the modifier keys for map interaction do not seem to work on the Linux version of Chrome. Keys tested: Control, Alt, Shift, combinations of all. @ibgreen let me know this appears to be a problem upstream in `react-map-gl`.  Checked the open issues and didn't find anything regarding this issue.  So here it is, totally broke. ;)

Environment:
OS: Linux 4.4.x
Desktop: Gnome 3.18
User-Agent: `Mozilla/5.0 (X11; Linux x86_64) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/53.0.2785.143 Safari/537.36`

Let me know what debug info you need. Cheers!

gnavvy:
Fixes #138 Fixes #103.
Users can now use any one of the function key {command, ctrl, shift, alt} + drag for the perspective control.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
-# Version 1.7.1
+# Version 1.7.2
 * Use any one of the function keys {command, shift, ctrl, alt} to enable the
   perspective mode.
+
+# Version 1.7.1
+* Bump Mapbox version to 0.26
 
 # Version 1.7.0
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+# Version 1.7.1
+* Use any one of the function keys {command, shift, ctrl, alt} to enable the
+  perspective mode.
+
 # Version 1.7.0
 
 * Provide a way to control per-layer interactivity - onClickFeatures and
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,7 @@
+# Version 1.7.1
```
ibgreen: There is already a 1.7.1, so this will have to be 1.7.2.

for 1.7.1, I think @Pessimistress bumped mapbox version but it didn't make it to the changelog, maybe you can add it when you update this PR?

gnavvy: will do.

PR #148, opened 2016-10-30
make geojson res more efficient

KrishMunot:


src/map.react.js
```diff
@@ -234,6 +234,21 @@ export default class MapGL extends Component {
     this._callOnChangeViewport(map.transform);
     this._updateQueryParams(mapStyle);
   }
+  
+  function updateSource(map, update) {
```
Apercu: You can't a class method like this.

src/map.react.js
```diff
@@ -234,6 +234,21 @@ export default class MapGL extends Component {
...
+     var newSource = update.source.toJS();
+ 
+     if (newSource.type === "geojson") {
+         var oldSource = map.getSource(update.id);
+         if (oldSource.setData != null) {
```
Apercu: Please use triple equality.

Issue #130, opened 2016-09-23
Post 1.6.0, fitBounds method returns a NaN zoom value


tangential:
This seems to possibly coincide with the MapboxGL 0.24 upgrade. Using fitBounds, and applying the returned values on to map props would cause the map component to fail.

gaving:
Having a similar issue here, did you ever get to the bottom of this?

Appears `tr.scaleZoom(tr.scale * Math.min(scaleX, scaleY));` being passed tr.scale = 1, scaleX = 35.78821257289714, scaleY = -36.83207022162391

<img width="944" alt="screen shot 2016-10-30 at 08 09 33" src="https://cloud.githubusercontent.com/assets/43741/19835275/addc7362-9e78-11e6-9a39-c09d3025450c.png">

I'm guessing because it's taking the negative point from the `min()` call before going to `tr.scaleZoom()`, but I've a limited understanding here of why exactly since the returned lat/long coord is correct, just the zoom that is wrong.

jwarning:
This seems to be due to passing a negative value to the scaleZoom method in some cases. If the line is changed to `const zoom = tr.scaleZoom(tr.scale * Math.abs(Math.min(scaleX, scaleY)));` it works as expected again.

abmai:
Hey guys. Were you able to come up with a solution for this - if so, can you drop a PR?

jwarning:
Yeah I could probably do so. The fix relies on ensuring that the scale passed to the scaleZoom method is not negative. The line I posted above fixes the issue.

Apercu:
Resolved by #159.

PR #159, opened 2016-12-07
Ensure fitBounds doesn't return NaN zoom value

jwarning:
This should fix the NaN zoom value in issue https://github.com/uber/react-map-gl/issues/130.


Apercu:
Just out of curiosity, doesn't this result in zoom values being incoherent with the current state of the map?

jwarning:
In what way would it be incoherent? The issue as far as I understand is that the function at https://github.com/mapbox/mapbox-gl-js/blob/master/js/geo/transform.js#L180 is being passed a negative scale, which when given to Math.log() will return NaN. My guess could be that this happens when dealing with negative latitude values perhaps?

Apercu:
Yeah I was wondering if we could the exact cause of this behavior.

This should looks good to me, but let me add @abmai as reviewer too.

abmai:
lgtm

PR #160, opened 2016-12-15
Add onLoad event handler

Apercu:
Ability to provide an optional `onLoad` prop that will get called once the map has loaded.

Also take the opportunity to cleanup some things.

Need to investigate why Travis is not calling it though.

### DIFF
--- a/src/map.react.js
+++ b/src/map.react.js
@@ -162,19 +162,26 @@ const PROP_TYPES = {
   /**
     * Specify the bearing of the viewport
     */
-  bearing: React.PropTypes.number,
+  bearing: PropTypes.number,
 
   /**
     * Specify the pitch of the viewport
     */
-  pitch: React.PropTypes.number,
+  pitch: PropTypes.number,
 
   /**
     * Specify the altitude of the viewport camera
     * Unit: map heights, default 1.5
     * Non-public API, see https://github.com/mapbox/mapbox-gl-js/issues/1137
     */
-  altitude: React.PropTypes.number
+  altitude: PropTypes.number,
+
+  /**
+    * The load callback is called when all dependencies have been loaded and
+    * the map is ready.
+    */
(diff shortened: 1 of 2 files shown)


src/map.react.js
```diff
@@ -234,6 +242,10 @@ export default class MapGL extends Component {
       // attributionControl: this.props.attributionControl
     });
 
+    if (this.props.onLoad) {
```
ibgreen: By setting `defaultProps` for `onLoad` to `noop` you can skip the if statement.

test/map-spec.js
```diff
@@ -1,34 +1,39 @@
 import MapGL from '../src';
 
-import document from 'global/document';
-/* global process */
 import React from 'react';
 import ReactDOM from 'react-dom';
+import document from 'global/document';
 import test from 'tape-catch';
 
-/* eslint-disable no-process-env */
```
ibgreen: Is this no longer a lint warning?

Apercu: I disabled it only on one line

ibgreen: OK. Admit it is a little cleaner code wise, but it is also nice to show explicitly what warnings we decide to ignore. Is there a one line syntax that allows us to disable a specific warning? Just curious, I am fine either way.

Apercu: Yep we could specify which one we need to remove, I can also change that.

abmai: I lost some context here, did we decide to go with `// esline-disable-line` instead of the multi-line `/* eslint-disable .... */`?

Apercu: Yes, it's more concise, in this specific use-case at least.

Apercu:
Sorry for the wait @contra, Travis being Travis (not my boss, the other one).

contra:
@Apercu Thanks!

PR #161, opened 2016-12-20
Remove mapbox Point dependency from map-interactions. 

ibgreen:
Update eslint settings.
Use static for propTypes (now that it static members is included in stage-2).


src/map-interactions.react.js
```diff
@@ -206,6 +214,7 @@ export default class MapInteractions extends Component {
   /* eslint-disable complexity, max-statements */
   @autobind
   _onWheel(event) {
+    event.stopPropagation();
```
abmai: We had an issue and a subsequent PR to remove this `stopPropagation` a couple of weeks ago. Is there a reason why propagation needs to be stopped here?

src/map.react.js
```diff
@@ -278,10 +282,6 @@ export default class MapGL extends Component {
     return 'inherit';
   }
 
-  _getMap() {
```
abmai: Many applications have come to depend on this to get to the internal map instance. We should add a console warning to this for a couple of versions before deprecating it completely. Can we add this back with a console warning? Something like "you can access the `_map` object directly and this method will eventually be deprecated" will do.

src/map.react.js
```diff
@@ -505,21 +504,33 @@ export default class MapGL extends Component {
     this._onMouseDown({pos});
   }
 
+  @autobind _onTouchDrag({pos}) {
```
abmai: Do we need to use de-structuring here vs. just passing the entire argument as `opt` through?

src/map.react.js
```diff
@@ -505,21 +504,33 @@ export default class MapGL extends Component {
...
+    this._onMouseDrag({pos});
+  }
+
+  @autobind _onTouchRotate({pos, startPos}) {
```
abmai: Ditto as above.

src/map-interactions.react.js
ibgreen: > Is there a reason why propagation needs to be stopped here?
Good catch - removed.

src/map.react.js
ibgreen: > Do we need to use de-structuring here vs. just passing the entire argument as opt through?

Updated all the touch events to just pass on `opt`

src/map-interactions.react.js
Apercu: https://github.com/uber/react-map-gl/pull/158 for reference

src/map.react.js
abmai: Yep that sounds good. We can move towards removing `_getMap()` eventually as well but we should add deprecation warnings for a couple of versions so people aren't caught completely off guard.

abmai:
Are the changes already made? I don't see them updated in the PR - sorry for the confusion, asking since your comments indicate that you already made the changes.

ibgreen:
> @ibgreen Are the changes already made? I don't see them updated in 

Now they are. Was responding while I was editing the code, sorry for the confusion.

abmai:
lgtm, merging

PR #162, opened 2016-12-21
Split map into static-map and interactive-map. Inversion of control to enable customized control schemes.

ibgreen:
@abmai - this is a non-functional PR, for exposition/early review.

Notably, it shows how `map.react.js` is being split into two components: `map-display` and `mercator-interactions` (tentative names, happy to take suggestions).

The old map-interactions has been renamed `event-manager`.

If looks ok, the trick will now be to get all the details right and the code working again.


ibgreen:
@abmai @vicapow This PR is now functional and ready for review.

It is a major refactor that separates the concern of viewport event handling (from the static display of a map) into two separate components.

src/static-map.js
```diff
@@ -451,241 +377,81 @@ export default class MapGL extends Component {
...
-          ...this.props.style,
-          width: this.props.width,
-          height: this.props.height,
-          position: 'relative'
-        } }>
-
-        { content }
+      <div style={mapContainerStyle}>
+
+        <EventManager
```
abmai: I think there is something odd around the `StaticMap` wrapping an `EventManager` here. I understand the want is to expose hover/click events, but perhaps we shouldn't even do that. A `StaticMap` should just take in a set of mapbox-gl compatible props and render the map.

One other reason to perhaps take out the EventManager here is to lessen the layers when someone is trying to render an `InteractiveMap`. Right now that one render of `InteractiveMap` would look like this (to my understanding):
```
MapControls -> EventManager -> EventManager -> MapGL
```

src/map-controls.js
```diff
@@ -0,0 +1,373 @@
...
+    const mapEventLayerStyle = {
+      ...style,
+      width,
+      height,
+      position: 'relative',
+      cursor: this._getCursor()
+    };
+
+    return (
+      <div className={className} style={mapEventLayerStyle}>
```
abmai: Is it necessary to render this `<div>` here or could we possibly just return `<EventManager>`?

abmai:
Thanks, Ib! This is an awesome start!! I think having this code to chat about is going to be super helpful. I like that we're trying to provide smaller components that have explicit responsibilities.

Towards that end though, I think we could better define the roles of these 3 separate components:

`StaticMap`
`MapControls`
`EventManager`

Right now it seems at least like they're still quite coupled together (both `StaticMap` and `MapControls` are dependent on `EventManager`). And a single render of `InteractiveMap` requires 4 container divs whereas before it only required 2.

ibgreen:
Closing. Reopened as #170 towards dev.

PR #164, opened 2016-12-22
use 'changedTouches' for 'touchend'/'touchcancel' events

### LINKED ISSUE #139: touchend event triggered for incorrect position
I believe the `touchend` event handler should be using the `changedTouches` property on the `event` rather than the `touches` property. 

I noticed that the `onTouchTap` handler is always called with position 0 / 0 because the `touches` property is always an empty array during the `touchend` event (compare: [The `touchend` event](https://www.w3.org/TR/touch-events/#dfn-touchend)). I think the [`_onTouchEnd`](https://github.com/uber/react-map-gl/blob/master/src/map-interactions.react.js#L183-L192) handler should use the `changedTouches` property instead to find the coordinates of the tap.

Note that this bug does not seem to manifest when tested on android devices, since there are (correct) mouse events triggered following the (incorrect) touch events.

forrert:
Closes #139

### DIFF
--- a/src/map-interactions.react.js
+++ b/src/map-interactions.react.js
@@ -45,13 +45,24 @@ function getMousePosition(el, event) {
 function getTouchPositions(el, event) {
   const points = [];
   const rect = el.getBoundingClientRect();
-  for (let i = 0; i < event.touches.length; i++) {
+  const touches = getTouches(event);
+  for (let i = 0; i < touches.length; i++) {
     points.push([
-      event.touches[i].clientX - rect.left - el.clientLeft,
-      event.touches[i].clientY - rect.top - el.clientTop
+      touches[i].clientX - rect.left - el.clientLeft,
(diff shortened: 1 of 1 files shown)


src/map-interactions.react.js
```diff
@@ -45,15 +45,24 @@ function getMousePosition(el, event) {
...
-      event.touches[i].clientX - rect.left - el.clientLeft,
-      event.touches[i].clientY - rect.top - el.clientTop
+      touches[i].clientX - rect.left - el.clientLeft,
+      touches[i].clientY - rect.top - el.clientTop
     ]);
   }
   return points;
 }
 
+function getTouches(event) {
```
ibgreen: Would be good with a one or two line comment describing this function and why this is needed. Essentially, a one line version of your comment in the issue (#139).

PR #170, opened 2017-01-08
Split map into static-map and interactive-map. Inversion of control to enable customized control schemes.

ibgreen:
Reopening #162 towards dev (instead of master). Copying @abmai's comments from that PR.

Thanks, @ibgreen! This is an awesome start!! I think having this code to chat about is going to be super helpful. I like that we're trying to provide smaller components that have explicit responsibilities.

Towards that end though, I think we could better define the roles of these 3 separate components:

* StaticMap
* MapControls
* EventManager

Right now it seems at least like they're still quite coupled together (both StaticMap and MapControls are dependent on EventManager). And a single render of InteractiveMap requires 4 container divs whereas before it only required 2.


ibgreen:
Merging this (new components, old ones still in place) to make it easier to iterate.

PR #171, opened 2017-01-08
update to mapbox-gl 0.29.0; add maxZoom as a prop

abmai:
This includes modifying the `startDragLngLat` from an array to an object, which might be a breaking change if people have been using this variable other than just passing it back into the map instance. @ibgreen what do you think we should do here?

For context, it was changed to an object because the `transform` utilities in the new mapbox version expects an object that looks like `{lng, lat}` and not an array anymore.

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "d3-scale": "^1.0.3",
     "d3-selection": "^1.0.2",
     "global": "^4.3.0",
-    "mapbox-gl": "0.26.0",
+    "mapbox-gl": "0.29.0",
     "pure-render-decorator": "^1.1.0",
     "svg-transform": "0.0.3",
     "viewport-mercator-project": "^2.0.1"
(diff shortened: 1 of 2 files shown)


ibgreen:
> @ibgreen what do you think we should do here?

I think we should stay with lnglat array if possible.

While the `startDragLngLat` is rarely manipulated by the user, we have gone to a lot of effort standardizing on simple lnglat arrays in the rest of our web mapping stack (deck.gl, viewport-mercator-project etc). I felt that this standardization was important as it is confusing to deal with different representations of the same thing in an untyped programming language.

I think we should have a really good reason if we want to diverge from that pattern in our public API. Converting back and forth between an array and an object during event handling is cheap, and doing so would allow us to maintain a simple message: that we always use `lnglat` arrays for coordinates.

abmai:
Thanks! I'll keep using the [lng, lat] as a convention then.

PR #173, opened 2017-01-09
Dev 0.29

ibgreen:
@abmai - Trying to merge mapbox 0.29 into webpack based dev branch - ran into trouble with the transform.js dependency - the mapbox code base now uses flow syntax in their JavaScript so our practice of hand-importing selected source files (instead of their prebuilt "bundle") causes problems in non-browserify based workflows.


abmai:
hmm. why would this be webpack specific? perhaps 2.0 is also a good time for us to experiment with flow in our codebase as well - would that help here?

PR #174, opened 2017-01-09
Dev 0.29

ibgreen:
Reopening towards dev

@abmai - Trying to merge mapbox 0.29 into webpack based dev branch - ran into trouble with the transform.js dependency - the mapbox code base now uses flow syntax in their JavaScript so our practice of hand-importing selected source files (instead of their prebuilt "bundle") causes problems in non-browserify based workflows.

### DIFF
--- a/package.json
+++ b/package.json
@@ -24,7 +24,7 @@
     "bowser": "^1.2.0",
     "d3-selection": "^1.0.2",
     "global": "^4.3.0",
-    "mapbox-gl": "0.26.0",
+    "mapbox-gl": "0.29.0",
     "react-addons-shallow-compare": "^15.4.2",
     "viewport-mercator-project": "^2.0.1"
   },
--- a/package.json
+++ b/package.json
@@ -39,7 +39,6 @@
     "immutable": "^3.7.6",
     "jsdom": "^9.9.1",
     "module-alias": "^2.0.0",
-    "node-jsx": "^0.13.3",
     "pre-commit": "^1.2.2",
     "react": "^15.4.0",
     "react-addons-test-utils": "^15.4.2",
--- a/src/components/map-interactions.js
+++ b/src/components/map-interactions.js
@@ -46,13 +46,24 @@ function getMousePosition(el, event) {
 function getTouchPositions(el, event) {
   const points = [];
   const rect = el.getBoundingClientRect();
-  for (let i = 0; i < event.touches.length; i++) {
+  const touches = getTouches(event);
+  for (let i = 0; i < touches.length; i++) {
     points.push([
-      event.touches[i].clientX - rect.left - el.clientLeft,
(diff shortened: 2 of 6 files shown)


ibgreen:
> hmm. why would this be webpack specific? 

My data is partly based on mapbox's own comments in [mapbox-gl-js README](https://github.com/mapbox/mapbox-gl-js#using-mapbox-gl-js-with-other-module-systems), and partly based on the issues we have been hearing from our react-map-gl users trying to use mapbox-gl with non-browserify environments.

> perhaps 2.0 is also a good time for us to experiment with flow in our codebase as well - would that help here?

Yes should be a good discussion. As for my starting point, I'm 100% for flow **in comments**. But I'm not excited about adding more non-standard syntax to our JavaScript. 

Adding more non-standard syntax would go against the cleanup I did here the last few days, removing non-ES6 constructs and replacing trivial uses of JSX with `createElement`. In fact, as of today, on the dev branch in this repo, the babel transpiler is no longer even installed, and the tests run under Node 6+ as raw, untranspiled ES6. If we put in flow syntax, that would likely have to be reversed.

abmai:
I am at my wit's ends for the night with this. Adding the `remove-flow-types-loader` works for bundling + running examples but doesn't work for tests. Tried to do something smart with `externals` (let webpack include the mapbox-gl js utils in our bundle) but still run into the same flow annotation error.

Will try again at some point tomorrow. In mean time, they also released 0.30 so I'll probably look into pushing us there on master as well.

abmai:
My changes are up on the `dev029` branch if you want to take a look.

ibgreen:
>Adding the remove-flow-types-loader works for bundling + running examples but doesn't work for tests.

@abmai 

- The tests fail because the tests are currently set up to run without transpilation (i.e. they run as ES6 code)

- If we don't care about that, the easiest way is probably to create a second webpack config file for bundling a full test suite (with all dependencies, as opposed to just bundling the library files). deck.gl dev does this for running tests in the browser.

- Another concern is that this solution pushes the burden of applying the `remove-flow-type-transform` to the app (your example shows this) which doesn't feel great.

- As an alternative, I started playing with including the "deflowed" transform.js in the output bundle, but didn't get all the way. Setting it up so that the consuming app doesn't need to fiddle with `resolves` would also be desirable.

- Ultimately, we should another push on removing the need for directly importing `transform`. We are almost there...

abmai:
There are a few choices, but I think our solution here depends on the last point - whether we see ourselves continuing to use their `Transform` class which isn't officially supported for public use.

I tried doing your third point as well (fiddling with `externals` so that `mapbox-gl/js/geo` and `mapbox-gl/js/util` are bundled (and therefore stripped of flow syntax) to no avail yet.

Another option here is to have a separate webpack config just for tests - which might be desirable just to keep the two environment more separate.

PR #186, opened 2017-01-26
Adapt exhibits for react-map-gl

Apercu:


exhibits/app.js
```diff
@@ -0,0 +1,50 @@
+import 'babel-polyfill';
```
abmai: What's the `babel-polyfill` used for here? Maybe a couple of comments to explain its use would help with understanding.

exhibits/browserify/package.json
```diff
@@ -0,0 +1,33 @@
+{
+  "name": "browserify-exhibit",
+  "version": "0.0.0",
+  "description": "A browserify/react-map-gl exhibit",
+  "scripts": {
+    "link": "(cd .. && rm -f node_modules && ln -s browserify/node_modules)",
```
ibgreen: For what its worth, while this is clever, I think this should be removed. The purpose here is to make it super easy for someone to get started - either with browserify or with webpack. 

The simplest way to do that is to provide two complete example folders. Duplicating a file like app.js is not an issue.

This line is not easy to understand for a lot of people, and will not work/make sense when they try to copy this to start their own app.

exhibits/app.js
ibgreen: babel-polyfill might not be needed in react-map-gl?

abmai:
TBH aren't `exhibits` and `examples` the same thing? We can just use the example name itself to differentiate and everything can just be `examples`.

exhibits/app.js
Apercu: ah yeah it might, not rechecked

Apercu:
Yes my bad, will move them

PR #188, opened 2017-01-27
update package.json engines to reflect new node and npm requirements

abmai:


package.json
```diff
@@ -79,7 +79,7 @@
     "react": "0.14.x - 15.x"
   },
   "engines": {
-    "node": "0.10.x",
-    "npm": "3.x"
+    "node": ">= 4.x",
```
chrisirhc: not sure that `.x` is compatible, can probably do `>=4`

abmai: ah good point. it's not necessary with this semantic anywho

abmai:
The space should be ok. Tested with a version switcher locally and it's also used in `express`'s package.json. Thanks for the review!

PR #189, opened 2017-02-04
remove d3-selection and extra example map

abmai:
I've also examined the use of `global` and `react-addons-shallow-compare` and for now I think:

`global` need to stay because it's a pretty safe alternative to make sure `document` and `window` work across node and browser. It only add hundreds of bytes to the final bundle.

`react-addons-shallow-compare` will require additional thought, since we can move to `React.PureComponent` if we lock onto the latest versions of React. I don't want to do work to come up with a custom `shouldComponentUpdate` only to use `PureComponent` later. That will come in a separate diff.


examples/main/app.js
```diff
@@ -51,7 +51,6 @@ export default class App extends Component {
         <NotInteractiveExample {...common}/>
         <StyleDiffingExample {...common}/>
         <ClickExample {...common }/>
-        <StyleDiffingExample {...common}/>
```
ibgreen: This leaves the style diffing example source file in the tree right? Either delete it, or move it to a separate standalone example.

ibgreen:
> react-addons-shallow-compare will require additional thought, since we can move to React.PureComponent if we lock onto the latest versions of React. I don't want to do work to come up with a custom shouldComponentUpdate only to use PureComponent later. That will come in a separate diff.

I don't think we should mandate React 15.4 just yet which I believe would be required for `PureComponent`. If you decide to do a custom `shouldComponentUpdate`, you can always reuse the battle-tested one from deck.gl.

PR #191, opened 2017-02-09
split flow-remove-types script into prestart and postinstall

abmai:


package.json
```diff
@@ -61,13 +61,14 @@
   },
   "scripts": {
     "build": "babel src -d dist --copy-files --source-maps inline",
+    "prestart": "flow-remove-types node_modules/mapbox-gl/js/**/*.js --out-dir . >/dev/null 2>&1",
```
ibgreen: How about `npm test`? With this approach, won't we need a `pretest` as well (or we have to run `npm start` before `npm test`)?

Also, having `flow-remove-types` run every time we `npm start` seems unnecessary. We really only need to run it once after an install.

How about adding it to `npm build`?
```
"build": "npm run flow-remove-types && babel src -d dist --source-maps inline",
"flow-remove-types": "flow-remove-types node_modules/mapbox-gl/js/**/*.js --out-dir . >/dev/null 2>&1",
```

abmai:
@ibgreen makes sense. I was a bit confused between `npm build` and `npm run build` and thought the latter would be ran during `npm install` as well but that's not the case.

@Apercu this works with npm@3 >= and is only needed for development. Production installs will still run `postinstall` normally.

It seems though that this solution isn't the one we want, anyway. Since there's still concerns here about it working on Windows. I'll look into `shelljs`.

Or we could be beating around the wrong bush here since all of this is hacking around `mapbox-gl`'s flow types annotation..

ibgreen:
> Or we could be beating around the wrong bush here since all of this is hacking around mapbox-gl's flow types annotation..

Yes, the best way to solve this completely is to stop importing internal mapbox-gl files (i.e. transform.js). Then we'd just consume the prebuilt bundle from the mapbox-gl dist folder. 

The necessary refactoring work has started on the dev branch but it would probably take 1-2 days of work to complete and the work is a bit math heavy as it involves perspective projections.

PR #198, opened 2017-03-15
Fix polyfill under IE11

Pessimistress:


package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "react-map-gl",
   "description": "A React wrapper for MapboxGL-js and overlay API.",
-  "version": "1.7.2",
+  "version": "1.7.3",
```
ibgreen: I think the version needs to bump past the two "broken" versions that were made outside of master, so change this to 1.8.2 or 1.9 (not needing babel-polyfill is kind of a new feature).

abmai:
Looks like Travis failed on an unrelated error. It's fine to just land this, I think.

PR #206, opened 2017-04-12
use prop-types package

joeduncan:
replace deprecated `React.PropTypes` with `prop-types` package

### DIFF
--- a/package.json
+++ b/package.json
@@ -38,6 +38,7 @@
     "bowser": "^1.2.0",
     "immutable": "*",
     "mapbox-gl": "0.32.1",
+    "prop-types": "^15.5.7",
     "viewport-mercator-project": "^4.0.0-alpha.5"
   },
   "devDependencies": {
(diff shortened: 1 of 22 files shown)


package.json
```diff
@@ -27,7 +27,6 @@
     "build-es5": "mkdir -p dist && babel src --out-dir dist --plugins=transform-es2015-modules-commonjs --source-maps inline",
     "build-es6": "mkdir -p dist-es6 && babel src --out-dir dist-es6 --source-maps inline",
     "clean": "rm -fr dist/* dist-es6/*",
-    "clean": "rm -fr dist/* dist-es6/*",
```
ibgreen: Why remove the "clean" script? I'd prefer that we didn't. If there is a reason to do it, it should be done in a separate PR.

joeduncan: it was in there twice...but I see it was removed by last commit https://github.com/uber/react-map-gl/commit/6e3a934613a8711728c929cb85558b1ada073c6f

joeduncan: I'll rebase, give me a sec

abmai:
Looks great, and looks like clean deletion is removed. Thanks, @joeduncan!

PR #211, opened 2017-04-27
Fix various map control bugs

Pessimistress:
- Smooth rotation
- Smooth pan
- Zoom around mouse position

### DIFF
--- a/src/components/map-controls.js
+++ b/src/components/map-controls.js
@@ -91,10 +91,6 @@ const propTypes = {
   /** Pitch when current perspective drag operation started */
   startPitch: PropTypes.number,
 
-  /* Hooks to get mapbox help with calculations. TODO - replace with Viewport */
-  unproject: PropTypes.func,
-  getLngLatAtPoint: PropTypes.func,
-
   /**
     * True means key must be pressed to rotate instead of pan
     * false to means key must be pressed to pan
@@ -113,9 +109,6 @@ const defaultProps = {
   minZoom: 0,
   maxPitch: MAX_PITCH,
   minPitch: 0,
-
-  unproject: null,
-  getLngLatAtPoint: null,
 
   pressKeyToRotate: true
 };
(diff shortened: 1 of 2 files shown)


src/components/map-controls.js
```diff
@@ -240,7 +240,7 @@ export default class MapControls extends PureComponent {
   }
 
   _unproject(pos) {
-    const viewport = new PerspectiveMercatorViewport(Object.assign({}, this.props));
+    const viewport = new PerspectiveMercatorViewport(this.props);
     const lngLat = this.props.unproject ?
```
ibgreen: I think we can remove this.props.unproject now (it was there to allow injecting mapbox version during development)

src/components/map-controls.js
```diff
@@ -251,10 +251,7 @@ export default class MapControls extends PureComponent {
   // TODO - We should have a mapbox-independent implementation of panning
   // Panning calculation is currently done using an undocumented mapbox function
   _calculateNewLngLat({startDragLngLat, pos, startPos}) {
-    const viewport = new PerspectiveMercatorViewport(Object.assign({}, this.props, {
-      longitude: startDragLngLat[0],
-      latitude: startDragLngLat[1]
-    }));
+    const viewport = new PerspectiveMercatorViewport(Object.assign({}, this.props));
```
ibgreen: Remove the `Object.assign` here as well if we're doing it above?

src/components/map-controls.js
```diff
@@ -282,8 +279,8 @@ export default class MapControls extends PureComponent {
     if (yDelta > 0) {
       // Dragging downwards, gradually decrease pitch
       if (Math.abs(this.props.height - startPos[1]) > PITCH_MOUSE_THRESHOLD) {
-        const scale = yDelta / (this.props.height - startPos[1]);
-        pitch = (1 - scale) * PITCH_ACCEL * startPitch;
+        const scale = yDelta / (this.props.height - startPos[1]) * PITCH_ACCEL;
```
ibgreen: While I don't doubt this behaves better, this still looks weird to me, is `scale` still the right name after this change (if it ever was) or is there something more intuitive.

abmai: Perhaps just modify the equation on line 286 instead? i.e. `(1 - scale * PITCH_ACCEL) * startPitch`

PR #212, opened 2017-04-27
Separating input events and semantic transform in map control

Pessimistress:
*This is a work in progress, for discussion purpose only.*

The proposal is to separate raw input events (drag, wheel, pinch, etc.) and user intent (pan, rotate, zoom). The calculation of semantic transform is handled in the DOM-independent `MapState` class. The `MapControls` component subscribes to DOM events and translates them into user intent. By modifying `MapControls` or implementing their own components, users can easily achieve the following example scenarios:
- Switch between drag to rotate and drag to pan
- Drag with right mouse button to rotate
- Navigate with keyboard
- Use a 3rd-party event manager such as hammer.js

This abstraction also make the [NavigationControl](https://www.mapbox.com/mapbox-gl-js/example/navigation/) essentially a variation of the `MapControls` component.

It also allows us to reuse the EventManager class for non-geospatial viewports in other context.

### DIFF
--- a/src/components/map-controls.js
+++ b/src/components/map-controls.js
@@ -17,53 +17,23 @@
 // LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
-import {PerspectiveMercatorViewport} from 'viewport-mercator-project';
 import React, {PureComponent} from 'react';
 import PropTypes from 'prop-types';
-import assert from 'assert';
 
 import autobind from '../utils/autobind';
 
 // MapControls uses non-react event manager to register events
 import EventManager from '../utils/event-manager';
+import MapState from '../utils/map-state';
 
 import config from '../config';
-
-function mod(value, divisor) {
-  const modulus = value % divisor;
-  return modulus < 0 ? divisor + modulus : modulus;
-}
-// MAPBOX LIMITS
-const MAX_PITCH = 60;
-const MAX_ZOOM = 40;
 
 // EVENT HANDLING PARAMETERS
 const PITCH_MOUSE_THRESHOLD = 5;
 const PITCH_ACCEL = 1.2;
+const ZOOM_ACCEL = 0.01;
 
-const propTypes = {
-  /** The width of the map */
-  width: PropTypes.number.isRequired,
-  /** The height of the map */
-  height: PropTypes.number.isRequired,
-  /** The latitude of the center of the map. */
-  latitude: PropTypes.number.isRequired,
-  /** The longitude of the center of the map. */
-  longitude: PropTypes.number.isRequired,
-  /** The tile zoom level of the map. */
-  zoom: PropTypes.number.isRequired,
-  /** Specify the bearing of the viewport */
-  bearing: React.PropTypes.number,
-  /** Specify the pitch of the viewport */
-  pitch: React.PropTypes.number,
-  /**
-    * Specify the altitude of the viewport camera
-    * Unit: map heights, default 1.5
-    * Non-public API, see https://github.com/mapbox/mapbox-gl-js/issues/1137
-    */
-  altitude: React.PropTypes.number,
-
-  constraints: React.PropTypes.object,
+const propTypes = Object.assign({}, MapState.propTypes, {
 
   /** Enables perspective control event handling */
   perspectiveEnabled: PropTypes.bool,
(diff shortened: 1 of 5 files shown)


shaojingli:
This is the correct way to do it. 

We also need this for separating user intents for controlling camera and interactively modifying data.

abmai:
This is great! I would add that we should add tests for the more DOM-independent pieces as we add new code to master from now on.

src/utils/map-state.js
```diff
@@ -0,0 +1,332 @@
...
+// MAPBOX LIMITS
+const MAX_PITCH = 60;
+const MAX_ZOOM = 20;
+
+function mod(value, divisor) {
+  const modulus = value % divisor;
+  return modulus < 0 ? divisor + modulus : modulus;
+}
+
+const propTypes = {
```
ibgreen: Something about putting the prop types here doesn't feel quite right.

The MapState class is React independent, and PropTypes is still a React-related thing (it is certainly intended for React here).

Maybe just separate in two utility files map-state and map-props?

test/utils/map-state.spec.js
```diff
@@ -0,0 +1,184 @@
...
+    maxPitch: 90
+  }
+];
+
+// Discard precision errors for comparison
+function toLowPrecision(input, precision = 11) {
+  if (typeof input === 'number') {
+    input = Number(input.toPrecision(precision));
+  }
+  /* eslint-enable guard-for-in */
```
ibgreen: eslint from copy paste? Cut?

test/utils/map-state.spec.js
```diff
@@ -0,0 +1,184 @@
...
+    zoom: 14,
+    pitch: 60,
+    bearing: 45,
+    maxPitch: 90
+  }
+];
+
+// Discard precision errors for comparison
+function toLowPrecision(input, precision = 11) {
+  if (typeof input === 'number') {
```
ibgreen: Nit: Maybe make this a one-line ternary? Making this shorter will detract less from the tests.

src/utils/map-state.js
```diff
@@ -0,0 +1,332 @@
...
+      minZoom,
+      maxPitch,
+      minPitch,
+      startPanLngLat,
+      startBearing,
+      startPitch
+    };
+  }
+
+  _updateViewport(opts) {
```
ibgreen: Always put the public method first and the private at the bottom, most users scan a class from top to bottom.

src/utils/map-state.js
```diff
@@ -0,0 +1,332 @@
...
+    // Update props
+    Object.assign(this.props, opts);
+    this._applyConstraints();
+    return this;
+  }
+
+  // Apply any constraints (mathematical or defined by props) to viewport params
+  _applyConstraints() {
+    const viewport = this.props;
```
ibgreen: We are mutating props here, is that desirable? 

Should we comment in that case, explain why this is OK.

Should we take the props as an argument instead of using `this.props`?

src/utils/map-state.js
```diff
@@ -0,0 +1,332 @@
...
+    minPitch = 0,
+
+    /** Interaction states */
+    startPanLngLat,
+    startZoomLngLat,
+    startBearing,
+    startPitch,
+    startZoom
+  } = {}) {
+    this.props = {
```
ibgreen: Not sure about calling this `props`. Understand that these are intended to be used as props, but the word becomes overloaded, a little hard to see at a glance if I am inside React or not.

src/utils/map-state.js
ibgreen: The name `_updateViewport` - this is not a `Viewport` and it doesn't really update the object.
- should it be `_getUpdatedMapState()` or similar

src/utils/map-state.js
```diff
@@ -0,0 +1,332 @@
...
+  zoomEnd() {
+    return this._updateViewport({
+      startZoomLngLat: null,
+      startZoom: null
+    });
+  }
+
+}
+
+MapState.propTypes = propTypes;
```
ibgreen: This looks particularly weird given this is not a react component.

src/components/map-controls.js
```diff
@@ -179,213 +109,88 @@ export default class MapControls extends PureComponent {
...
+  _onTouchRotate(opts) {
+    this._onMouseRotate(opts);
   }
 
-  _unproject(pos) {
-    const viewport = new PerspectiveMercatorViewport(this.props);
-    return viewport.unproject(pos, {topLeft: false});
+  _onMouseDown({pos}) {
+    const mapState = new MapState(this.props).panStart({pos}).rotateStart({pos});
+    this._updateViewport(mapState.props, {isDragging: true});
```
ibgreen: Should we pass in `mapState` instead so that these functions only need to know about that class and not care about its structure.

src/components/map-controls.js
```diff
@@ -179,213 +109,88 @@ export default class MapControls extends PureComponent {
...
-    // if (viewport.startDragLngLat) {
-    //   const dragViewport = new FlatMercatorViewport(Object.assign({}, this.props, {
-    //     longitude: viewport.startDragLngLat[0],
-    //     latitude: viewport.startDragLngLat[1]
-    //   }));
-    //   this.setState({dragViewport});
-    // }
-
-    return this.props.onChangeViewport(viewport);
+  _updateViewport(...opts) {
```
ibgreen: Since there is no longer any references to `Viewports`, should we call this `_updateMapState` instead to complete the nice `MapState` abstraction here and avoid conflating concepts? 

For consistency, it would probably have been nice to change the callback name to `onChangeMapState` as well but that is not worth it.

Pessimistress: I think `onChangeViewport` is still valuable because it allows the user to apply additional constraints. We can have a separate discussion around the interface of InteractiveMap, until then I'll keep it unchanged.

PR #217, opened 2017-05-01
Consistent handling of map children

Pessimistress:
- Assign the same class name to children container between InteractiveMap and StaticMap
- Always render map children under the same node (avoid unmount/mount when the map is shown/hidden)

### DIFF
--- a/src/components/interactive-map.js
+++ b/src/components/interactive-map.js
@@ -90,14 +90,14 @@ export default class InteractiveMap extends PureComponent {
           ref: map => {
             this._map = map;
           },
-          children: mapVisible && this.props.children
+          children: null
         })),
         createElement('div', {
-          key: 'map-children',
-          style: {position: 'absolute', left: 0, top: 0}
-        },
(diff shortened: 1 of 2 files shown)


src/components/interactive-map.js
```diff
@@ -90,14 +90,14 @@ export default class InteractiveMap extends PureComponent {
...
+          children: null
         })),
         createElement('div', {
-          key: 'map-children',
-          style: {position: 'absolute', left: 0, top: 0}
-        },
-          !mapVisible && this.props.children
-        )
+          key: 'map-overlays',
+          className: 'overlays',
```
ibgreen: Did this class name need to be synched with static-map? Should we add comments in both places to that effect?

Pessimistress: It doesn't have to. The class name of the container makes it easier to do css styling. Having the same class name makes css reusable for both static and interactive maps. If we decide it's not necessary then both should be removed?

PR #218, opened 2017-05-01
bump mapbox-gl to 0.27 to prevent its devDependencies from being installed.

lxe:
mapbox-gl contained a shrinkwrap with all of its devDependencies listed:
https://github.com/mapbox/mapbox-gl-js/issues/3374

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,7 +26,7 @@
     "d3-scale": "^1.0.3",
     "d3-selection": "^1.0.2",
     "global": "^4.3.0",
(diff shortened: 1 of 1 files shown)


ibgreen:
@abmai pretty sure this will cause flow syntax issues in apps (as I recall, handling flow syntax in 0.27 is the main reason we cut react-map-gl v2).

We need to find an earlier version mapbox version instead.

abmai:
Closing this as we shouldn't be bumping `v1` to mapbox 0.27.

PR #223, opened 2017-05-03
Improve MapState class and tests

Pessimistress:
Addressing comments on [PR#212](https://github.com/uber/react-map-gl/pull/212)
- `MapControl` takes a `MapState` instance instead of list of viewport props
- Always return a new `MapState` object after transform
- Better naming of properties and methods
- Code consistency and error checking
- Tests cover more use cases

### DIFF
--- a/src/utils/map-state.js
+++ b/src/utils/map-state.js
@@ -1,149 +1,296 @@
-import PropTypes from 'prop-types';
 import {PerspectiveMercatorViewport} from 'viewport-mercator-project';
 import assert from 'assert';
-
-// MAPBOX LIMITS
-const MAX_PITCH = 60;
-const MAX_ZOOM = 20;
 
 function mod(value, divisor) {
   const modulus = value % divisor;
   return modulus < 0 ? divisor + modulus : modulus;
 }
 
-const propTypes = {
-  /** The width of the map */
-  width: PropTypes.number.isRequired,
-  /** The height of the map */
-  height: PropTypes.number.isRequired,
-  /** The latitude of the center of the map. */
-  latitude: PropTypes.number.isRequired,
-  /** The longitude of the center of the map. */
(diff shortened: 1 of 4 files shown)


src/utils/map-state.js
```diff
@@ -1,183 +1,105 @@
...
-      minZoom,
-      maxPitch,
-      minPitch,
+      bearing: Number.isFinite(bearing) ? bearing : defaultState.bearing,
+      pitch: Number.isFinite(pitch) ? pitch : defaultState.pitch,
+      altitude: Number.isFinite(altitude) ? altitude : defaultState.altitude,
+      maxZoom: Number.isFinite(maxZoom) ? maxZoom : defaultState.maxZoom,
+      minZoom: Number.isFinite(minZoom) ? minZoom : defaultState.minZoom,
+      maxPitch: Number.isFinite(maxPitch) ? maxPitch : defaultState.maxPitch,
+      minPitch: Number.isFinite(minPitch) ? minPitch : defaultState.minPitch,
```
abmai: Nit: Can we extract this logic to a helper above? Like `ensureFinite` that takes the value and a defaultValue. It's terrifying to see so many ternary in a row.

src/utils/map-state.js
```diff
@@ -1,183 +1,105 @@
-import PropTypes from 'prop-types';
 import {PerspectiveMercatorViewport} from 'viewport-mercator-project';
 import assert from 'assert';
 
-// MAPBOX LIMITS
-const MAX_PITCH = 60;
```
ibgreen: Going beyond 60 hides the map, this might be unexpected/undesired default behavior.

Pessimistress: `maxPitch` is still default to 60, I moved it to `defaultState`.

PR #225, opened 2017-05-03
Add marker component

Pessimistress:
A React equivalent to Mapbox's [Marker control](https://www.mapbox.com/mapbox-gl-js/api/#marker):
screen shot 2017-05-03 at 3 34 45 pm

Notes on using context:
- Pro: avoids destructing a long list of viewport props to pass down to all map overlay components
- Pro: the parent component that renders the Markers can be agnostic of the viewport
- Con: I believe context is still marked as an experimental feature of React
- Con: not supported by PureComponent. Though in this case it's not unreasonable to always re-render markers when the map updates, which is a very thin wrapper, and render the children of the markers as PureComponents.

### DIFF
--- a/examples/custom-interactions/view.js
+++ b/examples/custom-interactions/view.js
@@ -38,15 +38,12 @@ const locations = Immutable.fromJS(new Array(4000).fill(0).map(() => [
 ]));
 
 const MARKER_STYLE = {
-  fill: '#d00',
-  transform: 'translate(-50%,-100%)',
   cursor: 'pointer'
 };
 
-const MARKER_ICON = `M45.4,39.7C45.4,39.7,45.4,39.7,45.4,39.7c2.8-4.1,4.4-8.9,4.4-14.2
- C49.9,11.7,38.7,0.6,25,0.6S0.1,11.7,0.1,25.4c0,5,1.5,9.6,4,13.4c0.1,0.2,0.3,0.5,0.5,
- 0.8c0.1,0.1,0.2,0.3,0.3,0.4 c0.6,0.8,1.1,1.5,1.8,2.2C13.1,49.8,25,61,25,61s11.9-11.2,
- 18.3-18.7c0.6-0.7,1.2-1.4,1.8-2.2C45.2,39.9,45.3,39.8,45.4,39.7z`;
+const MARKER_ICON = `M18.2,15.7L18.2,15.7c1.1-1.6,1.8-3.6,1.8-5.7c0-5.6-4.5-10-10-10S0,4.5,0,10c0,2,0.6,3.9,1.6,5.4c0,0.1,0.1,0.2,0.2,0.3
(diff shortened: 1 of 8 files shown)


examples/custom-interactions/view.js
```diff
@@ -21,7 +21,7 @@
 import React, {Component} from 'react';
 import PropTypes from 'prop-types';
 import autobind from 'react-autobind';
-import {InteractiveMap, ScatterplotOverlay} from 'react-map-gl';
+import {InteractiveMap, ScatterplotOverlay, Marker} from 'react-map-gl';
```
ibgreen: Putting the Marker in this example is fine for now, but we might want to consider how we organize the examples, e.g:
- One example with all the overlays. 
- One example with markers/other react component etc.
- This particular example was initially created to demo breaking mapbox-gl's 60 degree pitch limit.

src/components/interactive-map.js
```diff
@@ -27,6 +29,12 @@ export default class InteractiveMap extends PureComponent {
     autobind(this);
   }
 
+  getChildContext() {
```
ibgreen: If I understand it correctly, the downside of the context approach, compared to child prop injection, is that every component that wraps `StaticMap` needs to define it in this precise way for the `Marker`s to work? 

That seems to be a weakness as this InteractiveMap would need to be defined by apps that want to build their own event handling, which arguably is a key thing we are enabling in v3.0.0

Thoughts?

src/components/marker.js
```diff
@@ -0,0 +1,61 @@
...
+};
+
+export default class Marker extends Component {
+
+  render() {
+    const {longitude, latitude} = this.props;
+
+    const [x, y] = this.context.viewport.project([longitude, latitude]);
+
+    return createElement('div', {
```
ibgreen: So a `Marker` is just a div that get x,y positioned absolutely on projected coordinate?

Any differences from/possible confusion with similar Mapbox features that need to be mentioned in documentation? (I am not fully up-to-date on the original API).

src/components/interactive-map.js
Pessimistress: If you render the markers as children of the StaticMap then StaticMap will provide the context. Because we support hiding the map in this component, the markers are not nested under the map.
For the custom control scenario, we could add an optional prop "MapControl" that allows users to pass in a custom component without implementing InteractiveMap again.

examples/custom-interactions/view.js
Pessimistress: Will do in another PR.

src/components/marker.js
Pessimistress: Using Mapbox you create a htm node using `document.createElement`, wrap it in a Marker or Popup control class, then call `.addTo(map)` (https://www.mapbox.com/mapbox-gl-js/api/#popup#setdomcontent). The default Marker in Mapbox is just an empty div. Popup is essentially a Marker with additional features such as alignment, close button etc.

PR #226, opened 2017-05-04
Add controls example

Pessimistress:
- Revert change to custom-interactions example
- Add a separate example app for Marker control


examples/controls/package.json
```diff
@@ -0,0 +1,30 @@
+{
+  "name": "webpack-exhibit",
```
ibgreen: Delete name/version/description/author etc - goal is to have as few lines of config as possible

examples/controls/webpack.config.js
```diff
@@ -0,0 +1,65 @@
...
+      warnings: false
+    }
+  },
+
+  devtool: 'source-maps',
+
+  resolve: {
+    alias: {
+      // Work against the latest base library in this repo
+      'react-map-gl': resolve('../..'),
```
ibgreen: If we haven't already copied the local development webpack setup of deck and luma here, we should. For most users the example is most valuable if it works against an installed version as it can then be copied and used outside of this repo.

examples/controls/webpack.config.js
```diff
@@ -0,0 +1,65 @@
...
+      immutable: resolve('./node_modules/immutable'),
+      // Per mapbox-gl-js README for non-browserify bundlers
+      'mapbox-gl$': resolve('./node_modules/mapbox-gl/dist/mapbox-gl.js')
+    }
+  },
+
+  module: {
+    rules: [
+      {
+        enforce: 'pre',
```
ibgreen: Why do we need this in react-map-gl v3?

examples/controls/webpack.config.js
```diff
@@ -0,0 +1,65 @@
...
+        test: /\.js$/,
+        loader: 'remove-flow-types-loader',
+        include: [/node_modules\/mapbox-gl\/js/]
+      },
+      {
+        // Compile ES2015 and JSX using buble
```
ibgreen: I saw babelrc configs in package.json - let's go with one of the two.

examples/controls/README.md
```diff
@@ -0,0 +1,12 @@
+<div align="center">
+  <img src="https://avatars3.githubusercontent.com/u/2105791?v=3&s=200" />
```
ibgreen: Reminds me that we still haven't made up our minds on where we want images to be stored... if in repo it is easier to maintain, but they do take some space.

examples/controls/index.html
```diff
@@ -0,0 +1,21 @@
+<!doctype html>
+<html>
+  <head>
+    <meta charset='UTF-8' />
+    <title>react-map-gl Controls Example</title>
+    <style>
+  body {
```
ibgreen: Wonder if it is better to do these here in CSS or programatically. Maybe create a css file to make it a little easier to find?

examples/controls/package.json
```diff
@@ -0,0 +1,30 @@
...
+  "version": "0.0.0",
+  "description": "A webpack/deck.gl exhibit",
+  "scripts": {
+    "start": "webpack-dev-server --progress --hot --open",
+    "start-local": "webpack-dev-server --env.local --progress --hot --open"
+  },
+  "author": "Balthazar Gronon <bgronon@gmail.com>",
+  "license": "MIT",
+  "dependencies": {
+    "immutable": "^3.8.1",
```
ibgreen: react-map-gl v3 should already install immutable, try to remove it.

examples/controls/package.json
```diff
@@ -0,0 +1,30 @@
...
+    "react": "^15.4.1",
+    "react-dom": "^15.4.1",
+    "react-map-gl": "^2.0.1"
```
ibgreen: Let's use 3.0.0-alpha...

examples/controls/package.json
```diff
@@ -0,0 +1,30 @@
...
+  },
+  "devDependencies": {
+    "buble-loader": "^0.4.0",
+    "remove-flow-types-loader": "^1.0.0",
+    "webpack": "^2.2.0-rc.3",
```
ibgreen: 2.4 is the latest no need to use rc

examples/controls/webpack.config.js
```diff
@@ -0,0 +1,65 @@
...
+
+  // Allow setting mapbox token using environment variables
+  plugins: [
+    new webpack.EnvironmentPlugin(['MAPBOX_ACCESS_TOKEN', 'MapboxAccessToken']),
+    new webpack.LoaderOptionsPlugin({minimize: false, debug: true})
+  ]
+};
+
+// DELETE THIS LINE WHEN COPYING THIS EXAMPLE FOLDER OUTSIDE OF DECK.GL
+// It enables bundling against src in this repo rather than installed deck.gl module
```
ibgreen: I've stopped checking in yarn.lock files in the examples. A gitignore entry like this would be good:
```
*/**/yarn.lock
yarn-error.log
```
Top level yarn.lock triggers Travis so it has a purpose, but others just take space.

examples/custom-interactions/view.js
```diff
@@ -21,7 +21,7 @@
 import React, {Component} from 'react';
 import PropTypes from 'prop-types';
 import autobind from 'react-autobind';
-import {InteractiveMap, ScatterplotOverlay, Marker} from 'react-map-gl';
+import {InteractiveMap, ScatterplotOverlay} from 'react-map-gl';
```
ibgreen: Does this file use ScatterplotOverlay? Is linting enabled on examples?

examples/controls/webpack.config.js
Pessimistress: Not sure, all copied from `webpack-exhibit`. We should clean up that project as well.

examples/custom-interactions/view.js
Pessimistress: Yes. I believe so.

PR #227, opened 2017-05-04
Popup control

Pessimistress:
React equivalent of Mapbox's [Popup control](https://www.mapbox.com/mapbox-gl-js/api/#popup)

### DIFF
--- a/examples/controls/app.js
+++ b/examples/controls/app.js
@@ -1,13 +1,14 @@
 /* global window */
 import React, {Component} from 'react';
 import {render} from 'react-dom';
-import MapGL, {Marker} from 'react-map-gl';
+import MapGL, {Marker, Popup} from 'react-map-gl';
 
 import CityPin from './components/city-pin';
 
 import CITIES from './data/cities.json';
 
 import './app.css';
+import 'mapbox-gl/dist/mapbox-gl.css';
 
 const token = process.env.MAPBOX_ACCESS_TOKEN || // eslint-disable-line
   'Set MAPBOX_ACCESS_TOKEN environment variable or put your token here.';
@@ -26,14 +27,16 @@ class Root extends Component {
         longitude: -100,
         zoom: 4,
         bearing: 0,
-        pitch: 60,
+        pitch: 0,
       },
       width: 500,
       height: 500,
+      popupInfo: null
     };
 
     this._updateViewport = this._updateViewport.bind(this);
     this._resize = this._resize.bind(this);
+    this._renderCityMarker = this._renderCityMarker.bind(this);
   }
 
   componentDidMount() {
@@ -58,10 +61,25 @@ class Root extends Component {
 
   _renderCityMarker(city, index) {
     return (
-      <Marker key={index}
-        longitude={city.longitude} latitude={city.latitude} >
-        <CityPin size={20} />
+      <Marker key={`marker-${index}`}
+        longitude={city.longitude}
+        latitude={city.latitude} >
+        <CityPin size={20} onClick={() => this.setState({popupInfo: city})} />
       </Marker>
+    );
+  }
+
+  _renderPopup() {
+    const {popupInfo} = this.state;
+
+    return popupInfo && (
+      <Popup tipSize={5}
+        anchor="top"
+        longitude={popupInfo.longitude}
+        latitude={popupInfo.latitude}
+        onClose={() => this.setState({popupInfo: null})} >
+        {popupInfo.cityName}
+      </Popup>
     );
   }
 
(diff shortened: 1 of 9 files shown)


examples/controls/app.js
```diff
@@ -1,13 +1,14 @@
...
 import {render} from 'react-dom';
-import MapGL, {Marker} from 'react-map-gl';
+import MapGL, {Marker, Popup} from 'react-map-gl';
 
 import CityPin from './components/city-pin';
 
 import CITIES from './data/cities.json';
 
 import './app.css';
+import 'mapbox-gl/dist/mapbox-gl.css';
```
abmai: Echoing the other diff, this only works for webpack? We should add a `note` here saying that if that's the case and we want to continue using this method.

src/components/popup.js
```diff
@@ -0,0 +1,157 @@
...
+// all copies or substantial portions of the Software.
+
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+import {Component, createElement} from 'react';
```
abmai: Let's use `PureComponent` instead. We're asking for React >= `15.4.x` anyway.

src/utils/dynamic-position.js
```diff
@@ -0,0 +1,78 @@
...
+  anchor,
+  padding = 0
+}) {
+  let {x: anchorX, y: anchorY} = ANCHOR_POSITION[anchor];
+
+  let top = y - anchorY * selfHeight;
+  let bottom = top + selfHeight;
+  const yStep = 0.5;
+
+  if (top < padding) {
```
abmai: There's a ton of logic and loops in this function, can we add comments along the way, maybe at every `if` clause to guide the reader on what's happening in terms of the UI (e.g. `prevent exceeding bottom edge` or something less terrible :p)?

src/utils/dynamic-position.js
```diff
@@ -0,0 +1,78 @@
...
+      left += xStep * selfWidth;
+    }
+  } else if (right > width - padding) {
+    while (right > width - padding && anchorX <= 1 - xStep) {
+      anchorX += xStep;
+      right -= xStep * selfWidth;
+    }
+  }
+
+  return Object.keys(ANCHOR_POSITION).find(positionType => {
```
abmai: Nit: parens around single param function declaration please

examples/controls/app.js
Pessimistress: I also added style-loader and css-loader to the webpack config.
Browserify has a css transform as well.

src/components/popup.js
Pessimistress: PureComponent doesn't update when context changes. The only way is to implement our own `shouldComponentUpdate` here. Considering the parent component (`StaticMap` or `InteractiveMap`) is pure, and map re-render is almost always triggered by a viewport change, we almost definitely need to re-position this component as well.

abmai: Ahh ok

examples/controls/app.js
Pessimistress: I've updated the doc to cover different ways to include the Mapbox style.

src/components/popup.js
ibgreen: > PureComponent doesn't update when context changes. 

Might be worth adding a comment for maintainers? Even experienced React programmers are not always familiar with the finer points of contexts.

src/components/marker.js
```diff
@@ -54,7 +54,7 @@ export default class Marker extends Component {
     };
 
     return createElement('div', {
-      className: 'map-control-marker',
+      className: 'mapboxgl-marker',
```
Pessimistress: This one is not dependent on the mapbox stylesheet because it's already in-lined (position: absolute). The change of class name is just to align with what's used in the other controls.
The Popup control on the other hand has ~150 lines of css. I'm not convinced that it's worth the effort to recreate it. Alternatively we could:
- Offer our own stylesheet built on top of the mapbox-gl one (users still need to import it)
- Embed the stylesheet in JavaScript (will be harder for users to override)

PR #228, opened 2017-05-05
Attribution Control and Navigation Control

Pessimistress:
- Fix the `attributionControl` prop on `StaticMap`
- Add `NavigationControl` (equivalent of Mapbox's [NavigationControl](https://www.mapbox.com/mapbox-gl-js/api/#navigationcontrol))

### DIFF
--- a/docs/controls/navigation-control.md
+++ b/docs/controls/navigation-control.md
@@ -1,0 +1,22 @@
+# Navigation Control
+
+This is a React equivalent of Mapbox's [NavigationControl](https://www.mapbox.com/mapbox-gl-js/api/#navigationcontrol):
+
+```js
+import ReactMap, {Marker} from 'react-map-gl';
+
+export default MyMap = ({viewport, updateViewport}) => {
+  return (
+    <ReactMap {...viewport} onChangeViewport={updateViewport}>
+      <div style={{position: 'absolute', right: 0}}>
+        <NavigationControl onChangeViewport={updateViewport} />
+      </div>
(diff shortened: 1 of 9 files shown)


src/index.js
```diff
@@ -23,6 +23,7 @@ export {default as InteractiveMap} from './components/interactive-map';
 export {default as StaticMap} from './components/static-map';
 export {default as MapControls} from './components/map-controls';
 export {default as Marker} from './components/marker';
+export {default as NavigationControl} from './components/navigation-control';
```
ibgreen: I'd put these additional components in their own section below, before the overlays. I guess I see them as "extras".

docs/controls/navigation-control.md
```diff
@@ -0,0 +1,25 @@
+# Navigation Control
+
+This is a React equivalent of Mapbox's [NavigationControl](https://www.mapbox.com/mapbox-gl-js/api/#navigationcontrol):
+
+```js
+import ReactMap, {Marker} from 'react-map-gl';
+
+/* Needs Mapbox's stylesheet to work! */
+import 'mapbox-gl/dist/mapbox-gl.css';
```
ibgreen: Just realized that this only works in webpack. Our example source should not be bundler dependent.

Pessimistress: You will need to add configs to either bundler if you want this to work
I added style-loader and css-loader to the webpack config in the example app.
Browserify has a css transform as well.
I will add more comments here.

PR #229, opened 2017-05-05
Map components props

Pessimistress:
- Remove `onChangeViewport` prop from `StaticMap`
- Add missing prop checks/documentation to `InterativeMap`
- Add new prop `ControllerClass` to `InterativeMap` that allows advanced user to replace the default map controlls

### DIFF
--- a/src/components/interactive-map.js
+++ b/src/components/interactive-map.js
@@ -4,18 +4,61 @@ import autobind from '../utils/autobind';
 
 import StaticMap from './static-map';
 import MapControls from './map-controls';
-import MapState from '../utils/map-state';
+import MapState, {MAPBOX_MAX_PITCH, MAPBOX_MAX_ZOOM} from '../utils/map-state';
 import {PerspectiveMercatorViewport} from 'viewport-mercator-project';
 
 const propTypes = Object.assign({}, StaticMap.propTypes, {
-  displayConstraints: PropTypes.object.isRequired
+  // Additional props on top of StaticMap
+
+  /** Viewport constraints */
+  // Max zoom level
+  maxZoom: PropTypes.number,
+  // Min zoom level
+  minZoom: PropTypes.number,
+  // Max pitch in degrees
+  maxPitch: PropTypes.number,
+  // Min pitch in degrees
+  minPitch: PropTypes.number,
+
+  /** Interaction states, required to calculate change during transform
+   * These should be managed by the map controller and updated via the
+   * `onChangeViewport` callback
+   */
+  startPanLngLat: PropTypes.arrayOf(PropTypes.number),
+  startZoomLngLat: PropTypes.arrayOf(PropTypes.number),
+  startBearing: PropTypes.number,
+  startPitch: PropTypes.number,
+  startZoom: PropTypes.number,
+
+  /**
+   * `onChangeViewport` callback is fired when the user interacted with the
+   * map. The object passed to the callback contains viewport properties
+   * such as `longitude`, `latitude`, `zoom` and additional interaction
+   * state information.
+   */
+  onChangeViewport: PropTypes.func,
+
+  /** Advanced features */
+  // Contraints for displaying the map. If not met, then the map is hidden.
+  displayConstraints: PropTypes.object.isRequired,
+  // A React component class definition to replace the default map controls
+  ControllerClass: PropTypes.func
 });
 
 const defaultProps = Object.assign({}, StaticMap.defaultProps, {
+  onChangeViewport: null,
+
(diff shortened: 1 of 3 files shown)


src/components/interactive-map.js
```diff
@@ -4,18 +4,61 @@ import autobind from '../utils/autobind';
...
+  minPitch: 0,
+
   displayConstraints: {
-    maxZoom: 20,
-    maxPitch: 60
-  }
+    maxZoom: MAPBOX_MAX_ZOOM,
+    maxPitch: MAPBOX_MAX_PITCH
+  },
+  ControllerClass: MapControls
```
abmai: Mm what's the use case of this here? I'm a bit wary of passing classes around as props. Is it possible to use a Higher-Order Component approach here instead?

src/components/interactive-map.js
```diff
@@ -4,18 +4,61 @@ import autobind from '../utils/autobind';
...
+  /** Viewport constraints */
+  // Max zoom level
+  maxZoom: PropTypes.number,
+  // Min zoom level
+  minZoom: PropTypes.number,
+  // Max pitch in degrees
+  maxPitch: PropTypes.number,
+  // Min pitch in degrees
+  minPitch: PropTypes.number,
+
```
ibgreen: Technically I believe the system supports constraints on all props. No value in listing them here, but add a comment to this effect.

src/components/interactive-map.js
```diff
@@ -4,18 +4,61 @@ import autobind from '../utils/autobind';
...
 const defaultProps = Object.assign({}, StaticMap.defaultProps, {
+  onChangeViewport: null,
+
+  /** Viewport constraints */
+  maxZoom: MAPBOX_MAX_ZOOM,
+  minZoom: 0,
+  maxPitch: MAPBOX_MAX_PITCH,
+  minPitch: 0,
+
   displayConstraints: {
```
ibgreen: We flattened out displayConstraints on the `propTypes` but not here?

src/components/interactive-map.js
ibgreen: I was considering whether a child would work. We could clone it and add the props we need so the app doesn't need to care.

src/utils/map-state.js
```diff
@@ -2,16 +2,16 @@ import {PerspectiveMercatorViewport} from 'viewport-mercator-project';
 import assert from 'assert';
 
 // MAPBOX LIMITS
-const MAX_PITCH = 60;
-const MAX_ZOOM = 20;
+export const MAPBOX_MAX_PITCH = 60;
+export const MAPBOX_MAX_ZOOM = 20;
```
ibgreen: Q: Is this still a MAPBOX default limit? I.e. is it reasonable to call it `MAPBOX_`?

Putting `MAPBOX` constants is ok, although one could argue that in many ways this class implements completely completely generic (while it implements a compatible interaction model, the critical mapbox synchronization is handled by the Viewport).

src/components/interactive-map.js
abmai: We can find a better solution later, then. Happy to merge this to unblock what's necessary.

src/components/interactive-map.js
Pessimistress: There's viewport constraints (prop limits) and there display constraints (show/hide map). The system in this component is for display constraints.

src/components/interactive-map.js
Pessimistress: The flattened props are viewport constraints for legacy reasons. I haven't give much thought to how display constraints should be defined.

src/utils/map-state.js
Pessimistress: These are limitations of Mapbox. These two numbers are only ever used as default prop values.

src/components/interactive-map.js
Pessimistress: Actually does anyone ever need to change displayConstraints?

src/components/interactive-map.js
Pessimistress: Theoretically controllers will always render a relatively positioned div. We could refactor MapControls to be a React-independent component. Let's meet some time next week to review all the APIs.

src/components/interactive-map.js
ibgreen: I'm up for a discussion.

src/utils/map-state.js
ibgreen: The MAX_ZOOM limit was raised to 40 a while ago.

Pessimistress: I believe satellite image is still at 20.

PR #239, opened 2017-05-14
Fix webpack examples link

G2Jose:
The current README links to a URL that no longer exists (https://github.com/uber/deck.gl/tree/master/exhibits) for examples on using react-map-gl with webpack 1. This PR fixes the link.

### DIFF
--- a/README.md
+++ b/README.md
@@ -24,10 +24,10 @@
 
 * `browserify` - react-map-gl is extensively tested with `browserify` and works without configuration.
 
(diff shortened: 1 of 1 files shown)


G2Jose:
Not sure if @CLAassistant is working correctly, I've signed the Contributor License Agreement. Clicking the 'recheck' link does not seem to do anything.

abmai:
Thanks, @G2Jose! I'm gonna give it a few more hours to see if the @CLAAssistant catches up.

Apercu:
Ah you committed from the github editor, that won't work. We should be good then, merging

PR #244, opened 2017-05-17
New EventManager

Pessimistress:
- Integrate @ericsoco 's EventManager
- Support mobile in main example app

Will continue the discussed refactor of `MapControl` in follow up PR.


src/components/map-controls.js
```diff
@@ -72,24 +71,46 @@ export default class MapControls extends PureComponent {
...
   }
 
+  componentWillUnmount() {
+    if (this._eventManager) {
+      // Must destroy because hammer adds event listeners to window
+      this._eventManager.destroy();
+    }
+  }
+
+  /* Event utils */
```
abmai: Let's add here what the `event` object that's being inputed to all these functions. I'm assuming it's something we get back from `hammer.js` and assuming that they're all the same `event` object?

src/components/map-controls.js
```diff
@@ -122,79 +143,98 @@ export default class MapControls extends PureComponent {
...
 
-  _onMouseDrag({pos, startPos, modifier}) {
-    if (!this.props.onChangeViewport) {
-      return;
-    }
-
-    if (modifier) {
-      this._onMouseRotate({pos, startPos});
+  _onDrag(event) {
+    if (this._isFunctionKeyPressed(event)) {
```
abmai: hm, maybe this this cleaner with a ternary?
```javascript
this._isFunctionKeyPressed(event) ? this._onRotate(event) : this._onPan(event);
```
up to you though

src/components/map-controls.js
```diff
@@ -122,79 +143,98 @@ export default class MapControls extends PureComponent {
...
+  _onRotate(event) {
+    if (!this.props.perspectiveEnabled) {
       return;
     }
 
-    const xDelta = pos[0] - startPos[0];
-    const yDelta = pos[1] - startPos[1];
+    const {deltaX, deltaY} = event;
+    const pos = this._getCenter(event);
+    const startPos = [pos[0] - deltaX, pos[1] - deltaY];
```
abmai: nit (take or leave): is it worth going further here and have semantic names for all of these, e.g.
```javascript
const [centerX, centerY] = this._getCenter(event);
const [startX, startY] = [centerX - deltaX, centerY - deltaY];
```

src/components/map-controls.js
```diff
@@ -122,79 +143,98 @@ export default class MapControls extends PureComponent {
...
 
-    const xDeltaScale = xDelta / this.props.width;
+    const xDeltaScale = deltaX / this.props.width;
```
abmai: nit: since we're switching to having X/Y at the end, let's get these guys to: `deltaScaleX, deltaScaleY`

src/components/map-controls.js
```diff
@@ -122,79 +143,98 @@ export default class MapControls extends PureComponent {
...
-  _onMouseUp() {
-    const newMapState = this.props.mapState.panEnd().rotateEnd();
-    this._updateViewport(newMapState, {isDragging: false});
-  }
-
-  _onWheel({pos, delta}) {
+  _onWheel(event) {
+    const pos = this._getCenter(event);
+    const {delta} = event;
     let scale = 2 / (1 + Math.exp(-Math.abs(delta * ZOOM_ACCEL)));
```
abmai: quick comment on what is being calculated here?

src/utils/event-manager/event-manager.js
```diff
@@ -0,0 +1,97 @@
...
+ * Delegates event registration and handling to Hammer.js.
+ */
+export default class EventManager {
+  constructor(element, options = {}) {
+    // TODO: support overriding default RECOGNIZERS by passing
+    // recognizers / configs, keyed to event name.
+
+    this._onBasicInput = this._onBasicInput.bind(this);
+    this.manager = new Manager(element, {recognizers: RECOGNIZERS})
+    .on('hammer.input', this._onBasicInput);
```
abmai: nit: indent

src/utils/event-manager/wheel-input.js
```diff
@@ -0,0 +1,126 @@
...
+    this.callback({
+      center: pos,
+      delta,
+      srcEvent,
+      target: this.element
+    });
+  }
+
+  _setState(settings) {
+    Object.assign(this._state, settings);
```
abmai: should this be `Object.assign({}, this._state, settings)`?

src/utils/event-manager/wheel-input.js
```diff
@@ -0,0 +1,126 @@
...
+  handler(event) {
+
+    event.preventDefault();
+    let value = event.deltaY;
+    // Firefox doubles the values on retina screens...
+    if (firefox && event.deltaMode === window.WheelEvent.DOM_DELTA_PIXEL) {
+      value /= window.devicePixelRatio;
+    }
+    if (event.deltaMode === window.WheelEvent.DOM_DELTA_LINE) {
+      value *= 40;
```
abmai: a few of these numbers are used again, e.g. `40` and `4` - can we extract them to constants at the top? it'll also give more context to why these specific values were chosen

src/components/map-controls.js
ericsoco: We (I) should document the event object format generated by event-manager (hammer.js). It varies somewhat by event type but there are consistent commonalities too.

src/components/map-controls.js
```diff
@@ -72,24 +71,46 @@ export default class MapControls extends PureComponent {
...
-      onTouchRotate: this._onTouchRotate,
-      onWheel: this._onWheel,
-      onWheelEnd: this._onWheelEnd,
-      mapTouchToMouse: true
+    // Register event handlers
+    const {canvas} = this.refs;
+
+    this._eventManager = new EventManager(canvas)
+    .on({
+      panstart: this._onDragStart,
```
ericsoco: Why are these handlers called `onDrag*` instead of `onPan*`?

src/components/map-controls.js
ericsoco: `(this._isFunctionKeyPressed(event) ? this._onRotate : this._onPan)(event);` FTW ;)

src/utils/event-manager/event-manager.js
```diff
@@ -0,0 +1,97 @@
...
+ * and gestural input (e.g. 'click', 'tap', 'panstart').
+ * Delegates event registration and handling to Hammer.js.
+ */
+export default class EventManager {
+  constructor(element, options = {}) {
+    // TODO: support overriding default RECOGNIZERS by passing
+    // recognizers / configs, keyed to event name.
+
+    this._onBasicInput = this._onBasicInput.bind(this);
+    this.manager = new Manager(element, {recognizers: RECOGNIZERS})
```
ericsoco: Hunh, since you don't need `mousemove` you can just work completely around the whole `inputClass` thing. Good for you.

src/utils/event-manager/event-manager.js
```diff
@@ -0,0 +1,97 @@
...
+    .on('hammer.input', this._onBasicInput);
+
+    // Handle mouse wheel events as well
+    this.wheelInput = new WheelInput(element, event => {
+      this.manager.emit('wheel', event);
+    });
+  }
+
+  destroy() {
```
ericsoco: Thanks for adding this.

src/utils/event-manager/event-manager.js
```diff
@@ -0,0 +1,97 @@
...
+      const recognizerName = EVENT_RECOGNIZER_MAP[event];
+      if (recognizerName) {
+        // Enable recognizer for this event.
+        this.manager.get(recognizerName).set({enable: true});
+      }
+
+      // Alias to a recognized gesture as necessary.
+      const gestureEventAlias = GESTURE_EVENT_ALIASES[event];
+      // Register event handler.
+      this.manager.on(gestureEventAlias || event, handler);
```
ericsoco: This will fire `'click'` event handlers with `'tap'` events. The `gestureAliases` in my `event-manager` was intended to ensure that event object types passed to handlers match the event name they were registered against. Your solution is cleaner, but that event name mismatch might be an issue for e.g. an event handler that switches on event type.

src/utils/event-manager/wheel-input.js
```diff
@@ -0,0 +1,126 @@
...
+  window.navigator.userAgent.toLowerCase() : '';
+const firefox = ua.indexOf('firefox') !== -1;
+
+const WHEEL_EVENTS = [
+  // Chrome, Safari
+  'wheel',
+  // IE
+  'mousewheel',
+  // legacy Firefox
+  'DOMMouseScroll'
```
ericsoco: Ugh.

src/utils/event-manager/wheel-input.js
ericsoco: It's private state so it shouldn't matter. Considering mouse wheel events can fire very rapidly, avoiding the additional allocation is nice if we can get away with it.

ericsoco:
Good to see the `event-manager` was almost plug-and-play. Feels like we're getting close to something general-purpose. My only beef is with your removal of `gestureAliases` / `_wrapAliasedGestureHandler`.

src/components/map-controls.js
Pessimistress: The properties used here are all from the hammer.js [event object](http://hammerjs.github.io/api/#event-object). Will add to comment.

src/utils/event-manager/event-manager.js
Pessimistress: I see. This piece had a bug of firing the same event multiple times. I'll bring the wrapper back and try to resolve it.

PR #245, opened 2017-05-18
MapControls Refactor

Pessimistress:
- Remove React component `MapControls`, merge its event management into `InteractiveMap`
- Add ES6 class `MapControls`
- Remove `ControllerClass` prop from `InteractiveMap`
- Add `mapControls` prop to `InteractiveMap`


src/index.js
```diff
@@ -35,6 +34,7 @@ export {default as NavigationControl} from './components/navigation-control';
 
 // Utilities
 // export {default as fitBounds} from './utils/fit-bounds';
+export {default as MapControls} from './utils/map-controls';
```
ibgreen: There was an intention not to export the new internal event management so that we could have a bit more time to let the new event handling APIs "bake" before we freeze them.

Any strong reason to include this in the export list? Is this export necessary to use `react-map-gl`?

Without this export, I's be happy to approve this PR pretty much as is. With this line in place, we should have at least add a documentation page for the MapControls class and I'd want to take a more thorough look at the API.

src/components/interactive-map.js
```diff
@@ -78,11 +100,57 @@ export default class InteractiveMap extends PureComponent {
...
+
+  componentWillUnmount() {
+    if (this._eventManager) {
+      // Must destroy because hammer adds event listeners to window
+      this._eventManager.destroy();
+    }
+  }
+
+  _handleEvent(event) {
+    const {mapControls} = this.props;
```
abmai: hm, there's an odd circular reference here? `mapControls` is inside `this.props` but its state is also `this.props`?

src/components/interactive-map.js
```diff
@@ -116,12 +184,19 @@ export default class InteractiveMap extends PureComponent {
       overflow: 'hidden'
     };
 
+    const eventCanvasStyle = Object.assign({}, {
```
abmai: what's the reason for `Object.assign` here? You're not merging any existing objects right?

src/utils/map-controls.js
```diff
@@ -0,0 +1,207 @@
...
+  }
+
+  /* Callback util */
+  // formats map state and invokes callback function
+  updateViewport(mapState, extraState = {}) {
+    if (!this.onChangeViewport) {
+      return false;
+    }
+
+    return this.onChangeViewport(Object.assign(
```
ibgreen: Worth checking if nothing has changed? Perhaps not even a possibility?

src/utils/map-controls.js
```diff
@@ -0,0 +1,207 @@
...
+    this.onChangeViewport = onChangeViewport;
+    this.isDragging = isDragging;
+    this.perspectiveEnabled = perspectiveEnabled;
+  }
+
+  /**
+   * Callback for events
+   * @param {hammer.Event} event
+   */
+  handle(event) {
```
ibgreen: There is something here that gives off an odd vibe. This class is clearly written around handling hammer specific events, and yet it doesn't handle the registration of those events.  So this knowledge of the event manager becomes spread out between the components. Perhaps it is the right call, but it seems like an important thing to make clear.

Again if MapControls is not exported, I think we could land this now and take another look at it before we make the class public.

src/components/interactive-map.js
Pessimistress: Indeed... though `mapControls` only extracts states that it recognizes.
The scenario I'm trying to accommodate is custom callbacks (say I need a custom map control that has callbacks `onMove`, `onZoom` etc.)

src/index.js
Pessimistress: We need to extend this class for internal use cases. I will add a doc page.

src/utils/map-controls.js
Pessimistress: I'm in no rush to land this, let's make it right.

src/components/interactive-map.js
abmai: Got it. Let's add a comment here about that.

docs/advanced/custom-map-controls.md
```diff
@@ -0,0 +1,66 @@
+# Custom Map Controls
+
+## Overriding The Default Map Controller
+
+To change the default behavior of map interaction, you can implement/extend the `MapControls`
```
ericsoco: Might want to be explicit that any custom implementation (that does not directly extend MapControls) must implement the API as documented below.

src/index.js
ericsoco: That's the reason for the whole refactor, no? ;)

src/utils/map-controls.js
ericsoco: Instead of switching on event type, can `MapControls` expose a map of its event names -> handlers for `InteractiveMap` to pass through to `EventManager`? Or, alternatively,`InteractiveMap` could pass a reference to `EventManager` into `MapControls` for event registration.

ericsoco: Agreed with @ibgreen that this doesn't quite feel right. `InteractiveMap` owns both `EventManager` and `MapControls`, but (with the exception of instantiation and destruction) it doesn't really do anything with `EventManager`.

I think the intent here is for both `EventManager` and `MapControls` to be exposed for custom implementation. However:
- `EventManager` is not specified as a prop on `InteractiveMap` so it can't currently be overridden / extended;
- `InteractiveMap` probably doesn't need to care about `EventManager` at all, with the exception of exposing it to application developers. So it could be passed through to `MapControls`; we'd then also need a `destroy()` hook on `MapControls` to destroy the `EventManager` as well.

I'd like to see this straightened out before landing...think it's almost there.

shaojingli: > I think the intent here is for both EventManager and MapControls to be exposed for custom implementation.

I noticed this part too but feel that we could go with current solution. It's very hard for me to imagine that users would want to reimplement low-level EventManager since even we ourselves don't really want to do it for various reasons like cross-platform compatibility, etc... 

Of course, we could make EventManager a prop to all Control classes and provide a default value but we really should guard against "flexibility over everything" mindset. 

Anyway, this should be discussed and decided when we eventually decide to export it

src/index.js
shaojingli: Thanks @ibgreen for catching this. Yes, if we'd like to export MapControls, we should at lease implement it as a subclass of Controls so that user can extend the Controls class to create OrbitControls, FreeCameraControls, FPSControls, etc....

And I also think MapControls should really be MapViewControl or MapViewportControl

PR #134, opened 2016-10-09
Provide prop for simple disabling of 'scroll zoom' behaviour

cammanderson:
This merge will allow a new prop on MapGL "scrollZoomDisabled" boolean. It's default value is 'false', meaning the current behaviour is preserved. 

When the scrollZoomDisabled prop is set to 'true', the mouse scroll will no longer zoom in/out. 

This addresses a usability issue when you wish to scroll down the page in a web browser, but the react-map-gl MapGL component captures the scroll events.

Mapbox GL JS provides an API option to disable the zoom scroll handler. This merge will give the user the flexibility to disable this also.

PR #246, opened 2017-05-18
[2.0-release] add prop to disable scroll zoom

abmai:
Following what was done in #134 by @cammanderson but switched it to be more consistent with the `perspectiveEnabled` pattern.

This PR is for 2.0-release and feature should be discussed for inclusion in v3.

### DIFF
--- a/src/map-interactions.react.js
+++ b/src/map-interactions.react.js
@@ -93,7 +93,8 @@ export default class Interactions extends Component {
     onTouchEnd: PropTypes.func,
     onTouchTap: PropTypes.func,
     onZoom: PropTypes.func,
-    onZoomEnd: PropTypes.func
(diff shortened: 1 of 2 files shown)


ibgreen:
Didn't notice this was for 2.0-release but I assume the intention would be to merge forward to 3.0?

shaojingli:
@abmai For this and other PRs that are intended for a release branch, please indicate explicitly that "this PR is for 2.0-release and will be cherry-picked to 3.0" in the description.

abmai:
Yeah sorry this was needed internally. I should've clarified that it was for v2. Let's definitely add this to v3 at some point because it's a useful flag to toggle, but the method will have to depend on where we land with the new `MapControls` stuff over there.

abmai:
Updated description here as well as for #246

PR #251, opened 2017-05-23
Api Audit

Pessimistress:
- Add `visibility` prop to `StaticMap` for showing/hiding the map
- Add `onLoad` prop to both components, called when the MapboxGL instance is loaded
- Rename `_getMap` method to `getMap`
- Add `queryRenderedFeatures` method that exposes the MapboxGL API with the same name
- Remove all interactivity related props from `StaticMap`
- Remove intermediate state props from `InteractiveMap`: `startPanLngLat`, `startZoomLngLat`, `startBearing`, `startPitch`, `startZoom`. These are stored in the map controls.
- `InteractiveMap` is now stateful (`isDragging` and `isHover`)
- Rename `onClickFeatures` and `onHoverFeatures` to `onClick` and `onHover`. Remove `ignoreEmptyFeatures` prop. The callbacks are invoked with an event object with `features` and `lngLat` fields.
- New `getCursor` prop of `InteractiveMap`: returns cursor style from the current map state
- Rename `displayConstraints` to `visibilityConstraints`
- Remove `perspectiveEnabled` prop. Add `scrollZoom`, `dragPan`, `dragRotate`, `doubleClickZoom`, `touchZoomRotate` props.

### DIFF
--- a/examples/custom-interactions/view.js
+++ b/examples/custom-interactions/view.js
@@ -80,9 +80,7 @@ export default class Example extends Component {
         longitude: location.longitude,
         zoom: 11,
         bearing: 180,
-        pitch: 60,
-        startDragLngLat: null,
-        isDragging: false
+        pitch: 60
       },
       mapStyle: buildStyle({stroke: '#FF00FF', fill: 'green'})
     };
@@ -107,8 +105,8 @@ export default class Example extends Component {
     this.setState({viewport: opt});
   }
 
-  _onClickFeatures(features) {
-    window.console.log(features);
+  _onClickFeatures(event) {
+    window.console.log(event.features);
   }
 
   render() {
@@ -122,8 +120,7 @@ export default class Example extends Component {
         { ...viewport }
         maxPitch={85}
         onChangeViewport={ this._onChangeViewport }
-        onClickFeatures={ this._onClickFeatures }
-        perspectiveEnabled={ true }
+        onClick={ this._onClickFeatures }
         // setting to `true` should cause the map to flicker because all sources
         // and layers need to be reloaded without diffing enabled.
         preventStyleDiffing={ false }>
--- a/examples/custom-interactions/view.js
+++ b/examples/custom-interactions/view.js
@@ -105,10 +105,6 @@ export default class Example extends Component {
     this.setState({viewport: opt});
   }
 
-  _onMapLoaded(event) {
-    window.console.log(event);
-  }
-
   _onClickFeatures(event) {
     window.console.log(event.features);
   }
@@ -124,7 +120,6 @@ export default class Example extends Component {
         { ...viewport }
         maxPitch={85}
         onChangeViewport={ this._onChangeViewport }
-        onLoad={ this._onMapLoaded }
         onClick={ this._onClickFeatures }
         // setting to `true` should cause the map to flicker because all sources
         // and layers need to be reloaded without diffing enabled.
--- a/examples/custom-interactions/view.js
+++ b/examples/custom-interactions/view.js
@@ -105,6 +105,10 @@ export default class Example extends Component {
     this.setState({viewport: opt});
   }
 
+  _onMapLoaded(event) {
+    window.console.log(event);
+  }
+
   _onClickFeatures(event) {
     window.console.log(event.features);
   }
@@ -120,6 +124,7 @@ export default class Example extends Component {
         { ...viewport }
         maxPitch={85}
         onChangeViewport={ this._onChangeViewport }
+        onLoad={ this._onMapLoaded }
         onClick={ this._onClickFeatures }
         // setting to `true` should cause the map to flicker because all sources
         // and layers need to be reloaded without diffing enabled.
--- a/src/components/interactive-map.js
+++ b/src/components/interactive-map.js
@@ -32,21 +32,24 @@ const propTypes = Object.assign({}, StaticMap.propTypes, {
   onChangeViewport: PropTypes.func,
 
   /** Enables control event handling */
+  // Scroll to zoom
   scrollZoomEnabled: PropTypes.bool,
+  // Drag to pan
   dragPanEnabled: PropTypes.bool,
+  // Drag to rotate
   dragRotateEnabled: PropTypes.bool,
+  // Double click to zoom
   doubleClickZoomEnabled: PropTypes.bool,
+  // Pinch to zoom / rotate
   touchZoomRotateEnabled: PropTypes.bool,
 
  /**
     * Called when a feature is hovered over. Uses Mapbox's
     * queryRenderedFeatures API to find features under the pointer:
     * https://www.mapbox.com/mapbox-gl-js/api/#Map#queryRenderedFeatures
-    * To query only some of the layers, set the `interactive` property in the
+    * To make a layer interactive, set the `interactive` property in the
     * layer style to `true`. See Mapbox's style spec
     * https://www.mapbox.com/mapbox-gl-style-spec/#layer-interactive
-    * If no interactive layers are found (e.g. using Mapbox's default styles),
-    * will fall back to query all layers.
     * @callback
     * @param {array} features - The array of features the mouse is over.
     */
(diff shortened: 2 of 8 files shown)


src/components/interactive-map.js
```diff
@@ -41,41 +31,86 @@ const propTypes = Object.assign({}, StaticMap.propTypes, {
...
+  onHover: null,
+
+  scrollZoomEnabled: true,
+  dragPanEnabled: true,
+  dragRotateEnabled: true,
+  doubleClickZoomEnabled: true,
+  touchZoomRotateEnabled: true,
+
+  clickRadius: 0,
+  getCursor: ({isDragging, isHovering}) => isDragging ? config.CURSOR.GRABBING :
```
abmai: nit: define this function at the top of the file and use it down here?

src/components/interactive-map.js
```diff
@@ -121,102 +168,125 @@ export default class InteractiveMap extends PureComponent {
...
   _handleEvent(event) {
-    const {mapControls} = this.props;
-    // MapControls only extracts the states that it recognizes.
-    // This allows custom map controls to add new states and callbacks.
-    mapControls.setState(Object.assign({}, this.props, {
-      mapState: new MapState(this.props)
-    }));
-
-    return mapControls.handle(event);
+    const controlOptions = Object.assign({
```
abmai: Hm, was this done specifically so that `onChangeState` could be overridden from props?

Pessimistress: Not really... the `onChangeState` callback is specifically introduced to pull the `isDragging` state from the map control since it affects rendering.

src/components/static-map.js
```diff
@@ -232,24 +189,35 @@ export default class StaticMap extends PureComponent {
   }
 
   // External apps can access map this way
-  _getMap() {
+  getMap() {
```
abmai: Hm actually. I see that we added an `onLoad` prop to here. Does it make sense to have both a `getMap()` method and an `onLoad` method? Can we just return the map instance for `onLoad` as well? I feel that it'll be more react-like that way, rather than exposing an external method.

examples/custom-interactions/view.js
```diff
@@ -107,8 +105,12 @@ export default class Example extends Component {
     this.setState({viewport: opt});
   }
 
-  _onClickFeatures(features) {
-    window.console.log(features);
+  _onMapLoaded(event) {
```
ibgreen: Is `event` a good name here?

src/components/interactive-map.js
```diff
@@ -3,7 +3,7 @@ import PropTypes from 'prop-types';
 import autobind from '../utils/autobind';
 
 import StaticMap from './static-map';
-import MapState, {MAPBOX_MAX_PITCH, MAPBOX_MAX_ZOOM} from '../utils/map-state';
+import {MAPBOX_MAX_PITCH, MAPBOX_MAX_ZOOM} from '../utils/map-state';
```
ibgreen: Q: should we make a namespace so the user doesn't have to import a "laundry list" of these 

`MAPBOX_LIMITS.MAX_PITCH` etc?

src/components/interactive-map.js
```diff
@@ -23,16 +23,6 @@ const propTypes = Object.assign({}, StaticMap.propTypes, {
...
-   * These should be managed by the map controller and updated via the
-   * `onChangeViewport` callback
-   */
-  startPanLngLat: PropTypes.arrayOf(PropTypes.number),
-  startZoomLngLat: PropTypes.arrayOf(PropTypes.number),
-  startBearing: PropTypes.number,
-  startPitch: PropTypes.number,
-  startZoom: PropTypes.number,
-
   /**
```
ibgreen: Now that we have separate markdown docs (which are the source of truth) we can cut and keep the comments here minimal, one liner or better inline comments.

src/components/interactive-map.js
```diff
@@ -41,41 +31,86 @@ const propTypes = Object.assign({}, StaticMap.propTypes, {
...
+  dragPanEnabled: PropTypes.bool,
+  // Drag to rotate
+  dragRotateEnabled: PropTypes.bool,
+  // Double click to zoom
+  doubleClickZoomEnabled: PropTypes.bool,
+  // Pinch to zoom / rotate
+  touchZoomRotateEnabled: PropTypes.bool,
+
+ /**
+    * Called when the map is hovered over.
```
ibgreen: Again, I'd make sure the markdown is correct and cut the duplicated docs here.

src/components/interactive-map.js
```diff
@@ -41,41 +31,86 @@ const propTypes = Object.assign({}, StaticMap.propTypes, {
...
+  /** Accessor that returns a cursor style to show interactive state */
+  getCursor: PropTypes.func,
 
   /** Advanced features */
   // Contraints for displaying the map. If not met, then the map is hidden.
-  displayConstraints: PropTypes.object.isRequired,
+  // Experimental! May be changed in minor version updates.
+  visibilityConstraints: PropTypes.object.isRequired,
   // A map control instance to replace the default map controls
   // The object must expose one property: `events` as an array of subscribed
```
ibgreen: Would love to see this moved to markdown and expanded upon. Semantics are not yet fully clear to me.

src/components/interactive-map.js
```diff
@@ -41,41 +31,86 @@ const propTypes = Object.assign({}, StaticMap.propTypes, {
    */
   onChangeViewport: PropTypes.func,
 
-  /** Enables perspective control event handling */
-  perspectiveEnabled: PropTypes.bool,
```
ibgreen: This will be a pretty big backwards compatibility breaker. Should we consider having a section of deprecated props at the bottom. At least check and warn?

src/components/interactive-map.js
```diff
@@ -101,16 +143,21 @@ export default class InteractiveMap extends PureComponent {
...
     const events = {};
     mapControls.events.forEach(eventName => {
       events[eventName] = this._handleEvent;
     });
 
-    this._eventManager = new EventManager(eventCanvas, {events});
+    const eventManager = new EventManager(eventCanvas, {events});
+
+    // Register event handlers for click and hover
+    eventManager.on('mousemove', this._onMouseMove);
```
ibgreen: Will this override these events, or add additional events? Could be good to comment for those of use who are not `EventManager` experts.

If override, I guess we could just use Object.assign on the `{events}`.

src/components/interactive-map.js
```diff
@@ -121,102 +168,125 @@ export default class InteractiveMap extends PureComponent {
...
-    // This allows custom map controls to add new states and callbacks.
-    mapControls.setState(Object.assign({}, this.props, {
-      mapState: new MapState(this.props)
-    }));
-
-    return mapControls.handle(event);
+    const controlOptions = Object.assign({
+      onChangeState: this._onInteractiveStateChange
+    }, this.props);
+    return this.props.mapControls.handle(event, controlOptions);
```
ibgreen: `handle` looks rather ambiguous here. I suspect that `handleEvent` would read better?

src/components/interactive-map.js
```diff
@@ -121,102 +168,125 @@ export default class InteractiveMap extends PureComponent {
...
-      if (maxPropName in displayConstraints && props[propName] > displayConstraints[maxPropName]) {
+      if (maxPropName in visibilityConstraints &&
+        props[propName] > visibilityConstraints[maxPropName]) {
         return false;
       }
     }
     return true;
   }
 
+  _getFeatures({pos, radius}) {
```
ibgreen: As stated would like another round of discussion before we move model event handling out of the model class. The remaining changes are fine.

src/components/interactive-map.js
```diff
@@ -121,102 +168,125 @@ export default class InteractiveMap extends PureComponent {
...
+        const viewport = new PerspectiveMercatorViewport(this.props);
+        event.lngLat = viewport.unproject(pos);
+        event.features = features;
+
+        this.props.onHover(event);
+      }
+    }
+  }
+
+  _onMouseClick(event) {
```
ibgreen: Blocking on this

examples/custom-interactions/view.js
Pessimistress: This is the `event` object generated by Mapbox's event system

src/components/interactive-map.js
Pessimistress: These are additional events.

PR #252, opened 2017-05-24
Fix EventManager import under node

Pessimistress:
@ericsoco 's EventManager fix from https://github.com/uber/deck.gl/pull/658


src/utils/event-manager/event-manager.js
```diff
@@ -1,13 +1,30 @@
-import {Manager} from 'hammerjs';
+import WheelInput from './wheel-input';
+import MoveInput from './move-input';
+import isBrowser from '../is-browser';
+
+// Hammer.js directly references `document` and `window`,
```
ibgreen: Nit: I think we could get the gist across with a little less text.

PR #255, opened 2017-05-25
copy and serve documentation markdown from ./docs folder

abmai:
With this, we can now continue to edit `/docs` separately without having to use the `/static` folder in here for that. However, we'll still have to modify the structure in `/constants/pages` manually.

CopyWebpackPlugin will basically copies over contents of `./docs` to `./doc-pages/dist/docs` to serve publicly during build. It does the same to webpack's in-memory filesystem during development.


doc-pages/src/static/docs/upgrade-guide.md
```diff
@@ -1,116 +0,0 @@
-# Upgrade Guide
-
-## Upgrading from deck.gl v3 to v4
```
ibgreen: I would argue that we do want an upgrade guide so I'd leave a place holder here and in the index.

doc-pages/src/static/docs/whats-new.md
```diff
@@ -1,350 +0,0 @@
-# deck.gl v4.0
```
ibgreen: Same here, we have a lot of new features to present to our users. It would also follow the pattern from our other frameworks.

doc-pages/webpack/config.js
```diff
@@ -1,9 +1,6 @@
-const {resolve, join} = require('path');
+const {resolve} = require('path');
 const webpack = require('webpack');
-
-const rootDir = join(__dirname, '../..');
-const demoDir = join(__dirname, '..');
-const demoSources = join(demoDir, 'src');
+const CopyWebpackPlugin = require('copy-webpack-plugin');
 
 // Otherwise modules imported from outside this directory does not compile
```
ibgreen: Wonder if this is still an issue? This setup looks pretty unusual.

doc-pages/webpack/config.js
```diff
@@ -72,8 +60,19 @@ module.exports = {
 
   plugins: [
     new webpack.DefinePlugin({
-      MAPBOX_ACCESS_TOKEN: `"${process.env.MAPBOX_ACCESS_TOKEN}"` // eslint-disable-line
-    })
+      // eslint-disable-next-line
+      MAPBOX_ACCESS_TOKEN: `"${process.env.MAPBOX_ACCESS_TOKEN}"`
+    }),
+    new CopyWebpackPlugin([
```
ibgreen: Nice!

doc-pages/webpack/config.js
abmai: No idea. This is a carry-over from the deck.gl stuff. It was pulling in files from the main `/src` to use as examples. If we agree to install from `npm` instead (like a regular app) then this won't be a problem. We can tweak this as the docs become more finished.

doc-pages/.eslintrc
```diff
@@ -0,0 +1,18 @@
+{
```
abmai: Planning on unifying things for the entire code base if possible - this is to clean this codebase up first.

PR #256, opened 2017-05-25
Add `fitBounds` util

Pessimistress:
- Bring in `fitBounds` from v2 and remove mapbox dependency
- Tests

### DIFF
--- a/src/utils/fit-bounds.js
+++ b/src/utils/fit-bounds.js
@@ -18,16 +18,14 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-// NOTE: Transform is not a public API so we should be careful to always lock
-// down mapbox-gl to a specific major, minor, and patch version.
 import {PerspectiveMercatorViewport} from 'viewport-mercator-project';
 
 /**
  * Returns map settings {latitude, longitude, zoom}
  * that will contain the provided corners within the provided
  * width.
- * @param {Number} width - viewport width
- * @param {Number} height - viewport height
+ * @param {Number} viewport.width - viewport width
+ * @param {Number} viewport.height - viewport height
  * @param {Array} bounds - [[lon, lat], [lon, lat]]
  * @param {Number} options.padding - The amount of padding in pixels to add to the given bounds.
  * @param {Array} options.offset - The center of the given bounds relative to the map's center,
@@ -35,9 +33,7 @@ import {PerspectiveMercatorViewport} from 'viewport-mercator-project';
  * @returns {Object} - latitude, longitude and zoom
  */
 export default function fitBounds(
-  width,
-  height,
-  // bounds
+  viewportProps,
   bounds,
   // options
   {
@@ -45,6 +41,7 @@ export default function fitBounds(
     offset = [0, 0]
   } = {}
 ) {
+  const {width, height} = viewportProps;
   const [[west, south], [east, north]] = bounds;
 
   const viewport = new PerspectiveMercatorViewport({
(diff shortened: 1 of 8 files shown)


src/utils/fit-bounds.js
```diff
@@ -0,0 +1,80 @@
...
+/**
+ * Returns map settings {latitude, longitude, zoom}
+ * that will contain the provided corners within the provided
+ * width.
+ * @param {Number} width - viewport width
+ * @param {Number} height - viewport height
+ * @param {Array} bounds - [[lon, lat], [lon, lat]]
+ * @param {Number} options.padding - The amount of padding in pixels to add to the given bounds.
+ * @param {Array} options.offset - The center of the given bounds relative to the map's center,
+ *    [x, y] measured in pixels.
```
ibgreen: Can we support perspective with this approach? Do we need extra params to specify pitch and bearing?

src/utils/fit-bounds.js
```diff
@@ -0,0 +1,80 @@
...
+
+// THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+// IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+// FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+// AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+// LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+// OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+// THE SOFTWARE.
+
+// NOTE: Transform is not a public API so we should be careful to always lock
```
abmai: remove comment?

src/utils/fit-bounds.js
```diff
@@ -0,0 +1,80 @@
...
+ * width.
+ * @param {Number} width - viewport width
+ * @param {Number} height - viewport height
+ * @param {Array} bounds - [[lon, lat], [lon, lat]]
+ * @param {Number} options.padding - The amount of padding in pixels to add to the given bounds.
+ * @param {Array} options.offset - The center of the given bounds relative to the map's center,
+ *    [x, y] measured in pixels.
+ * @returns {Object} - latitude, longitude and zoom
+ */
+export default function fitBounds(
```
abmai: Should we move over to an object params approach, for more clarity? i.e.
```js
function fitBounds({
  width,
  height,
  bounds,
  options = {
    padding = 0,
    offset = [0, 0]  
  }
} = {})
```

test/utils/fit-bounds.spec.js
```diff
@@ -1,20 +1,72 @@
 import test from 'tape-catch';
 import {fitBounds} from 'react-map-gl';
 
+/* fitBounds from Mapbox for comparison */
+import Transform from 'mapbox-gl/js/geo/transform';
```
abmai: Does this introduce dependency on versions of mapbox that still export this? We're pretty behind with their public release.

test/utils/fit-bounds.spec.js
```diff
@@ -1,20 +1,72 @@
...
+import {LngLatBounds, Point} from 'mapbox-gl';
```
abmai: Same here.

test/utils/fit-bounds.spec.js
Pessimistress: Possible, but since this is only used for testing, the module itself won't be affected.

src/utils/fit-bounds.js
Pessimistress: I haven't given it much thought, but it might be tricky since with perspective we won't be able to calculate pixel size of the bounding box like this any more.
In case we do want to support it in the future, maybe it's better to change the api to something more generic like `fitBounds(viewport, bounds, options)`?
@abmai

abmai: `viewport` would be an object, `bounds` an array, and `options` another object? If that's what you recommend then I'm in favor of adopting it.

PR #258, opened 2017-05-26
Fix flickering bug when drag over popups

Pessimistress:
Problem: hammer.js uses client coordinates to recognize gestures, which is the expected behavior. When we calculate the pointer position relative to the map, we used to assume `event.target` is the container. However, if an overlay element is interactive, when dragging the pointer over it, it becomes the `event.target`.

This conversion is also needed in deck.gl. Should we handle it inside the EventManager instead?

### DIFF
--- a/src/components/interactive-map.js
+++ b/src/components/interactive-map.js
@@ -184,6 +184,13 @@ export default class InteractiveMap extends PureComponent {
     const controlOptions = Object.assign({}, this.props, {
       onChangeState: this._onInteractiveStateChange
     });
+    /*
+     * The map control needs `event.target` to calculate pointer positions
+     * relative to the map.
+     * If an overlay is interactive, when dragging over it, it becomes the
+     * `event.target`. Overwrite it with the event canvas here.
+     * The original target element is accessible in `event.srcEvent`.
+     */
     event.target = this.refs.eventCanvas;
     return this.props.mapControls.handleEvent(event, controlOptions);
   }
--- a/src/components/interactive-map.js
+++ b/src/components/interactive-map.js
@@ -184,14 +184,6 @@ export default class InteractiveMap extends PureComponent {
     const controlOptions = Object.assign({}, this.props, {
       onChangeState: this._onInteractiveStateChange
     });
-    /*
-     * The map control needs `event.target` to calculate pointer positions
-     * relative to the map.
-     * If an overlay is interactive, when dragging over it, it becomes the
-     * `event.target`. Overwrite it with the event canvas here.
-     * The original target element is accessible in `event.srcEvent`.
-     */
-    event.target = this.refs.eventCanvas;
     return this.props.mapControls.handleEvent(event, controlOptions);
   }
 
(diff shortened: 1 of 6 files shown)


src/components/interactive-map.js
```diff
@@ -184,6 +184,14 @@ export default class InteractiveMap extends PureComponent {
...
       onChangeState: this._onInteractiveStateChange
     });
+    /*
+     * The map control needs `event.target` to calculate pointer positions
+     * relative to the map.
+     * If an overlay is interactive, when dragging over it, it becomes the
+     * `event.target`. Overwrite it with the event canvas here.
+     * The original target element is accessible in `event.srcEvent`.
+     */
+    event.target = this.refs.eventCanvas;
```
abmai: Overriding a field in the raw event feels odd to me. Happy to go with this for now, but maybe we should pass the `eventCanvas` down explicitly as well if that's what needed universally anyway?

ibgreen: > Overriding a field in the raw event feels odd to me. Happy to go with this for now, but maybe we should pass the eventCanvas down explicitly as well if that's what needed universally anyway?

Yes - it feels like the wrong place to do something so "low level", if we modify the event it should be done by EventManager as @Pessimistress suggested. Do we have all the information we need to do this in EventManager (access to the canvas etc)?

ericsoco:
If I'm understanding correctly, the EventManager change would be similar to what I did when implementing [drag events in deck.gl](https://github.com/uber/deck.gl/blob/master/src/lib/draw-and-pick.js#L83-L88) -- lock `event.target` at the start of a drag (on mousedown) and maintain it through the rest of the interaction lifecycle.

In order to do this, we'll also have to maintain state separately for each touch, to properly support multitouch.

So, it's doable, but non-trivial. Note that what @Pessimistress is modifying here is not the "raw event", it's the wrapper created by hammer.js. So I'm a bit more ok with it here...but is there a chance that this will interfere with other application event handlers? I'm not familiar enough with `react-map-gl` to know how non-map events are handled, but I'm guessing they don't rely on EventManager.

Pessimistress:
Per review feedback, I removed the hack from InteractiveMap and added a new field `srcElement` to EventManager's event object.

ericsoco:
I think `rootElement` (or similar) would be a better name, to avoid confusion with the unrelated `srcElement`.

[Here](https://github.com/hammerjs/hammer.js/blob/a28c530297f6e66785476013a6c9162305a2ce42/src/inputjs/compute-input-data.js#L64-L69) is where hammer.js decides what to report as `target` on the event wrapper -- if `srcEvent.target` is a descendant of the "root" element to which EventManager/Hammer is bound, it's used as `event.target`, else it falls back to that root element.

I don't fully understand your use case, though -- what's the DOM look like when you run into this condition? Since we're passing a `<canvas>` in as the "root" element, there can't be any descendants that would end up being reported as `event.target`.

Also of note: [`move-input` is already setting the event wrapper `target` to that "root" element](https://github.com/uber/deck.gl/blob/master/src/utils/events/move-input.js#L29).

Pessimistress:
@ericsoco I'm actually passing in a `div` to the event manager - it just happens to be called eventCanvas for legacy reasons. This `div` is the container of the map and all overlays.

abmai:
Agree with @ericsoco on the name, `rootElement` is probably clearer in this case because `src` could also be confusing.

Otherwise LGTM!

PR #259, opened 2017-05-26
[deprecate-start] onChangeViewport is now onViewportChange; add depWarn util

abmai:
* also use `depWarn` for other deprecated props


src/utils/dep-warn.js
```diff
@@ -0,0 +1,33 @@
...
+ */
+export default function checkDeprecatedProps(props = {}) {
+  /* eslint-disable no-console, no-undef */
+  DEPRECATED_PROPS.forEach((depProp) => {
+    if (props.hasOwnProperty(depProp.old)) {
+      let warnMessage = getDeprecatedText(depProp.old);
+      if (depProp.new) {
+        warnMessage = `${warnMessage} ${getNewText(depProp.new)}`;
+      }
+      console.warn(warnMessage);
```
ibgreen: We should only log each deprecation once to avoid generating a cascade. That should be easy to add here.

src/utils/dep-warn.js
```diff
@@ -0,0 +1,33 @@
...
+
+/**
+ * Checks props object for any prop that is deprecated and insert a console
+ * warning to the user. This will also print out the recommended new prop/API
+ * if one exists.
+ */
+export default function checkDeprecatedProps(props = {}) {
+  /* eslint-disable no-console, no-undef */
+  DEPRECATED_PROPS.forEach((depProp) => {
+    if (props.hasOwnProperty(depProp.old)) {
```
ibgreen: Is `hasOwnProperty` needed?

src/components/navigation-control.js
```diff
@@ -40,7 +44,9 @@ export default class NavigationControl extends Component {
   _updateViewport(opts) {
     const {viewport} = this.context;
     const mapState = new MapState(Object.assign({}, viewport, opts));
-    this.props.onChangeViewport(mapState.getViewportProps());
+    // TODO(deprecate): remove this check when `onChangeViewport` gets deprecated
+    const onViewportChange = this.props.onChangeViewport || this.props.onViewportChange;
```
ibgreen: Since we already have a table with prop mappings in dep-warn, couldn't we have a function that uses that table and applies the value of old props if new ones are not supplied, instead of duplicating that logic here?

src/utils/dep-warn.js
```diff
@@ -0,0 +1,33 @@
+// 'new' is optional
```
ibgreen: Nit: about the name of the file/function, it feels too abbreviated. We usually type things out.

Also, if we add the function suggested above (to resolve values using the old props if new ones are not provided) then maybe this file does not need to named after the function. It could be `deprecation-utils.js`

src/components/interactive-map.js
```diff
@@ -24,11 +25,11 @@ const propTypes = Object.assign({}, StaticMap.propTypes, {
   minPitch: PropTypes.number,
 
   /**
-   * `onChangeViewport` callback is fired when the user interacted with the
+   * `onViewportChange` callback is fired when the user interacted with the
```
ibgreen: Since we are going to change this prop, let's give the name a second round of audit. E.g. is `onViewportChange` the best choice? I have a feeling I will still find it hard to remember exactly without looking up in the docs. How about just calling it `onChange`? Would that be confusing?

src/utils/map-controls.js
```diff
@@ -108,10 +108,10 @@ export default class MapControls {
     const oldViewport = this.mapState.getViewportProps();
     const newViewport = newMapState.getViewportProps();
 
-    if (this.onChangeViewport &&
+    if (this.onViewportChange &&
```
ibgreen: Q: Should we even try to keep the names identical here? This seems like a rather different class...

This code has changed a lot so I might not have the full picture. If my comment does not make sense, just ignore it.

src/utils/map-controls.js
```diff
@@ -121,16 +121,19 @@ export default class MapControls {
    * Extract interactivity options
    */
   setOptions({
+    // TODO(deprecate): remove this when `onChangeViewport` gets deprecated
```
ibgreen: This is a new internal class in v3 right? No need to worry about deprecations?

src/utils/map-controls.js
Pessimistress: This is the same prop. The MapControls class is the one firing the callback in the current design. Idea being a user can add their own callbacks (onMapMove, onMapRotate for example) by supplying their own control class.

src/utils/map-controls.js
Pessimistress: Same here - InteractiveMap delegate the handling of `onChangeViewport`, `dragRotate`, `scrollZoom` etc. to this class. Technically these props are not owned by InteractiveMap, and may be ignored if a custom controller is supplied.

src/utils/dep-warn.js
abmai: It's the best way to check if an object has a property that it owns (vs. inherited). We'd only want to log a warning if the user is trying to use an old prop imo.

src/utils/dep-warn.js
abmai: This is a general utility, the "once" log is handled through calling this in the `constructor` of the components. Or do you think there should only be 1 deprecation warning app-wide? I feel like we could be slightly annoying about this but happy to do it either way.

src/components/navigation-control.js
abmai: I wanted to make it as easy as possible to move ahead with the duplication later on (i.e. in this case all you'd have to do is delete half of the conditional here) vs. having to also get rid of an import and change code.

abmai: To further expand, this is code that should be eventually removed from the codebase. Making it ugly now would prioritize its removal, I think.

I'm happy to use a util as well though if you feel strongly about doing it that way.

src/components/interactive-map.js
abmai: I think `onChange` is quite generic in this context. I can't think of anything more specific than `onViewportChange` but will be happy to change it if we come up with something!

ericsoco: speaking as someone who admittedly hasn't used `react-map-gl` much but _has_ used other mapping frameworks, `onViewportChange` or `onViewportChanged` should work. Other frameworks generally provide a higher level of granularity, like `onZoom`, `onPan`, but I think `viewport` as one overall concept works.

PR #260, opened 2017-05-29
bump mapbox-gl to 0.37.0; remove postinstall.js and postinstall script

abmai:
Will also have to update tests that required importing `mapbox-gl` util functions.

Tested with a few apps and `postinstall` is no longer needed (hooray). Nothing also broke by going up to `0.37.0`.


test/utils/fit-bounds.spec.js
```diff
@@ -2,45 +2,45 @@ import test from 'tape-catch';
 import {fitBounds} from 'react-map-gl';
 
 /* fitBounds from Mapbox for comparison */
-import Transform from 'mapbox-gl/js/geo/transform';
-import {LngLatBounds, Point} from 'mapbox-gl';
-import {toLowPrecision} from '../test-utils';
+// import Transform from 'mapbox-gl/js/geo/transform';
```
abmai: How do we proceed here? Just create the expected result object manually and compare?

ibgreen:
> Will also have to update tests that required importing mapbox-gl util functions.

While it certainly doesn't hurt to have some fixed test cases, it would seem reasonable to move fitBounds into viewport-mercator-project which:
1) contains all the projection logic for react-map-gl
2) already uses an older fixed version of mapbox for correctness tests.
3) would be the right place for updating `fitBounds` to handle complete perspective support.

abmai:
> While it certainly doesn't hurt to have some fixed test cases, it would seem reasonable to move fitBounds into viewport-mercator-project which:

Happy to let that happen as well!

PR #264, opened 2017-06-02
Clarifying webpack 2 example location in README.md

pranspach:
Master has webpack 2 example: exhibit-webpack

The `dev` branch pointer sent me in the wrong direction when trying to get this working with webpack 2. 

This is a small change, but should help people orient themselves when starting development with react-map-gl.


abmai:
Thanks for the PR, @pranspach!

Issue #267, opened 2017-06-05
v3 - onLoad gone?


contra:
Seems like this doesn't work anymore - I see the test for it still on master, but the code has no reference to the load event anymore.

ibgreen:
@contra Good finding. While we haven't concluded the final API reviews, I do believe the intention so far is to keep supporting this callback in v3.

pranspach:
if you use master, it's added back @contra. IDK when the releases are tagged for this project.

abmai:
I'll cut a new alpha with this today/tomorrow.

abmai:
Publish `3.0.0-alpha.13`.

PR #269, opened 2017-06-06
Add onLoad callback to static map props and componentDidMount

pranspach:
This PR is in reference to #267 

The current test in `components/map.spec.js` seems to only test that an onLoad prop doesn't interfere with initialization, since InteractiveMap.supported returns false. I'm likely missing something relevant to unit testing this project.

The code is mostly pulled from the previous 2.0 releases. I'd prefer to get the unit test working before merging in, but I wanted some eyes from you guys before venturing down that path. 

Like issue #267 references, is onLoad planned for v3 support? Please let me know if this PR is either unwarranted or veers away from v3 plans @ibgreen.

### DIFF
--- a/src/components/static-map.js
+++ b/src/components/static-map.js
@@ -145,7 +145,7 @@ export default class StaticMap extends PureComponent {
     }
 
     // Attach optional onLoad function
-    map.once('load', () => this.props.onLoad());
+    map.once('load', this.props.onLoad);
 
     this._map = map;
     this._updateMapViewport({}, this.props);
--- a/src/components/static-map.js
+++ b/src/components/static-map.js
@@ -85,7 +85,8 @@ const defaultProps = {
   visible: true,
   bearing: 0,
   pitch: 0,
-  altitude: 1.5
+  altitude: 1.5,
+  onLoad: noop
 };
 
 const childContextTypes = {
@@ -144,9 +145,7 @@ export default class StaticMap extends PureComponent {
     }
 
     // Attach optional onLoad function
-    if (this.props.onLoad) {
-      map.once('load', () => this.props.onLoad());
-    }
+    map.once('load', () => this.props.onLoad());
 
     this._map = map;
     this._updateMapViewport({}, this.props);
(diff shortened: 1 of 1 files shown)


src/components/static-map.js
```diff
@@ -141,6 +143,11 @@ export default class StaticMap extends PureComponent {
       canvas.style.outline = 'none';
     }
 
+    // Attach optional onLoad function
+    if (this.props.onLoad) {
```
ibgreen: We could add a `noop` function in `defaultProps`, then we can call the callback here without an if statement.

pranspach:
I updated the property with `noop`. 

I haven't used that in my own projects yet, but it's nice to remove the extra if-checks. It doesn't guard against people passing in an undef/null property, but that's a trade-off. Thanks for the suggestion @ibgreen

src/components/static-map.js
```diff
@@ -141,6 +144,9 @@ export default class StaticMap extends PureComponent {
       canvas.style.outline = 'none';
     }
 
+    // Attach optional onLoad function
+    map.once('load', () => this.props.onLoad());
```
Pessimistress: Should we pass the Mapbox native event object to the callback?

abmai: I could be wrong, but it doesn't seem like Mapbox is passing back any param to the callback of this event. At least not that they documented anyway. https://www.mapbox.com/mapbox-gl-js/api/#map.event:load

pranspach: I was confused by the previous code wrapping onLoad too. From the previous implementation and unit test, not exposing the map object was desirable. Maybe this was a work around for a previous version?

It feels cleaner to do `map.once('load', this.props.onLoad);`. I was trying to retain original behavior with the wrapper. @Pessimistress @abmai ?

abmai: Agreed that it should be `map.once('load', this.props.onload);`.

pranspach:
Removed onLoad wrapping -- thanks @Pessimistress @abmai

abmai:
Awesome. Thank you!

PR #270, opened 2017-06-06
Add new examples

Pessimistress:
- Clean up `exhibit-webpack`
- Add `layers` example (show/hide layers, change color)
- Add `geojson` example (add/update custom data source)
- Add `interaction` example (event control toggles, zoom & pitch limits)


examples/exhibit-webpack/app.js
```diff
@@ -14,12 +15,27 @@ class Root extends Component {
...
+      height: 500
+    }
+  }
+
+  componentDidMount() {
+    window.addEventListener('resize', this._resize);
+    this._resize();
+  }
+
+  _resize = () => {
```
ibgreen: I personally think we should avoid non-es6 code usage and keep the exhibits as basic as possible

examples/exhibit-webpack/package.json
```diff
@@ -15,8 +16,7 @@
     "babel-preset-react": "^6.16.0",
     "babel-preset-stage-2": "^6.18.0",
     "webpack": "^2.4.0",
-    "webpack-dev-server": "^2.4.0",
-    "webworkify-webpack-dropin": "^1.1.9"
+    "webpack-dev-server": "^2.4.0"
```
ibgreen: I'd remove the stage-2 below here and in package.json

examples/exhibit-webpack/webpack.config.js
```diff
@@ -35,6 +35,5 @@ module.exports = {
   ]
 };
 
-// DELETE THIS LINE WHEN COPYING THIS EXAMPLE FOLDER OUTSIDE OF DECK.GL
-// It enables bundling against src in this repo rather than the installed version
-module.exports = require('../webpack.config.local')(module.exports);
+module.exports = process.env.NODE_ENV === 'development' ?
```
ibgreen: I think I see why you made this change, and I think it is a good call, however:
 
1. This (support for local development using an optionally injected config) is a common pattern across all our examples in all our repos. Good to give it some extra thought. If this is good change maybe we want to do everywhere, e.g. before the coordinated release? 
3. In other repos, we use the webpack `--env` parameter (e.g. `--env.local`) instead of environment variables. Should we decide which way is better and use it everywhere?
4. I can see that the comment needed changing but I don't think we should delete it completely. There is a note at the top of the file, but it may not easily be spotted from here. We could at least give the user a hint what this strange construct is doing and that it is OK for the user to delete it. My view is that nothing should be considered trivial in webpack config files, every single line should have a comment :)

examples/geojson/webpack.config.js
```diff
@@ -0,0 +1,39 @@
...
+const config = {
+  entry: {
+    app: resolve('./src/app.js')
+  },
+
+  devtool: 'source-map',
+
+  module: {
+    rules: [{
+      // Compile ES2015 using buble
```
ibgreen: buble vs babel, please check in all places!

examples/geojson/webpack.config.js
```diff
@@ -0,0 +1,39 @@
...
+  resolve: {
+    alias: {
+      // From mapbox-gl-js README. Required for non-browserify bundlers (e.g. webpack):
+      'mapbox-gl$': resolve('./node_modules/mapbox-gl/dist/mapbox-gl.js')
+    }
+  },
+
+  // Optional: Enables reading mapbox token from environment variable
+  plugins: [
+    new webpack.EnvironmentPlugin(['MAPBOX_ACCESS_TOKEN', 'MapboxAccessToken'])
```
ibgreen: Should we standardize on one of these? Seems weird to support both. I think the capitalized version was the original react-map-gl version.

examples/interaction/webpack.config.js
```diff
@@ -0,0 +1,39 @@
+// NOTE: To use this example standalone (e.g. outside of deck.gl repo)
+// delete the local development overrides at the bottom of this file
+
+// avoid destructuring for older Node version support
```
ibgreen: Node v4-5 ?

examples/example-map-style.json
```diff
@@ -0,0 +1,7607 @@
+{
```
ibgreen: Nit: This is pretty big at almost 8K lines, is that needed for the example? Could it be made a little smaller?

examples/exhibit-webpack/app.js
ibgreen: This comment is only about the exhibits, not about the examples in general.

examples/geojson/webpack.config.js
abmai: +1 to this. I'm happy to go with just `MapboxAccessToken`

examples/layers/src/app.js
```diff
@@ -0,0 +1,67 @@
+/* global window */
+import React, {Component} from 'react';
+import {render} from 'react-dom';
+import MapGL from 'react-map-gl';
+import StyleControls from './style-controls';
+
+const token = process.env.MAPBOX_ACCESS_TOKEN; // eslint-disable-line
```
abmai: Also check for `MapboxAccessToken` here? Since the `webpack` config is looking for both.

examples/example-map-style.json
Pessimistress: I can try...

PR #271, opened 2017-06-07
Update marker.md

aniquetahir:
Popup is imported instead or Marker


abmai:
Nice catch. Thank you!

PR #275, opened 2017-06-12
Update main examples to work with new react-map-gl; remove some old examples

### LINKED ISSUE #235: The main example is in a broken state
The main example is broken - the package.json has not been updated for quite a while, react-map-gl is still at 1.x.x. 

Running the example with the local edition of react-map-gl generates errors in the console.

abmai:
* Removed examples that depend on now-hidden `ScatterplotOverlay` and `DraggablePointsOverlay`. We can discuss on what to do here.
* Pulled in the `custom-interactions` example and used as hero example.
* With this, also updated the main `npm run start` to once again run the main example.

Fixes #235


examples/main/package.json
```diff
@@ -10,15 +10,17 @@
...
     "d3-random": "^1.0.2",
     "d3-scale": "^1.0.3",
+    "deck.gl": "^4.0.0-rc.3",
     "immutable": "^3.8.1",
+    "luma.gl": "^3.0.0",
     "react": "^15.4.1",
+    "react-autobind": "^1.0.6",
     "react-dom": "^15.4.1",
-    "react-map-gl": "^1.7.2"
+    "react-map-gl": "^3.0.0-alpha.13"
```
ibgreen: `^` doesn't work well for alpha releases, it will only install the specified version. `>=` is what we use in other repos for these situations.

examples/main/views/click.js
```diff
@@ -20,44 +20,37 @@
 
 import React, {Component} from 'react';
 import PropTypes from 'prop-types';
-import MapGL, {autobind} from 'react-map-gl';
+import {InteractiveMap, experimental} from 'react-map-gl';
 /* global window */
 
 const windowAlert = window.alert;
+const autobind = experimental.autobind;
```
ibgreen: I personally don't think autobind should be part of react-map-gl's experimental exports, and I don't think that our examples should be including any experimental exports (unless that is the topic of the example). We should write code in these examples that use standard coding practices. I'd probably just bind in the constructor to make things simple and standard ES6 compatible.

examples/main/views/click.js
```diff
@@ -20,44 +20,37 @@
 
 import React, {Component} from 'react';
 import PropTypes from 'prop-types';
-import MapGL, {autobind} from 'react-map-gl';
+import {InteractiveMap, experimental} from 'react-map-gl';
```
ibgreen: Don't we want to just use the default export and still call it `MapGL`?

examples/main/views/custom-interaction.js
```diff
@@ -18,11 +18,13 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-/* window */
 import React, {Component} from 'react';
 import PropTypes from 'prop-types';
-import MapGL, {autobind} from 'react-map-gl';
```
ibgreen: Ditto

examples/main/views/custom-interaction.js
```diff
@@ -62,7 +64,7 @@ const propTypes = {
   height: PropTypes.number.isRequired
 };
 
-export default class TiltExample extends Component {
+export default class Example extends Component {
```
ibgreen: Nit: I believe that react automatically sets display name based on class name, at least in unminified code this can be helpful during debugging...

examples/main/views/geodata-creator-example.js
```diff
@@ -1,136 +0,0 @@
-// Copyright (c) 2015 Uber Technologies, Inc.
```
ibgreen: Double check that this is moved to `examples/additional-overlays`

examples/main/views/route-overlay.js
```diff
@@ -21,12 +21,13 @@
 /* global window */
 import React, {Component} from 'react';
 import PropTypes from 'prop-types';
-import MapGL, {SVGOverlay, CanvasOverlay, autobind} from 'react-map-gl';
+import MapGL, {SVGOverlay, CanvasOverlay, experimental} from 'react-map-gl';
```
ibgreen: ditto

examples/main/views/scatterplot-overlay-example.js
```diff
@@ -1,87 +0,0 @@
-// Copyright (c) 2015 Uber Technologies, Inc.
```
ibgreen: ditto

examples/main/views/style-diffing.js
```diff
@@ -18,16 +18,17 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-/* window */
+/* global window */
 import React, {Component} from 'react';
 import PropTypes from 'prop-types';
-import MapGL, {autobind} from 'react-map-gl';
+import MapGL, {experimental} from 'react-map-gl';
```
ibgreen: ditto

examples/main/webpack.config.js
```diff
@@ -32,12 +32,6 @@ module.exports = {
   module: {
     rules: [
       {
-        enforce: 'pre',
```
ibgreen: Yay!

.gitignore
```diff
@@ -6,5 +6,6 @@ coverage
 npm-debug.log
 
 */**/yarn.lock
+*/**/package-lock.json
```
ibgreen: Fine. Just wondering: what is generating this file? Don't think I have seen it.

abmai: The new version of NPM, `npm 5`, automatically generates a lock file on install. We should discuss at some point if we want to throw all these lock files into `.gitignore`, including `yarn.lock` or pick one to keep.

ibgreen: The only reason I am including the root `yarn.lock` is that it triggers Travis CI to install using `yarn` which leads to faster CI cycles.

It is a pretty silly reason, if someone has a better way then I'd love to remove yarn.lock, all those versions of a 100+KB file just generate dead weight in the repository...

examples/main/views/custom-interaction.js
Pessimistress: Is there a reason this file is called custom-interaction? It's not using a custom controller class.

examples/main/views/custom-interaction.js
```diff
@@ -18,11 +18,13 @@
 // OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 // THE SOFTWARE.
 
-/* window */
 import React, {Component} from 'react';
 import PropTypes from 'prop-types';
-import MapGL, {autobind} from 'react-map-gl';
+import autobind from 'react-autobind';
+import {InteractiveMap} from 'react-map-gl';
+import DeckGL, {ArcLayer} from 'deck.gl';
```
Pessimistress: Do we want to pull in DeckGL into a basic example?

ibgreen: > Do we want to pull in DeckGL into a basic example?

Ideally not. That should probably be a special deck.gl example, or just a reference to deck.gl's examples.

What we need to make the point here is some kind of overlay component that remains visible after the map no longer displays due to pitch > 60. Maybe we can use your new markers - unless they are dependent on some map context that also disappears? 

Perhaps this example would be a good test case that all the new components support the non-visible map case?

Pessimistress: I use the markers component to achieve this in the `interaction` example. It doesn't require any external dependencies.

Pessimistress:
Update `package.json` and `webpack.config` to align with the new `exhibit-webpack`? (handling of local development, environment variable etc.)

examples/main/views/custom-interaction.js
abmai: this is copy & pasted from the `custom-interactions` example - so no idea why that was named `custom-interactions`. Happy to switch this to "MainExample" or "PerspectiveExample" or something like that though.

examples/main/views/geodata-creator-example.js
abmai: This is the example that tried to use the draggable-points-overlay. Do we need to provide a working example for that?

examples/main/views/scatterplot-overlay-example.js
abmai: This was the example. The actual over is inside `addtional-overlays`.

examples/main/views/custom-interaction.js
abmai: Will update this before landing.

abmai:
Which one is the latest set of `webpack.config` that we should use? The one inside `exhibit-webpack` folder? There's a few different `webpack.config` spreading around the repo, and we should consolidate.

PR #276, opened 2017-06-13
Update examples

Pessimistress:
- Update `controls` examples to use new template
- Add `viewport-animation` example


examples/controls/package.json
```diff
@@ -1,19 +1,28 @@
...
+    "start": "webpack-dev-server --progress --hot --port 3000 --open",
+    "start-local": "webpack-dev-server --env.local --progress --hot --port 3000 --open"
   },
-  "license": "MIT",
   "dependencies": {
+    "immutable": "^3.8.1",
     "react": "^15.4.1",
     "react-dom": "^15.4.1",
-    "react-map-gl": "^3.0.0-alpha.8"
+    "react-map-gl": "^3.0.0-alpha.12"
```
ibgreen: Use >= to ensure we get the latest alpha?

examples/controls/src/app.js
```diff
@@ -3,15 +3,11 @@ import React, {Component} from 'react';
...
 
-import './app.css';
-import 'mapbox-gl/dist/mapbox-gl.css';
-
-const token = process.env.MAPBOX_ACCESS_TOKEN || // eslint-disable-line
-  'Set MAPBOX_ACCESS_TOKEN environment variable or put your token here.';
+const token = process.env.MapboxAccessToken; // eslint-disable-line
 
 if (!token) {
   throw new Error('Please specify a valid mapbox token');
```
ibgreen: Couldn't we build this check into the MapGL component instead of repeating it in every example?

examples/controls/src/app.js
```diff
@@ -3,15 +3,11 @@ import React, {Component} from 'react';
...
 
-import CITIES from './data/cities.json';
+import CITIES from './cities.json';
 
-import './app.css';
-import 'mapbox-gl/dist/mapbox-gl.css';
-
-const token = process.env.MAPBOX_ACCESS_TOKEN || // eslint-disable-line
-  'Set MAPBOX_ACCESS_TOKEN environment variable or put your token here.';
+const token = process.env.MapboxAccessToken; // eslint-disable-line
```
ibgreen: I thought the MapGL component already looked in this environment var. But probably good to explicitly show it. Maybe a comment would be enough?

examples/controls/webpack.config.js
```diff
@@ -1,63 +1,40 @@
...
-    new webpack.LoaderOptionsPlugin({minimize: false, debug: true})
+    new webpack.EnvironmentPlugin(['MapboxAccessToken'])
   ]
 };
 
-// DELETE THIS LINE WHEN COPYING THIS EXAMPLE FOLDER OUTSIDE OF DECK.GL
-// It enables bundling against src in this repo rather than the installed version
-module.exports = require('../webpack.config.local')(module.exports);
+// Enables bundling against src in this repo rather than the installed version
+module.exports = env => env && env.local ?
```
ibgreen: Isn't the env check was done inside the required file? Or is that only in our other repos?

examples/controls/package.json
```diff
@@ -1,19 +1,28 @@
 {
   "scripts": {
-    "start": "webpack-dev-server --progress --hot --open",
-    "start-local": "webpack-dev-server --env.local --progress --hot --open"
+    "start": "webpack-dev-server --progress --hot --port 3000 --open",
```
ibgreen: Interestingly I recently approved a PR (perhaps in one of our other repos?) that removed this port specification, with the argument that it would let webpack-dev-server dynamically pick ports when running several examples. I haven't tested this extensively though, not sure which is better.

Pessimistress: Looks like a bad copy & paste. I'll align it with the rest of the examples.

examples/controls/src/app.js
Pessimistress: No we don't.

examples/controls/src/app.js
Pessimistress: I believe the environment variable is replaced by webpack at compile time. Not sure how it will work without requiring the user to add additional configs to their bundling setup.

examples/controls/src/app.js
ibgreen: I should have provided the reference. here it is: https://github.com/uber/react-map-gl/blob/master/src/utils/access-token.js#L15

examples/controls/webpack.config.js
Pessimistress: If we always require `webpack.config.local` then webpack will fail when the user copy this folder outside of our directory.

examples/controls/src/app.js
ibgreen: Yes, the user needs to add a line to their webpack or browserify configurations, but they don't need anything in the application source code, if they use the standard environment variable supported by react-map-gl.

examples/controls/webpack.config.js
ibgreen: I see, since we are running in node.js wrapping a require in an "if" might actually avoid importing the file (in contrast to what happens when bundling). Nice!

PR #277, opened 2017-06-13
Fix compass arrow in navigation control

Pessimistress:
We have a dependency on the mapbox stylesheet, so bumping mapbox version could potentially break the controls.
Alternatively we could export a stylesheet in our library...

### DIFF
--- a/TESTING.md
+++ b/TESTING.md
@@ -1,0 +1,15 @@
+# Testing
+
+```
+npm run test
+npm run test-browser
+```
+
+## Bumping Mapbox Version
+
+Always pin Mapbox to a specific release.
+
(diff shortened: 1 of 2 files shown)


ibgreen:
> We have a dependency on the mapbox stylesheet, so bumping mapbox version could potentially break the controls. Alternatively we could export a stylesheet in our library...

We do pin the mapbox release, so this could only break when we bump, not whenever mapbox bumps. 

If we added a test protocol document and made a note  to always check that stylesheets still work when we bump mapbox, I think we would be fine as is. 

Could you add a TESTING.md with this bullet in it to this PR?

PR #280, opened 2017-06-15
Pass classes to mapbox-gl map as a prop

EDFward:


abmai:
Paint Classes are slated for deprecation, so we didn't want to support it natively.
https://www.mapbox.com/mapbox-gl-js/style-spec/#layer-paint.*

What you can do instead is grab the underlying `map` instance via `getMap()` which will give you full access to all of Mapbox's APIs. You can then use the `addClass()` `setClasses()` methods:
https://www.mapbox.com/mapbox-gl-js/api/#map#setclasses

EDFward:
that's super helpful! thanks!

PR #281, opened 2017-06-15
render main examples inside documentation

abmai:


doc-pages/src/components/examples.js
```diff
@@ -0,0 +1,14 @@
+import React, {Component} from 'react';
+import ExampleApp from '../../../examples/main/app';
+
+// TODO(abm): This is made into a wrapper to improve later
+
+export default class Example extends Component {
```
ibgreen: Not sure I understand what this does.. Are we not going to have a list of separate clickable examples as the other repos do?

abmai: Yeah I wanted to redo the examples at some point to make it more seemless (e.g. same doc style whether you're running it through the docs or locally).

PR #283, opened 2017-06-16
Update event manager

Pessimistress:
Copy and pasted from deck.gl

- Fix bug where page scroll is blocked even with `scrollZoom: false`


abmai:
Are there plans to split the event-manage into a separate module at some point?

ericsoco:
Whyever would you ask @abmai? ;)

Our plan is to let it bake across react-map-gl and deck.gl for a little while, until the API seems solid. We're almost there. Then we'll decide where it should live long-term.

abmai:
Is this good to land? @Pessimistress

Issue #274, opened 2017-06-10
Examples don't run without a root install


G2Jose:
I wonder if it makes sense to have a README section for links to projects that use react-map-gl.

I know I personally would have benefitted quite a bit from something like that since some of the examples provided (at least then) seemed a bit out of date.

That said, I can see some good reasons both in favor of and against having such a section. What do you folks think?

Pessimistress:
@G2Jose We definitely want to make it easier for developers to get started.
In the [examples](https://github.com/uber/react-map-gl/tree/master/examples) folder `layers` `interaction` `controls` and `geojson` are newly added apps that showcase capabilities of v3. We are also planning to upgrade/clean up the legacy examples in there soon.

ibgreen:
> I wonder if it makes sense to have a README section for links to projects that use react-map-gl.

@G2Jose I like this idea. Are you aware of any open source projects that would represent good examples?

abmai:
+1 to this. We can start a section if there are some good initial examples and will welcome PR to add more as well.

jpdevries:
I'm getting a "Cannot find webpack error" when trying to run the examples. I'm going into an example directory and 

```bash
npm install
npm run start
```

I still get the error even after `npm install webpack`

ibgreen:
@jpdevries Can you be a bit more specific? 

* Which specific example directory are you working with?
* Or are they all broken? 

I looked through the `package.json` files in all the examples and they all do list `webpack` as a `devDependency`,  nautrally with the exception of the `exhibit-browserify` example.

* Does your `npm install` fail to install webpack?
* if you do `npm ls webpack` what do you get?
* Is there anything in `node_modules/webpack`

jpdevries:
@ibgreen from the latest commit on master I tested going into each example directory. For example:

```bash
cd examples/main
npm install
npm run start
```

> Does your npm install fail to install webpack?

It seems to install it 

> if you do npm ls webpack what do you get?

```bash
j15:main jP$ npm ls webpack
/Users/jP/Sites/uber/react-map-gl/examples/main
├─┬ buble-loader@0.4.1
│ └── webpack@2.6.1  deduped
└── webpack@2.6.1 
```

> Is there anything in node_modules/webpack

Yes. 

```bash
j15:main jP$ ls node_modules/webpack
LICENSE		bin		hot		node_modules	schemas
README.md	buildin		lib		package.json	web_modules
```

I get this error in the console:

<details>
  <summary>Console Message</summary>
[...]

PR #285, opened 2017-06-20
Update to mapbox-gl-js v0.38.0 & fix examples

abmai:


examples/main/webpack.config.js
```diff
@@ -49,6 +49,6 @@ module.exports = {
   ]
 };
 
-// DELETE THIS LINE WHEN COPYING THIS EXAMPLE FOLDER OUTSIDE OF DECK.GL
-// It enables bundling against src in this repo rather than the installed version
-module.exports = require('../webpack.config.local')(module.exports);
+// Enables bundling against src in this repo rather than the installed version
+module.exports = (env) => env && env.local ?
```
ibgreen: Do this in all examples? Then we could close #274. I think it is just this and custom-interactions that need updates...

examples/custom-interactions/webpack.config.js
```diff
@@ -2,7 +2,7 @@
 const {resolve} = require('path');
```
Apercu: I think the destructuring has been avoided in other examples to support older versions of node, maybe copy it?

abmai: Mapbox `v0.38.0` enforces Node `v6.4` and higher. We're gonna start to be clear on the API here. I still need to update the README as part of this PR.

PR #294, opened 2017-06-28
Website and examples polish

Pessimistress:
- Add geojson example
- Add introduction, credit and link to source to all examples
- CSS bug fixes
- Remove unused website components
- Header image


examples/controls/src/app.js
```diff
@@ -33,8 +33,8 @@ export default class App extends Component {
         zoom: 3.5,
         bearing: 0,
         pitch: 0,
-        width: window.innerWidth,
-        height: window.innerHeight,
+        width: 500,
```
abmai: Why does this have to default to 500 and not window.innerWidth? Is it because of server-side render?

website/src/components/examples.js
```diff
@@ -32,15 +26,22 @@ export default class Examples extends Component {
     window.onresize = null;
   }
 
+  _getSize() {
+    const {innerWidth, innerHeight} = window;
+
+    return {
+      width: innerWidth >= 576 ? innerWidth - WIDTH_OFFSET : innerWidth,
```
abmai: What's significant about `576` here?

Pessimistress: We have a media selector in css for mobile. Not sure if there are better ways to manage this variable other than hard coding it.

examples/main/constants/toc.js
```diff
@@ -1,15 +1,17 @@
 // Modules
-import MarkerExample from '../views/marker';
-import NotInteractiveExample from '../views/not-interactive';
-import CustomOverlayExample from '../views/custom-overlay';
-import RouteOverlayExample from '../views/route-overlay';
-import StyleDiffingExample from '../views/style-diffing';
-import ClickExample from '../views/click';
+// import MarkerExample from '../views/marker';
```
abmai: This will actually affect the `main` example when ran locally as well. Can't we just filter out the specific part of the array when rendering these? Or export different chunks of it.

examples/controls/src/app.js
Pessimistress: No particular reason really... Some examples were using window size, some were using default props, and I just wanted to align them.

abmai: Let's stick to `window.innerWidth` `window.innerHeight` where we can then (or at least whatever the `resize` handler does), since in most cases they seem to get reseted to whatever the `resize` controller does anyway

website/src/components/examples.js
ibgreen: > Not sure if there are better ways to manage this variable other than hard coding it.

A reasonable way to deal with hard coded numbers is to assign them to constants with descriptive names. Either at the top of the file or directly in the function. That gives the next programmer to work on this a little more context about what is going on.

PR #295, opened 2017-06-28
documentation polish

abmai:


docs/get-started/README.md
```diff
@@ -9,7 +9,7 @@ that aim to make working with external overlays, such as
 
 # Installation
 
-`ReactMapGL` requires `node >= v4` and `react >= 15.4`.
+`ReactMapGL` requires `node >= v6.4.0` and `react >= 15.4`.
```
Pessimistress: My understanding is node 6 is only required for dev? Is it still true after build? (v1 worked with 0.10)

docs/README.md
```diff
@@ -1,21 +1,21 @@
...
 
-react-map-gl is a react friendly API wrapper around
+react-map-gl is a suite of React components for
 [Mapbox GL JS](https://github.com/mapbox/mapbox-gl-js).
 
 ## Example
 
 ```js
-import MapGL from 'react-map-gl';
+import ReactMapGL from 'react-map-gl';
```
ibgreen: Fine, but I personally never include the `React` part in any other react component I use.
I.e:
`import Button from 'react-button'`
rather than
`import ReactButton from 'react-button'`

docs/get-started/README.md
ibgreen: What is the dependency on 6.4 again? 6 is a long time release, requiring 6.4 is pretty demanding. It is not only for building react-map-gl right, it is required just for installing it?

docs/get-started/using-with-isomorphic-rendering.md
```diff
@@ -2,4 +2,4 @@
 
 Running under Node.js
 
-TBA
\ No newline at end of file
+TBA
```
ibgreen: Maybe delete this page, since we don't have anything to say

docs/upgrade-guide.md
```diff
@@ -71,6 +51,26 @@ const viewport = new PerspectiveMercatorViewport({width: 600, height: 400}).fitB
 // }
 ```
 
+### Two Map Components
+
+v3 now exposes two React components: `StaticMap` and `InteractiveMap`.
+`InteractiveMap` is the default export.
```
ibgreen: 'and designed to be as compatible as possible with the v2 default export'?

website/src/components/home.js
```diff
@@ -36,21 +36,22 @@ export default class Home extends Component {
...
+                Part of Uber Visualization's Framework Suite
               </h3>
               <p>
-              deck.gl is a great match with React, supporting
-              efficient WebGL rendering under the Reactive programming
-              paradigm. And when used with Mapbox GL it automatically
-              coordinates with the Mapbox camera system to provide
-              compelling 2D and 3D visualizations on top of your Mapbox
-              based maps.
+              Use together with <a
```
ibgreen: with "e.g."?

docs/get-started/README.md
abmai: Good point, we can test to see if this works on lower versions of Node when built.

docs/README.md
abmai: I like to not rename things from the package wherever possible, to prevent confusions later if you needed to `grep` for anything.

docs/get-started/README.md
abmai: Yeah I'll give this a try with a regular app.

PR #298, opened 2017-06-29
move babel configs from package.json to webpack config to fix doc build issues

abmai:
Looks like all the build issues resulted from having `babel` config inside of `package.json`. I don't know why.

But removing them from `package.json` and putting into `webpack.config.js` solved all the things. Individual examples still work as well.

Also ensured modules are resolved to the working directory first before the imported file's directory with `modules`.

The doc website should build great now, simply go into `website/`, then `yarn && npm run start` and voila.

### DIFF
--- a/examples/controls/webpack.config.js
+++ b/examples/controls/webpack.config.js
@@ -5,6 +5,16 @@
 const resolve = require('path').resolve;
 const webpack = require('webpack');
 
+const BABEL_CONFIG = {
+  presets: [
+    'es2015',
+    'react',
+    'stage-2'
+  ].map(function configMap(name) {
+    return require.resolve(`babel-preset-${name}`);
+  })
+};
+
 const config = {
   entry: {
     app: resolve('./src/root.js')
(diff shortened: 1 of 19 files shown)


examples/controls/webpack.config.js
```diff
@@ -5,6 +5,16 @@
 const resolve = require('path').resolve;
 const webpack = require('webpack');
 
+const BABEL_CONFIG = {
```
ibgreen: Wouldn't it be great if there was a comment in front of each line in our webpack configs that explained why a particular configuration is needed?

Particularly in this case it could save someone else a lot of time later.

examples/geojson-animation/webpack.config.js
```diff
@@ -5,6 +5,16 @@
 const resolve = require('path').resolve;
 const webpack = require('webpack');
 
+const BABEL_CONFIG = {
```
ibgreen: Why is this needed inside an example? I thought this was only needed when one directory referred to another outside its tree. The `examples` folder is intended to contain stand-alone examples.

examples/controls/webpack.config.js
abmai: Got it.

examples/geojson-animation/webpack.config.js
abmai: Yeah, so this is needed because it's a safer way to load the babel configuration. The reason why the build was failing was because we had babel configuration in `package.json`. I'm not sure exactly why that was causing things to fail, but plucking it out of there and putting it here resolved the issues.

ibgreen: But why just for this example. Why was it not necessary for any of the others? 

How is the webpack.config of this example even involved when we build the website?

PR #300, opened 2017-06-29
Add links to other libraries

Pessimistress:
image

### DIFF
--- a/website/src/components/header.js
+++ b/website/src/components/header.js
@@ -1,7 +1,18 @@
 import React, {Component} from 'react';
 import {Link} from 'react-router';
 
+import {LIBRARIES} from '../constants/links';
+
 export default class Header extends Component {
+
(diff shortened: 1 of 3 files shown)


website/src/components/header.js
```diff
@@ -1,16 +1,27 @@
 import React, {Component} from 'react';
 import {Link} from 'react-router';
 
+import {LIBRARIES} from '../constants/links';
```
ibgreen: This is great for now.

To simplify maintenance when we have this list in several repos, should we load a JSON file from a raw github link (we could store it in the deck.gl repo for now) with the list of links?

We'd use the constant list until we load the live list?

abmai:
Oh didn't realize this had opacity. Can we just go with solid color? It's a bit weird on top of the `react-map-gl` text right now.

Issue #163, opened 2016-12-21
Add zoom funcionality for mobile enhancement


SuiOni:
How is it possible to add a pinch to zoom functionality or any other zoom function for mobile?
Right now I can only zoom via mouse wheel.

I can find touch handlers in the source code but there is no documentation how to set them up for zoom.

An other option would be to add a navigation button.
So how can i add the native mapbox-gl zoom controls in a reactive way?

Right now I had to add

map.addControl(new mapboxgl.NavigationControl());
to the source code.
It works but the overlays is off by one zoom level and only update after zoom is finished, so its buggy.

Thank you guys for the awesome work!

ibgreen:
@SuiOni - Because of the stateless nature of react-map-gl, we can't / don't want to rely on mapbox built-in event handling. We would need to implement pinch to zoom and rotate ourselves.

It is basically a matter of checking the number of touch points in the touch events and handling touch drags differently when there are multiple fingers. It would take no more than about a page of code I think, not much. There are several react touch-/react gesture ... type components out there that can serve as inspiration. We'll get to it eventually but so far it has not been high enough on our list.

primozs:
I have modified mapbox TouchZoomRotate handler. Zoom works ok for me but rotate does not. If anybody have an idea.
```
import { window } from './globals';
import { mousePos } from './utils';
import { unprojectFromTransform, cloneTransform } from './transform';

const significantScaleThreshold = 0.15;
const significantRotateThreshold = 4;

function noop() {}

export default class TouchZoomRotateHandler {
  constructor(map, container, props = {}) {
    let {
      onTouchZoomRotate = noop
    } = props;

    this._map = map;
    this._container = container;
    this.callbacks = {
      onTouchZoomRotate
    };

    this.enable();
  }

  isEnabled() {
    return !!this._enabled;
  }

  disableRotation() {
    this._rotationDisabled = true;
  }

  enableRotation() {
    this._rotationDisabled = false;
  }

  enable() {
    if (this.isEnabled()) return;
    this._container.addEventListener('touchstart', this._onStart);
    this._enabled = true;
  }

  disable() {
    if (!this.isEnabled()) return;
    this._container.removeEventListener('touchstart', this._onStart);
    this._enabled = false;
  }

  _onStart = (e) => {
    if (e.touches.length !== 2) return;

    const p0 = mousePos(this._container, e.touches[0]);
    const p1 = mousePos(this._container, e.touches[1]);

[...]

PR #303, opened 2017-06-30
Add zoom funcionality for mobile enhancement 2.0.3

AonanLi:
add properties to configure map handlers
- touchRotateZoomHandler
- doubleClickZoomHandler
#163


abmai:
Thanks for the PR, @AonanLi! We're actually adding all of this functionalities in v3, which will be released soon. You can check out `3.0.0-alpha.15` for a preview of what's coming or explore the docs on master: https://github.com/uber/react-map-gl/tree/master/docs

grantga:
It looks touch doesn't work in the examples/interaction example when running beta3--I don't see an open issue for this. Is there a work around or different prop that should be passed?

AonanLi:
@grantga I was working on a project using react-map-gl 2.0.3 which doesnt support this doubleClickZoom. I knew these funtionalities will be added in v3. I just thought maybe some users need this funtionality without upgrading the library to v3.

Issue #307, opened 2017-07-06
Page still scrolling when scrollZoom: true and zooming with gesture in pad.


cuevash:
When the  scrollZoom: true the map works as expected.. but the event is also affecting the page and it does scrolling when it should not!

Trying with simple example: deck.gl/examples/trips 

where the render() method from the app.js is changed to the code below. This is only to add some space around the map to be able to see the problem when trying to zoom in the component.

```
render() {
    const {viewport, buildings, trips, time} = this.state;

    return (
      <div>
        <p>asdaksdaksdhasd</p>
        <p>asdaksdaksdhasd</p>
        <p>asdaksdaksdhasd</p>
        <p>asdaksdaksdhasd</p>
        <div><p>JHKJAHKLAJ; kAJhs AJs AKshak</p></div>
      <MapGL
        width="100" height="500" 
        {...viewport}
        mapStyle="mapbox://styles/mapbox/dark-v9"
        onViewportChange={this._onViewportChange.bind(this)}
        mapboxApiAccessToken={MAPBOX_TOKEN}>
        <DeckGLOverlay viewport={viewport}
          buildings={buildings}
          trips={trips}
          trailLength={180}
          time={time}
          />
      </MapGL>
              <p>asdaksdaksdhasd</p>
        <p>asdaksdaksdhasd</p>
        <p>asdaksdaksdhasd</p>
        <p>asdaksdaksdhasd</p>
        <div><p>JHKJAHKLAJ; kAJhs AJs AKshak</p></div>
      </div>
    );
  }
```

It is probably related to code #283

abmai:
@Pessimistress any insights here?

PR #308, opened 2017-07-07
Add/remove event handlers based on user setting

Pessimistress:
This is to fix https://github.com/uber/react-map-gl/issues/307
*Problem*
`EventManager.emit` is async, and therefore `event.preventDefault` may not be effective in handlers.
When scrollZoom is disabled, we need to unregister the event instead of filter it out in the handler.

*Changes*
- Disable recognizers in `EventManager.off` if no more handlers are attached
- Let `MapControls` manage its own events by passing in the `EventManager` instance

### DIFF
--- a/src/components/interactive-map.js
+++ b/src/components/interactive-map.js
@@ -112,9 +112,7 @@ const defaultProps = Object.assign({}, StaticMap.defaultProps, MAPBOX_LIMITS, {
   clickRadius: 0,
   getCursor: getDefaultCursor,
 
-  visibilityConstraints: MAPBOX_LIMITS,
-
-  mapControls: new MapControls()
+  visibilityConstraints: MAPBOX_LIMITS
 });
 
 const childContextTypes = {
@@ -140,6 +138,10 @@ export default class InteractiveMap extends PureComponent {
       // Whether the cursor is over a clickable feature
       isHovering: false
     };
+
+    // If props.mapControls is not provided, fallback to default MapControls instance
+    // Cannot use defaultProps here because it needs to be per map instance
+    this._mapControls = props.mapControls || new MapControls();
   }
 
   getChildContext() {
@@ -151,7 +153,6 @@ export default class InteractiveMap extends PureComponent {
 
   componentDidMount() {
     const {eventCanvas} = this.refs;
-    const {mapControls} = this.props;
 
     const eventManager = new EventManager(eventCanvas);
 
@@ -160,14 +161,14 @@ export default class InteractiveMap extends PureComponent {
     eventManager.on('click', this._onMouseClick);
     this._eventManager = eventManager;
 
-    mapControls.setOptions(Object.assign({}, this.props, {
+    this._mapControls.setOptions(Object.assign({}, this.props, {
       onStateChange: this._onInteractiveStateChange,
       eventManager
     }));
   }
 
   componentWillUpdate(nextProps) {
-    this.props.mapControls.setOptions(nextProps);
+    this._mapControls.setOptions(nextProps);
   }
 
   componentWillUnmount() {
--- a/src/components/interactive-map.js
+++ b/src/components/interactive-map.js
@@ -112,7 +112,9 @@ const defaultProps = Object.assign({}, StaticMap.defaultProps, MAPBOX_LIMITS, {
   clickRadius: 0,
   getCursor: getDefaultCursor,
 
-  visibilityConstraints: MAPBOX_LIMITS
+  visibilityConstraints: MAPBOX_LIMITS,
+
+  mapControls: new MapControls()
 });
 
 const childContextTypes = {
@@ -149,7 +151,7 @@ export default class InteractiveMap extends PureComponent {
 
   componentDidMount() {
     const {eventCanvas} = this.refs;
-    const mapControls = this.props.mapControls || new MapControls();
+    const {mapControls} = this.props;
 
     const eventManager = new EventManager(eventCanvas);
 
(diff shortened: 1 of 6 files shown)


src/utils/event-manager/constants.js
```diff
@@ -33,7 +33,6 @@ export const BASIC_EVENT_ALIASES = {
  * this block maps event names to the Recognizers required to detect the events.
  */
 export const EVENT_RECOGNIZER_MAP = {
-  click: 'tap',
```
ericsoco: Why did you remove the `click` mapping?

src/utils/event-manager/move-input.js
```diff
@@ -1,5 +1,5 @@
 const MOUSE_EVENTS = ['mousedown', 'mousemove', 'mouseup'];
-const MOVE_EVENT_TYPES = ['mousemove', 'pointermove'];
+const EVENT_TYPE = 'pointermove';
```
ericsoco: Is this basically a continuation of the work we did in EventManager to move everything over to only `pointer*` events?

src/utils/event-manager/move-input.js
```diff
@@ -35,9 +35,9 @@ export default class MoveInput {
    * Enable this input (begin processing events)
    * if the specified event type is among those handled by this input.
    */
-  enableIfEventSupported(eventType) {
-    if (MOVE_EVENT_TYPES.indexOf(eventType) >= 0) {
-      this.options.enable = true;
+  enableIfEventSupported(eventType, enabled) {
```
ericsoco: Should probably change this to `setEnabledIfEventSupported`.

src/utils/event-manager/wheel-input.js
```diff
@@ -64,6 +64,7 @@ export default class WheelInput {
     if (!this.options.enable) {
       return;
     }
+    event.preventDefault();
```
ericsoco: Didn't we intentionally remove this previously, for some reason?

ericsoco: Yeah, [here](https://github.com/uber/react-map-gl/pull/283).
> Fix bug where page scroll is blocked even with scrollZoom: false

Oh, is the idea that instead of blocking it, we're now conditionally adding the handler only when `scrollZoom` is `true`?

src/utils/map-controls.js
```diff
@@ -25,36 +25,32 @@ const PITCH_MOUSE_THRESHOLD = 5;
...
-  'panmove',
-  'panend',
-  'pinchstart',
-  'pinch',
-  'pinchend',
-  'doubletap',
-  'wheel'
-];
+const EVENT_TYPES = {
+  wheel: ['wheel'],
```
ericsoco: These keys should probably be ALL_CAPS.

src/utils/event-manager/constants.js
Pessimistress: I changed the recognizer look up to use the aliased event name (click -> tap) so this entry is no longer necessary.

src/utils/event-manager/move-input.js
Pessimistress: `mousemove` is aliased to `pointermove` so this extra event is never used.

src/utils/event-manager/wheel-input.js
Pessimistress: Calling `event.preventDefault` from the event handler for `wheel` event does not stop the default of the trackpad scroll gesture. There's a timing issue related to how event manager emits events.

ericsoco: Ok, got it. It's just wheel events that are async, correct? (Due to the `setTimeout` in there.)

src/components/interactive-map.js
```diff
@@ -151,20 +149,24 @@ export default class InteractiveMap extends PureComponent {
...
   componentDidMount() {
     const {eventCanvas} = this.refs;
-    const {mapControls} = this.props;
-
-    // Register event handlers defined by map controls
-    const events = {};
-    mapControls.events.forEach(eventName => {
-      events[eventName] = this._handleEvent;
-    });
+    const mapControls = this.props.mapControls || new MapControls();
```
abmai: Is this to explicitly guarding against `null`?

src/utils/event-manager/event-manager.js
```diff
@@ -107,27 +107,32 @@ export default class EventManager {
     }
   }
 
+  /*
+   * Enable/disable recognizer for the given event
+   */
+  _enableRecognizer(name, enabled) {
```
abmai: nit: `_toggleRecognizer` might be a better name since this is used for both enabling/disabling

src/utils/map-controls.js
abmai: +1

src/utils/map-controls.js
```diff
@@ -116,27 +112,55 @@ export default class MapControls {
...
+
+    // Interaction toggles
     this.scrollZoom = scrollZoom;
     this.dragPan = dragPan;
     this.dragRotate = dragRotate;
     this.doubleClickZoom = doubleClickZoom;
     this.touchZoomRotate = touchZoomRotate;
   }
 
+  enableEvents(eventNames, enabled) {
```
abmai: nit: `toggleEvents` instead?

src/components/interactive-map.js
Pessimistress: We used to set a default value for the prop `mapControls`. The change introduced by this PR makes the state of `mapControls` dependent on `eventManager` - when the component mounts, a new `eventManager` is created, and `mapControls` must be reset.

This new behavior is potentially introducing bugs if the user is not managing their custom `mapControls` instance properly. We used to call this prop `MapControls` which expects a constructor. Any suggestions?

src/utils/event-manager/wheel-input.js
Pessimistress: Seems that way. The `setTimeout` code originates from v1 and I don't really know its purpose.

PR #316, opened 2017-07-12
v2.0.x: Fixed PropTypes and refs for react v16.x.x

nwlieb:
Hi, this fixes the v2.0 for react version 16 by touching up deprecated APIs. A merge of this would be very helpful to us and It shouldn't be a breaking change in any way.

Thank you for the consideration! (CLA has been signed)

### DIFF
--- a/src/map-interactions.react.js
+++ b/src/map-interactions.react.js
@@ -20,7 +20,8 @@
 
 // Portions of the code below originally from:
 // https://github.com/mapbox/mapbox-gl-js/blob/master/js/ui/handler/scroll_zoom.js
-import React, {PropTypes, Component} from 'react';
+import React, {Component} from 'react';
+import PropTypes from 'prop-types';
 import autobind from 'autobind-decorator';
 import document from 'global/document';
 import window from 'global/window';
--- a/src/map-interactions.react.js
+++ b/src/map-interactions.react.js
@@ -124,15 +124,16 @@ export default class Interactions extends Component {
       pos: null,
       mouseWheelPos: null
     };
+    this._container = null;
   }
 
   _getMousePos(event) {
-    const el = this.container;
+    const el = this._container;
     return getMousePosition(el, event);
   }
 
   _getTouchPos(event) {
-    const el = this.container;
+    const el = this._container;
     const positions = getTouchPositions(el, event);
     return centroid(positions);
   }
@@ -325,7 +326,7 @@ export default class Interactions extends Component {
 
   @autobind
   _containerRefCallback(container) {
-    this.container = container;
+    this._container = container;
   }
 
   render() {
--- a/src/map-interactions.react.js
+++ b/src/map-interactions.react.js
@@ -126,12 +126,12 @@ export default class Interactions extends Component {
   }
 
   _getMousePos(event) {
-    const el = this.refs.container;
+    const el = this.container;
     return getMousePosition(el, event);
   }
 
   _getTouchPos(event) {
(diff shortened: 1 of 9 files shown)


src/map-interactions.react.js
```diff
@@ -325,7 +326,9 @@ export default class Interactions extends Component {
   render() {
     return (
       <div
-        ref="container"
+        ref={container => {
```
abmai: If we're moving to doing this, can we initialize `this.container = null;` in the `constructor` so the variables are clear.

src/map.react.js
```diff
@@ -253,7 +254,7 @@ export default class MapGL extends Component {
       this.props.mapStyle;
 
     const map = new mapboxgl.Map({
-      container: this.refs.mapboxMap,
+      container: this.mapboxMap,
```
abmai: ditto for this

src/overlays/canvas.react.js
```diff
@@ -64,7 +65,9 @@ export default class CanvasOverlay extends Component {
     const pixelRatio = window.devicePixelRatio || 1;
     return (
       <canvas
-        ref="overlay"
+        ref={overlay => {
```
abmai: ditto (and all the other places below)

src/map-interactions.react.js
Pessimistress: We recently had to remove ref callbacks in v3.0 (https://github.com/uber/react-map-gl/pull/310) because it will be called with null when one of the container's child components render with error. Is there a reason to set `this.container` every render pass? Shall we guard it against `null` here?

nwlieb: >Is there a reason to set this.container every render pass?
It's just the way it's done in the react docs, do you have a proposed alternative?

String refs are removed in react@16.0.0-alpha.13 so they will have to be removed eventually regardless...

abmai: It's the recommended pattern by FB: https://facebook.github.io/react/docs/refs-and-the-dom.html 

According to those docs, it should only be called after mount and before unmounting. I'm not sure if our issue in v3 was because we're using `createElement` directly vs. going through jsx.

nwlieb: So if I understand, the proper way to circumvent #310 is create a bound class method for the ref callback like described in https://facebook.github.io/react/docs/refs-and-the-dom.html#caveats

If I change it to that, would that work for you?

abmai: Interesting...but I guess that would make sense. Should we do that here and in V3 as well, @Pessimistress?

Pessimistress: Yeah I wonder if inline definition is what's causing the callback to be called every frame in v3. I will investigate.

nwlieb: I the interest of moving this along I changed all the ref callbacks to bound class methods. Let me know what you think.

src/map-interactions.react.js
```diff
@@ -126,12 +127,12 @@ export default class Interactions extends Component {
   }
 
   _getMousePos(event) {
-    const el = this.refs.container;
+    const el = this.container;
```
abmai: Can we initialize these to `null` in the `constructor` above and follow convention of prefixing these with `_` to indicate they're supposed to be private? e.g.
```js
constructor(props) {
  ...
  this._container = null;
}
```

abmai:
Looks good. Just one more nit from me (commented above) and we can merge this. Thanks!

nwlieb:
@abmai done!

PR #317, opened 2017-07-12
Match viewport to Mapbox at low zoom levels

Pessimistress:
Fit and re-center map in viewport to avoid empty space around the map.
screen shot 2017-07-12 at 2 36 54 pm

### DIFF
--- a/src/utils/map-state.js
+++ b/src/utils/map-state.js
@@ -7,6 +7,7 @@ export const MAPBOX_LIMITS = {
   maxZoom: 20,
   minPitch: 0,
   maxPitch: 60,
+  // defined by mapbox-gl
   maxLatitude: 85.05113,
   minLatitude: -85.05113
 };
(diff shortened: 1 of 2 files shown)


Pessimistress:
This should fix https://github.com/uber/deck.gl/issues/771 @shaojingli

src/utils/map-state.js
```diff
@@ -6,7 +6,9 @@ export const MAPBOX_LIMITS = {
   minZoom: 0,
   maxZoom: 20,
   minPitch: 0,
-  maxPitch: 60
+  maxPitch: 60,
+  maxLatitude: 85.05113,
```
heshan0131: Is this some magic number? would be nice to add a comment on this

Pessimistress: These are copied from Mapbox. I'll add a comment.

PR #318, opened 2017-07-14
Fix scale for map affected by a CSS transform. Fixes #299 in v2

### LINKED ISSUE #299: When using css transform: scale() mouse events do not work properly
If the map has been scaled by a css transform, the dragging scale and the centering is wrong, as if the map were using the real pixel location of the cursor and not accounting for the scale. We are using version 3.0.0-alpha.14

Example:
```
<div style={{transform: 'scale(0.5, 0.5) translateX(-50%)'}}>
    <MapGL {...mapProps} />
</div>
```
edit: this affects both v2 and v3

nwlieb:
This fixes #299 for v2. v3 is a different story since this logic is now hidden behind hammerjs.

Not expecting a merge since this goes against the direction that v3 is going in, but hopefully it spurs some discussion.

Edit: hammerjs seems more or less dead so unfortunately it won't be easy to fix this upstream for v3.

### DIFF
--- a/src/map-interactions.react.js
+++ b/src/map-interactions.react.js
@@ -54,7 +54,7 @@ function getTouchPositions(el, event) {
   const touches = getTouches(event);
 
   // Fix scale for map affected by a CSS transform.
-  // See https://stackoverflogit w.com/a/26893663/3528533
+  // See https://stackoverflow.com/a/26893663/3528533
   const scaleX = rect.width / el.offsetWidth;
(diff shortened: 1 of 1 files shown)


abmai:
Hmm happy to merge this for v2 since we should still fix issues with it for the short term.

Pessimistress:
We're handling pointer positions outside of hammerjs https://github.com/uber/react-map-gl/blob/master/src/utils/event-manager/event-manager.js#L189 - I think the same fix can work.

PR #319, opened 2017-07-14
Remove string refs for React 16.x

Pessimistress:
Per discussion in https://github.com/uber/react-map-gl/pull/316


Pessimistress:
The bug reported in https://github.com/uber/react-map-gl/pull/310 was caused by `React.createElement` and using inline ref callback definition.

PR #320, opened 2017-07-16
Update NavigationControl component docs

alexyuly:
I'm currently using version 3.0.0-beta.3 in a project, and while looking at the docs I fixed a few minor problems in the example code:
- It should import `NavigationControl` and not `Popup`.
- It should use `onViewportChange` and not `onChangeViewport` which is deprecated.
- It should be made clear that `viewport` and `updateViewport` are component props.

Thanks to the maintainers for their great work on this package.


abmai:
Thanks for the changes, @alexyuly!

PR #329, opened 2017-07-26
Fix mobile interaction for website examples

Pessimistress:
For 3.0-release

Fix bug where inputs are not interactive on the phone


examples/controls/src/control-panel.js
```diff
@@ -1,16 +1,20 @@
 import React, {PureComponent} from 'react';
 
+const defaultContainer =  ({children}) => <div className="control-panel">{children}</div>;
```
abmai: nit: this and all the below, extra space + casing on name (DefaultContainer)

Issue #331, opened 2017-07-27
Pinch to zoom not working on website


shaojingli:
Pinch-to-zoom is not working on our website https://uber.github.io/react-map-gl/#/ when accessed from mobile devices.

I triaged it a bit and found that it's due to the react-map-gl itself.
The last known good version of react-map-gl is 3.0.0-beta.1 and the first known bad is 3.0.0-beta.2. So there must be a commit shown here

<img width="984" alt="screen shot 2017-07-27 at 9 57 29 am" src="https://user-images.githubusercontent.com/21322704/28682492-07ea6d56-72b2-11e7-9514-f408844184a8.png">
 
@Pessimistress

PR #332, opened 2017-07-27
Interaction bug fixes

Pessimistress:
For 3.0-release

- Fix pinch zoom
- Fix bug where wheel scrolling is blocked when `onViewportChange` is null


Pessimistress:
https://github.com/uber/react-map-gl/issues/331

PR #342, opened 2017-08-04
Corrects the env variable used

daumann:
MapboxAccessToken and not MAPBOX_ACCESS_TOKEN is used.
see https://github.com/uber/react-map-gl/blob/master/website/webpack/config.js#L67


abmai:
Thanks for the fix!

PR #346, opened 2017-08-08
Fix scale for map affected by a CSS transform in v3

Pessimistress:
**For 3.0-release**

Fix pointer position when the map is scaled by css transform.
Same technique as @nwlieb used in https://github.com/uber/react-map-gl/pull/318


nwlieb:
Thank you for pushing this through and apologies for not being on top of it 👍

PR #348, opened 2017-08-09
Add missing methods to mock manager.

### LINKED ISSUE #347: Error coming from event-manager in non browser environment.
We have some tests that render react-map-gl components using Jest and noticed an error after upgrading to v3:
TypeError: this.manager.get is not a function

It looks like the event-manager was updated to access more methods on the manager but the mock manager wasn't updated.  I fixed the problem locally by adding the chainedNoop functions for instance.get and instance.set though that may not be the best fix.  If it is, I'm happy to submit a PR for with the changes.  Thanks!

gsf-smclamb:
Added missing methods to mock manager to prevent errors in non browser environments.

This fixes #347 .
@ibgreen @ericsoco

### DIFF
--- a/src/utils/event-manager/event-manager.js
+++ b/src/utils/event-manager/event-manager.js
@@ -16,6 +16,8 @@ function ManagerMock(m) {
   instance.off = chainedNoop;
   instance.destroy = chainedNoop;
   instance.emit = chainedNoop;
+  instance.get = chainedNoop;
+  instance.set = chainedNoop;
   return instance;
 }
 


ibgreen:
@gsf-smclamb - will land as soon as you sign the CLA.

src/utils/event-manager/event-manager.js
```diff
@@ -16,6 +16,8 @@ function ManagerMock(m) {
   instance.off = chainedNoop;
   instance.destroy = chainedNoop;
   instance.emit = chainedNoop;
+  instance.get = chainedNoop;
```
ericsoco: This mock is incorrect -- `get()` should return a value, not the instance.
See also [deck.gl](https://github.com/uber/deck.gl/blob/master/src/controls/events/event-manager.js#L16)...is it time to break off `event-manager` into a module/package shared between these libs? /cc @ibgreen @Pessimistress

pelotom:
We're awaiting this fix with much anticipation, any timeline for a new release?

PR #352, opened 2017-08-29
Fixed deck.gl link

joao:
Link was pointing to react-mal-gl repo.
Replaced it with deck.gl's website.


abmai:
Oh woops, nice catch. Thanks for the fix, @joao!

PR #355, opened 2017-09-08
Import EventManager from mjolnir.js

Pessimistress:


src/components/interactive-map.js
```diff
@@ -6,7 +6,7 @@ import StaticMap from './static-map';
 import {MAPBOX_LIMITS} from '../utils/map-state';
 import {PerspectiveMercatorViewport} from 'viewport-mercator-project';
 
-import EventManager from '../utils/event-manager/event-manager';
+import {EventManager} from 'mjolnir.js';
```
ibgreen: Nit: Should `EventManager` also be a default export?

PR #357, opened 2017-09-18
Package.json changes to support React 16

ibgreen:
React 16 rc.3 is out so it should be a good time to start testing this. 
* This PR mainly deals with package updates, I did a quick verification that examples still render but we will need some time to ensure that all examples are working as expected. 
* My current view is that React 16 support should go into v3.1, but let's see how big the impact is.


Apercu:
I'm all for testing it, but if we envision to create frameworks with high-standards and with the least possibility of breakage in production, I don't think we should draft new releases with alpha/beta/rc dependencies and just wait for their major as much as possible.

ibgreen:
> I don't think we should draft new releases with alpha/beta/rc dependencies and just wait for their major as much as possible.

That is a fair point! 

To be clear, I am not suggesting we hold off releasing 3.1 if it turns out that React 16 is not ready by the time we are ready. 
My thinking is that moving a couple of dependencies back to React 15 before the release would be an easy step. But getting the testing done on React 16 starting now is valuable. 

The reason I update 'What's New' is to help us remember what to review before we cut the release, and make sure we don't need to scramble to remember what we did at that time. It doesn't mean it is a done deal.

akrawchyk:
During `npm install` using this PR's latest commit hash, I get warnings about deprecated modules, see the log below. I'm wondering if it would make sense to update those package names in this PR?

```
➜ npm install https://github.com/uber/react-map-gl.git#283b212b6f0344e61079f69751af5789fb447a01                                     
npm WARN deprecated point-geometry@0.0.0: This module has moved: please install @mapbox/point-geometry instead
npm WARN deprecated vector-tile@1.3.0: This module has moved: please install @mapbox/vector-tile instead
npm WARN deprecated geojson-area@0.1.0: This module is now under the @mapbox namespace: install @mapbox/geojson-area instead

> mapbox-gl@0.38.0 postinstall /home/andrew/Developer/example-map/node_modules/mapbox-gl
> cd src/style-spec && yarn || npm install

yarn install v1.0.1
[1/4] Resolving packages...
[2/4] Fetching packages...
[3/4] Linking dependencies...
[4/4] Building fresh packages...
warning Your current version of Yarn is out of date. The latest version is "1.1.0" while you're on "1.0.1".
info To upgrade, run the following command:
$ sudo apt-get update && sudo apt-get install yarn
Done in 0.77s.
npm WARN react-map-gl@3.1.0-alpha.1 requires a peer of react@>=15.4.x but none is installed. You must install peer dependencies your
self.

+ react-map-gl@3.1.0-alpha.1
added 151 packages, removed 94 packages and updated 1 package in 14.167s
```

ibgreen:
>  I'm wondering if it would make sense to update those package names in this PR?

@akrawchyk Those files look like indirect mapbox dependencies (they are not listed in react-map-gl package.json), so to fix them we will likely need to bump mapbox. That should be done in a separate PR. Opened #372

Apercu:
React  16 is now available in a non-rc version

ibgreen:
> React 16 is now available in a non-rc version

Updates

PR #372, opened 2017-10-03
Bump mapbox to 0.40.1

ibgreen:


abmai:
Any gotchas of updating to a newer version of Mapbox that we should worry about? Or we're just gonna test it after landing.

ibgreen:
> Any gotchas of updating to a newer version of Mapbox that we should worry about? 

Scanned their release notes, there is a bunch of breaking changes but they seem unlikely to affect us. 

But yes we need to do a round of testing. Maybe more so for React 16 than this bump.

PR #373, opened 2017-10-03
Fixes to no-token warning

ibgreen:


ibgreen:
> Great addition. We're not using jsx in the components to avoid using loaders and keep it simple right?

luma and deck are written in pure ES6 so that we can run in Chrome and Node without transpilation. To ensure that we have consistent build tooling, we have adopted the same approach in all the frameworks in the suite.

Note that the pure ES6 restriction does not apply to examples, just to the core lib.

PR #375, opened 2017-10-03
Add babel-runtime to dependencies

Pessimistress:
Fix https://github.com/uber/react-map-gl/issues/374 as instructed by [transform-runtime documentation](https://babeljs.io/docs/plugins/transform-runtime/).

This is only required by the transpiled es5 version.

### DIFF
--- a/package.json
+++ b/package.json
@@ -33,10 +33,11 @@
     "publish-beta": "npm run build && npm run test && npm publish --tag beta"
   },
   "dependencies": {
+    "babel-runtime": "^6.23.0",
     "bowser": "^1.2.0",
-    "mjolnir.js": ">=0.1.0",
     "immutable": "*",
     "mapbox-gl": "0.38.0",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -18,7 +18,8 @@
   "module": "dist-es6/index.js",
   "files": [
     "src",
-    "dist"
+    "dist",
+    "dist-es6"
```
ibgreen: To minimize risk of things going wrong, I am not sure that this change should be part of the same PR. This will cause webpack to start using a different dist that doesn't have the transform, increasing the test matrix. 

Note that `dist-es6` is not pure ES6, it is completely transpiled to ES5 **except** import/exports, so it could still be affected by babel issues.

Pessimistress: Happy to drop it from this PR.
While I am not sure if the `module` field is actually effective, what is the point of building to es6 if the transpiled code is not included in the package?

PR #376, opened 2017-10-05
Marker extended with custom className and onClick event fn

rshtg:
It was emergency necessary for me. Hope it would be useful for others. (To avoid local overrides)

### DIFF
--- a/src/components/marker.js
+++ b/src/components/marker.js
@@ -51,7 +51,7 @@ const contextTypes = {
 export default class Marker extends Component {
 
   render() {
-    const {longitude, latitude, offsetLeft, offsetTop} = this.props;
+    const {longitude, latitude, offsetLeft, offsetTop, className, onClick} = this.props;
 
     const [x, y] = this.context.viewport.project([longitude, latitude]);
     const containerStyle = {
@@ -61,9 +61,10 @@ export default class Marker extends Component {
     };
 
     return createElement('div', {
-      className: 'mapboxgl-marker',
+      className: className,
       style: containerStyle,
-      children: this.props.children
+      children: this.props.children,
+      onClick: onClick
     });
   }
 


ibgreen:
@rsthg - Many thanks for the PR. Added the appropriate reviewers. 

If accepted, this PR would need to also contain 
* doc updates to mention the new props in the reference docs for marker
* Add a line in the What's New doc for the upcoming release to mention that new props are available.

I would also object to adding a feature like this in only a single component. Seems like it should be done across the board or not at all.

Saw that you had some build errors, I couldn't find them in your changes, but be aware we are using pure ES6, so no object destructuring (which seems to be the gist of the error messages).

Finally, I suspect you'd also want this to be cherry-picked into 3.0-release branch.

src/components/marker.js
```diff
@@ -61,9 +61,10 @@ export default class Marker extends Component {
     };
 
     return createElement('div', {
-      className: 'mapboxgl-marker',
+      className: className,
```
Pessimistress: I agree allowing custom class name would be desirable for all our control components. Because we rely on Mapbox's stylesheet, some of the native class names are still necessary. Could you change this to `mapboxgl-marker ${className}` (and maybe add the prop to `Popup` and `NavigationControl` as well)?

src/components/marker.js
```diff
@@ -61,9 +61,10 @@ export default class Marker extends Component {
...
       style: containerStyle,
-      children: this.props.children
+      children: this.props.children,
+      onClick: onClick
```
Pessimistress: This seems arbitrary. Any reason you cannot register the `onClick` event on the child component?

src/components/marker.js
rshtg: Yes, I can work on it

src/components/marker.js
rshtg: Just wanted to make it clearly.

Issue #304, opened 2017-06-30
OnClick, OnDrag and other events are propagated even though e.preventDefault() and e.stopPropagation() are called


aindeev:
When you add Markers or other Overlays on top of the map, the pointer events are still propagated to the map.

For example if we render a "control panel" on top of the map in the corner, we can still drag the map around through the overlay. The only way around this issue is to render such content outside of the map component

The same problem happens with Markers. In my case I am adding a onClick handler on the Marker itself and adding an additional onClick markers on the map itself. Both handlers are fired even if we specify:

```js
onMarkerClick(e) {
   e.preventDefault()
   e.stopPropagation()
}
```

```js
onMapClick(e) {
    // -> still fired
}
```

I am attempting to create something similar in nature to this.
https://www.mapbox.com/mapbox-gl-js/example/toggle-layers/

alex2wong:
Is there any comments on this? I have tested in simple react parent/child components, event can be stopped in event handler in child component which would not propagate to parent.

But dblClick event can always pass to mapComponent causing zoom, stopPropagation does not work. Weird.

kriscarle:
I just ran into this as well. It appears be because the eventCanvas is a parent of the overlays container where the children are inserted.  It is present in the examples too, for example in https://uber.github.io/react-map-gl/#/Examples/markers-popups you can drag the map while on top of the overlays

My work around is to put my controls outside of the react-map-gl component inside of passing them as children.

Pessimistress:
The problem here is caused by us using hammer.js under the hood. The `click` and `doubleClick` events are triggered by `pointerUp`, which always fires before the native `click` and `dblClick`.
@ericsoco @ibgreen @shaojingli

Pessimistress:
@aindeev @alex2wong @kriscarle In your use cases, would it work if the Marker/Popup components by default blocks map interaction (click and drag do not propagate) but also offers a prop to turn it on?

heathercodes:
I'm not sure if this fix was solved for @aindeev , but this propagation issue is still affecting my app despite the fix. 
If on the `ReactMapGL` you specify an `onClick` function, then have a Popup component as a child, any onClick callback functions in the Popup component will still propagate to the map.
Am I missing something? I've looked through the examples and documentation without luck

Pessimistress:
@heathercodes You can block the click event from propagating by setting `captureClick: false` on the Popup component (see https://github.com/uber/react-map-gl/pull/377).

[...]

Issue #195, opened 2017-02-20
Unable to disable the drag event


DinoJay:
I'm currently working on a custom overlay. I want to enable zoom without the mousedrag behavior. It works by setting the following props:  
```
viewport.isDragging = false;
viewport.startDragLngLat = false;
```

Unfortunately, I got the following exception when a drag event is fired:
```
Uncaught AssertionError {name: "AssertionError", actual: null, expected: true, operator: "==", message: "`startDragLngLat` prop is required for mouse drag behavior to calculate where to position the map."…}actual: nullexpected: truegeneratedMessage: falsemessage: "`startDragLngLat` prop is required for mouse drag behavior to calculate where to position the map."name: "AssertionError"operator: "=="stack: "AssertionError: `startDragLngLat` prop is required for mouse drag behavior to calculate where to position the map.    at MapGL._onMouseDrag (https://127.0.0.1:8888/bundle.js:58561:28)    at Interactions._onMouseDrag (https://127.0.0.1:8888/bundle.js:57560:20)"__proto__: Error
fail @ assert.js:195
ok @ assert.js:215
_onMouseDrag @ map.react.js:795
_onMouseDrag @ map-interactions.react.js:202
```

Is there a nicer way to achieve that functionality?

Best,
Jan

abmai:
hey @DinoJay - both `isDragging` and `startDragLngLat` are internal variables and shouldn't be modified directly. They're only exposed because of the stateless paradigm that the library was originally created under.

Are you trying to disable the entire drag event? Or you just don't want the map to be moved. If it's the latter, you can just ignore the `latitude/longitude` that gets passed back to `onChangeViewport` (and not pass it back into the react-map-gl instance).

DinoJay:
Actually, I'm working with a SVG overlay. But when I trigger a `drag` event the map get shifted and the SVG elements are not updated until I release the mouse again. What I want to achieve is to keep the map interactive by passing back the `latitude` and `longitude` props. But I also want to disable the drag to keep the svg points always in sync with the underlying map layer.

PR #377, opened 2017-10-05
Allow controls and overlays to block map interactions

Pessimistress:
Attempting to tackle problem raised in issues such as https://github.com/uber/react-map-gl/issues/304 https://github.com/uber/react-map-gl/issues/195

Adds the following props to `Marker`, `Popup`, `NavigationControl` and overlay components:
- `captureScroll` {Boolean}
- `captureDrag` {Boolean}
- `captureClick` {Boolean}
- `captureDoubleClick` {Boolean}

### DIFF
--- a/docs/components/marker.md
+++ b/docs/components/marker.md
@@ -36,13 +36,13 @@ Offset of the marker from the left in pixels, negative number indicates left.
 Offset of the marker from the top in pixels, negative number indicates up.
 
 ##### `captureScroll` {Boolean} - default: `false`
-Block map zoom when scrolling over this component.
+Stop propagation of mouse wheel event to the map component. Can be used to stop map from zooming when this component is scrolled.
 
 ##### `captureDrag` {Boolean} - default: `true`
-Block map pan & rotate when dragging this component.
+Stop propagation of dragstart event to the map component. Can be used to stop map from panning when this component is dragged.
 
 ##### `captureClick` {Boolean} - default: `true`
-Block map click when clicking on this component.
+Stop propagation of click event to the map component. Can be used to stop map from calling the `onClick` callback when this component is clicked.
 
 ##### `captureDoubleClick` {Boolean} - default: `true`
-Block map zoom when double clicking on this component.
+Stop propagation of dblclick event to the map component. Can be used to stop map from zooming when this component is double clicked.
--- a/docs/components/marker.md
+++ b/docs/components/marker.md
@@ -34,3 +34,15 @@ Offset of the marker from the left in pixels, negative number indicates left.
 
 ##### `offsetTop` {Number} - default: `0`
 Offset of the marker from the top in pixels, negative number indicates up.
+
+##### `preventScrollZoom` {Boolean} - default: `false`
+Block map zoom when scrolling over this component.
+
+##### `preventDragPanRotate` {Boolean} - default: `true`
+Block map pan & rotate when dragging this component.
+
+##### `preventClick` {Boolean} - default: `true`
+Block map click when clicking on this component.
+
+##### `preventDoubleClickZoom` {Boolean} - default: `true`
+Block map zoom when double clicking on this component.
--- a/docs/components/marker.md
+++ b/docs/components/marker.md
@@ -35,14 +35,14 @@ Offset of the marker from the left in pixels, negative number indicates left.
 ##### `offsetTop` {Number} - default: `0`
 Offset of the marker from the top in pixels, negative number indicates up.
 
-##### `preventScrollZoom` {Boolean} - default: `false`
+##### `captureScroll` {Boolean} - default: `false`
 Block map zoom when scrolling over this component.
 
-##### `preventDragPanRotate` {Boolean} - default: `true`
+##### `captureDrag` {Boolean} - default: `true`
 Block map pan & rotate when dragging this component.
 
-##### `preventClick` {Boolean} - default: `true`
+##### `captureClick` {Boolean} - default: `true`
 Block map click when clicking on this component.
 
-##### `preventDoubleClickZoom` {Boolean} - default: `true`
+##### `captureDoubleClick` {Boolean} - default: `true`
 Block map zoom when double clicking on this component.
--- a/src/components/base-control.js
+++ b/src/components/base-control.js
@@ -1,0 +1,111 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
+
+// Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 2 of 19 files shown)


abmai:
Is it worth just using the same prop names as the ones present in `InteractiveMap`? We can specify that overlays will take precedents over base map behavior.

What would be the behavior if there are multiple overlays, Marker, Popup, etc.., each with a different behavior on these interactions?

Pessimistress:
The setting only applies to events that happen *over* the overlay component. If multiple overlays overlap each other, then the one on top will take precedence.

Pessimistress:
- `preventScrollZoom` - blocks any wheel scroll event, without knowing what it does. 
- `preventDragPanRotate` - blocks drag, without differentiating pan and rotate. 
- `preventClick` - blocks click. (no equivalent toggle in `InteractiveMap`)
- `preventDoubleClickZoom` blocks any double click event, without knowing what it does.

Alternative A: rename to `captureScroll` or `captureDrag`
Alternative B: change boolean flags to callbacks e.g. `shouldPreventScrollZoom` `shouldPreventClick` to offer more user control

Note that we don't have a prop to block pinch zoom on mobile, as we cannot easily differentiate a gesture that's performed over the control vs. over the map.

abmai:
I like Alternative A. `capture` is more in line with describing what's happening (and is used normally when talking about mouse events). The event itself will still happen within these components, just not passed through to the underlying map.

docs/components/marker.md
```diff
@@ -34,3 +34,15 @@ Offset of the marker from the left in pixels, negative number indicates left.
...
 ##### `offsetTop` {Number} - default: `0`
 Offset of the marker from the top in pixels, negative number indicates up.
+
+##### `captureScroll` {Boolean} - default: `false`
+Block map zoom when scrolling over this component.
+
+##### `captureDrag` {Boolean} - default: `true`
+Block map pan & rotate when dragging this component.
+
+##### `captureClick` {Boolean} - default: `true`
```
abmai: I would elaborate here to say "capture mouse event at this component" or "prevent propagation of click to lower components, including map".

src/utils/map-state.js
```diff
@@ -134,9 +134,9 @@ export default class MapState {
   pan({pos, startPos}) {
     const startPanLngLat = this._interactiveState.startPanLngLat || this._unproject(startPos);
 
-    // take the start lnglat and put it where the mouse is down.
-    assert(startPanLngLat, '`startPanLngLat` prop is required ' +
-      'for mouse pan behavior to calculate where to position the map.');
+    if (!startPanLngLat) {
+      return this;
```
abmai: What's the motivation for returning `this` here?

Pessimistress: The event blocked by `captureDrag` is actually `panstart` (If you drag the map and the pointer moves over a popup, the map still pans. If you drag the popup, nothing happens). As a result the `panmove` event still fires with a missing `startPanLngLat`. In this case, instead of issuing an error, we just keep the current map state.

src/components/base-control.js
```diff
@@ -0,0 +1,110 @@
+// Copyright (c) 2015 Uber Technologies, Inc.
```
ibgreen: Nit: I am not keen on the word `base` in components. To me it emphasizes a programming language aspect rather than something semantic. Would calling it `MapControl` or just `Control` work?

src/components/base-control.js
```diff
@@ -0,0 +1,110 @@
...
+      wheel: this._onEvent,
+      panstart: this._onEvent,
+      click: this._onEvent,
+      dblclick: this._onEvent
+    };
+
+    if (ref) {
+      this.context.eventManager.on(events, ref);
+    } else {
+      this.context.eventManager.off(events);
```
ibgreen: Should component unmounting be considered?

src/overlays/canvas-overlay.js
```diff
@@ -18,22 +18,24 @@
...
-const propTypes = {
+const propTypes = Object.assign({}, BaseControl.propTypes, {
   redraw: PropTypes.func.isRequired
-};
+});
 
-const contextTypes = {
-  viewport: PropTypes.instanceOf(PerspectiveMercatorViewport),
-  isDragging: PropTypes.bool
+const defaultProps = {
```
ibgreen: Why are these listed here instead of inherited?

src/overlays/html-overlay.js
```diff
@@ -18,21 +18,23 @@
...
+const propTypes = Object.assign({}, BaseControl.propTypes, {
   redraw: PropTypes.func.isRequired,
   style: PropTypes.object
-};
+});
 
-const contextTypes = {
-  viewport: PropTypes.instanceOf(PerspectiveMercatorViewport),
-  isDragging: PropTypes.bool
+const defaultProps = {
```
ibgreen: ditto

src/overlays/svg-overlay.js
```diff
@@ -18,21 +18,23 @@
...
+const propTypes = Object.assign({}, BaseControl.propTypes, {
   redraw: PropTypes.func.isRequired,
   style: PropTypes.object
-};
+});
 
-const contextTypes = {
-  viewport: PropTypes.instanceOf(PerspectiveMercatorViewport),
-  isDragging: PropTypes.bool
+const defaultProps = {
```
ibgreen: ditto

src/components/base-control.js
Pessimistress: This callback is invoked before unmounting.

src/overlays/canvas-overlay.js
Pessimistress: The overlays' default props are different from those of map controls.

Pessimistress: Main difference being the map controls (e.g. the NavigationControl and Popup) are expected to capture click and drag events. The overlays, on the other hand, are full screen. If any of these props are `true` the app will be broken by default.

src/components/base-control.js
ibgreen: > _onContainerLoad

Thanks. This one took some time for me to grok. I see now it is being used as a React ref callback. I didn't know those were called on unmount. Could be worth a comment or improved naming.

Also the name of the callback becomes slightly misleading as it handles both load and unload callbacks: `_onContainerLoad` => `_onContainerRefChanged` ?

src/overlays/canvas-overlay.js
ibgreen: > The overlays' default props are different from those of map controls.

OK, I thought I had checked that they were the same, but you are right they are different. Thanks for clarifying.

PR #379, opened 2017-10-09
Marker render CSS optimization

rshtg:
This kind of optimization gives us about 50-60 FPS on one zoom level


Pessimistress:
@rshtg Thanks! Can you fix the lint errors https://travis-ci.org/uber/react-map-gl/jobs/287001934

src/components/marker.js
```diff
@@ -53,14 +53,25 @@ export default class Marker extends BaseControl {
...
       top: y + offsetTop
     };
 
+    const isVisible = !(top < 0 || left < 0);
+
+    let containerStyle = {};
+
+    if (isVisible) {
+      containerStyle = {
+        transform: `translate3d(${left}px, ${top}px, 0)`
```
Pessimistress: Why is CSS transform preferable than absolute positioning? Is it contributing to perf at all?

src/components/marker.js
```diff
@@ -53,14 +53,25 @@ export default class Marker extends BaseControl {
...
+      };
+    }
+
     return createElement('div', {
-      className: `mapboxgl-marker ${className}`,
+      // isVisible toggles the .visible to show/hide the element
+      className: `mapboxgl-marker ${className} ${isVisible ? 'visible' : ''}`,
```
Pessimistress: I can see that avoid rendering out-of-view markers provide perf benefit. Though I'm not convinced about the `visible` class. It adds an additional catch for the user to implement, and it doesn't account for the size of the marker.  @abmai

src/components/marker.js
rshtg: CSS transform is more productive property which uses GPU across rendering. I've complete some test with 50-100-150 markers on Mobile and it shows 2x better performance iOS Safari (Latest) / Android Chrome (Latest)

src/components/marker.js
rshtg: Am I need to put the marker sizes check. Or it would be better to roll it back?

Pessimistress: @rshtg We don't ask the user to specify the size of each marker today. Dynamically querying their sizes may cause perf problems. Shall we remove it for now to land the transform change?

src/components/marker.js
Pessimistress: Shall we apply this technique to the Popup component as well?

src/components/marker.js
rshtg: I think we just can keep only transformation without `visible` toggler. So the profit should be affected in any case

src/components/marker.js
rshtg: Yes, I think each component should be transformed as well

istarkov:
css transform can cause a lot of issues with hovering, 
visual issues like font smoothing,
and working with 3rd party components like tooltips.

May be this has been changed last 1-2 years, but in my library https://github.com/istarkov/google-map-react I also used style positioning and not because I don't have a knowledge about transforms and performance. 
Also @rshtg have you tried to use style transform with additional css on marker `will-change: left, top;` it must give the same effect on modern browsers as css transform.

PR #381, opened 2017-10-12
Right button drag to rotate

Pessimistress:
- Right button drag to rotate (match Mapbox behavior)
- Disable drag rotate on map controls

### DIFF
--- a/src/utils/map-controls.js
+++ b/src/utils/map-controls.js
@@ -86,6 +86,11 @@ export default class MapControls {
       srcEvent.ctrlKey || srcEvent.shiftKey);
   }
 
+  isRightButtonPressed(event) {
+    const {srcEvent: {which, buttons}} = event;
+    return buttons === undefined ? (which === 3) : Boolean(buttons & 2);
+  }
+
   setState(newState) {
     Object.assign(this._state, newState);
     if (this.onStateChange) {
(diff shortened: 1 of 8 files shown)


src/utils/map-controls.js
```diff
@@ -86,6 +86,11 @@ export default class MapControls {
       srcEvent.ctrlKey || srcEvent.shiftKey);
   }
 
+  isRightButtonPressed(event) {
+    const {srcEvent: {which, buttons}} = event;
+    return buttons === undefined ? (which === 3) : Boolean(buttons & 2);
```
abmai: Can we add a comment to indicate what the logic here is? i.e. what is `3` (assuming it's the right mouse code), and what `buttons & 2` does.

src/utils/map-controls.js
```diff
@@ -86,6 +86,11 @@ export default class MapControls {
       srcEvent.ctrlKey || srcEvent.shiftKey);
   }
 
+  isRightButtonPressed(event) {
```
ibgreen: Seems like something mjolnir should offer?

src/utils/map-controls.js
abmai: Actually, do we need to use `&` here?

Pessimistress: [buttons](https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/buttons) is a sum of the values of all active buttons:

src/utils/map-controls.js
Pessimistress: Ideally. The event may be generated from one of several listeners so the format is inconsistent. We need to do an API audit on mjolnir.js.

src/utils/map-controls.js
abmai: I see, thanks! It's definitely worth leaving a comment then, especially for the `&` and how it's an accurate way of detecting the right mouse press.

PR #382, opened 2017-10-13
Add keyboard support

Pessimistress:
Match Mapbox behavior documented [here](https://www.mapbox.com/mapbox-gl-js/api/#keyboardhandler).


src/utils/map-controls.js
```diff
@@ -86,17 +89,6 @@ export default class MapControls {
       srcEvent.ctrlKey || srcEvent.shiftKey);
   }
 
-  isRightButtonPressed(event) {
```
abmai: This moved into mjolnir?

Pessimistress: Yes. https://github.com/uber-web/mjolnir.js/pull/5

PR #383, opened 2017-10-19
Port TransitionManager from deck.gl

Pessimistress:
New `InteractiveMap` props:
- Transition props
  - `transitionDuration`
  - `transitionInterpolator`
  - `transitionInterruption`
  - `transitionEasing`
- Transition callbacks
  - `onTransitionStart`
  - `onTransitionInterrupt`
  - `onTransitionEnd`

The transition props are also introduced into `MapControls` as it needs to manipulate their values.

### DIFF
--- a/examples/viewport-animation/src/app.js
+++ b/examples/viewport-animation/src/app.js
@@ -1,20 +1,12 @@
 /* global window */
 import React, {Component} from 'react';
 import {render} from 'react-dom';
-import MapGL from 'react-map-gl';
+import MapGL, {experimental} from 'react-map-gl';
 import {PerspectiveMercatorViewport} from 'viewport-mercator-project';
-import TWEEN from 'tween.js';
 
 import ControlPanel from './control-panel';
 
 const MAPBOX_TOKEN = ''; // Set your mapbox token here
-
-// Required by tween.js
-function animate() {
-  TWEEN.update();
-  window.requestAnimationFrame(animate);
-}
-animate();
 
 export default class App extends Component {
 
@@ -39,33 +31,24 @@ export default class App extends Component {
     window.removeEventListener('resize', this._resize);
   }
 
-  _resize = () => {
-    this.setState({
-      viewport: {
-        ...this.state.viewport,
-        width: this.props.width || window.innerWidth,
-        height: this.props.height || window.innerHeight
-      }
+  _onViewportChange = viewport => this.setState({
+    viewport: {...this.state.viewport, ...viewport}
+  });
+
+  _resize = () => this._onViewportChange({
+    width: this.props.width || window.innerWidth,
+    height: this.props.height || window.innerHeight
+  });
+
+  _goToViewport = ({longitude, latitude}) => {
+    this._onViewportChange({
+      longitude,
+      latitude,
+      zoom: 11,
+      transitionInterpolator: experimental.viewportFlyToInterpolator,
+      transitionDuration: 3000
     });
   };
-
-  _easeTo = ({longitude, latitude}) => {
-    // Remove existing animations
-    TWEEN.removeAll();
-
-    const {viewport} = this.state;
-
-    new TWEEN.Tween(viewport)
-      .to({
-        longitude, latitude,
-        zoom: 11
-      }, 3000)
-      .easing(TWEEN.Easing.Cubic.InOut)
-      .onUpdate(() => this._onViewportChange(viewport))
-      .start();
-  };
-
-  _onViewportChange = viewport => this.setState({viewport});
 
   render() {
 
@@ -81,7 +64,7 @@ export default class App extends Component {
           dragToRotate={false}
           mapboxApiAccessToken={MAPBOX_TOKEN} />
         <ControlPanel containerComponent={this.props.containerComponent}
-          onViewportChange={this._easeTo} />
+          onViewportChange={this._goToViewport} />
       </div>
     );
   }
--- a/src/components/interactive-map.js
+++ b/src/components/interactive-map.js
@@ -313,7 +313,7 @@ export default class InteractiveMap extends PureComponent {
         style: eventCanvasStyle
       },
         createElement(StaticMap, Object.assign({}, this.props,
-          this._transitionManger && this._transitionManger.getTransionedViewport(),
+          this._transitionManger && this._transitionManger.getViewportInTransition(),
           {
             visible: this._checkVisibilityConstraints(this.props),
             ref: this._staticMapLoaded,
--- a/src/components/interactive-map.js
+++ b/src/components/interactive-map.js
@@ -6,6 +6,8 @@ import StaticMap from './static-map';
 import {MAPBOX_LIMITS} from '../utils/map-state';
 import {PerspectiveMercatorViewport} from 'viewport-mercator-project';
 
+import TransitionManager from '../utils/transition-manager';
+
 import {EventManager} from 'mjolnir.js';
 import MapControls from '../utils/map-controls';
 import config from '../config';
@@ -30,6 +32,20 @@ const propTypes = Object.assign({}, StaticMap.propTypes, {
    * such as `longitude`, `latitude`, `zoom` etc.
    */
   onViewportChange: PropTypes.func,
+
+  /** Viewport transition **/
+  // transition duration for viewport change
+  transitionDuration: PropTypes.number,
+  // function called for each transition step, can be used to perform custom transitions.
+  transitionInterpolator: PropTypes.func,
+  // type of interruption of current transition on update.
+  transitionInterruption: PropTypes.number,
+  // easing function
+  transitionEasing: PropTypes.func,
+  // transition status update functions
+  onTransitionStart: PropTypes.func,
+  onTransitionInterrupt: PropTypes.func,
+  onTransitionEnd: PropTypes.func,
 
   /** Enables control event handling */
   // Scroll to zoom
@@ -100,22 +116,25 @@ const getDefaultCursor = ({isDragging, isHovering}) => isDragging ?
   config.CURSOR.GRABBING :
   (isHovering ? config.CURSOR.POINTER : config.CURSOR.GRAB);
 
-const defaultProps = Object.assign({}, StaticMap.defaultProps, MAPBOX_LIMITS, {
-  onViewportChange: null,
-  onClick: null,
-  onHover: null,
+const defaultProps = Object.assign({},
+  StaticMap.defaultProps, MAPBOX_LIMITS, TransitionManager.defaultProps,
+  {
+    onViewportChange: null,
+    onClick: null,
+    onHover: null,
 
-  scrollZoom: true,
-  dragPan: true,
-  dragRotate: true,
-  doubleClickZoom: true,
-  touchZoomRotate: true,
+    scrollZoom: true,
+    dragPan: true,
+    dragRotate: true,
+    doubleClickZoom: true,
+    touchZoomRotate: true,
 
-  clickRadius: 0,
-  getCursor: getDefaultCursor,
+    clickRadius: 0,
+    getCursor: getDefaultCursor,
 
-  visibilityConstraints: MAPBOX_LIMITS
-});
+    visibilityConstraints: MAPBOX_LIMITS
+  }
+);
 
 const childContextTypes = {
   viewport: PropTypes.instanceOf(PerspectiveMercatorViewport),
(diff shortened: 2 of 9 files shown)


src/utils/map-state.js
```diff
@@ -53,6 +53,15 @@ export default class MapState {
     */
     altitude,
 
+    /** Viewport transitions */
+    transitionDuration,
```
Pessimistress: `MapState` does not consume these props, though it is used to normalize viewport props before sending them to callbacks.

src/utils/map-state.js
```diff
@@ -278,10 +293,16 @@ export default class MapState {
...
     // Normalize degrees
-    props.longitude = mod(props.longitude + 180, 360) - 180;
-    props.bearing = mod(props.bearing + 180, 360) - 180;
+    const {longitude, bearing} = props;
+    if (longitude < -180 || longitude > 180) {
+      props.longitude = mod(longitude + 180, 360) - 180;
+    }
+    if (bearing < -180 || bearing > 180) {
+      props.bearing = mod(bearing + 180, 360) - 180;
+    }
```
Pessimistress: These changes are to avoid changing `longitude` and `bearing` if unnecessary. Otherwise precision issue will break TransitionManager's `_areViewportsEqual`.

src/utils/transition-manager.js
```diff
@@ -0,0 +1,209 @@
...
+  transitionInterpolator: viewportLinearInterpolator,
+  transitionEasing: VIEWPORT_TRANSITION_EASING_FUNC,
+  transitionInterruption: TRANSITION_EVENTS.BREAK,
+  onTransitionStart: () => {},
+  onTransitionInterrupt: () => {},
+  onTransitionEnd: () => {}
+};
+
+export default class TransitionManager {
+  constructor(props) {
```
Pessimistress: I removed the `onTransitionUpdate` callback and have it use `props.onViewportChange`.

src/utils/transition-manager.js
```diff
@@ -0,0 +1,209 @@
...
+    return this.transitionContext.viewport;
+  }
+
+  // Process the vewiport change, either ignore or trigger a new transiton.
+  processViewportChange(nextProps) {
+    const isTransitionInProgress = this._isTransitionInProgress();
+    const isTransitionEnabled = this._isTransitionEnabled(nextProps);
+
+    // NOTE: Be cautious re-ordering statements in this function.
+    if (!isTransitionEnabled && !isTransitionInProgress ||
```
Pessimistress: This is to avoid the expensive comparisons in `_shouldIgnoreViewportChange` if transitions are not enabled.

src/utils/transition-manager.js
```diff
@@ -0,0 +1,209 @@
...
+    }
+    return true;
+  }
+  /* eslint-enable max-depth */
+
+  _createTransitionInterval(delay) {
+    if (this.transitionContext.interval) {
+      clearInterval(this.transitionContext.interval);
+    }
+    return setInterval(() => this._updateViewport(), delay);
```
Pessimistress: This is copied from deck.gl - should be changed to using `requestAnimationFrame`.

src/utils/viewport-transition-utils.js
```diff
@@ -0,0 +1,138 @@
+/* eslint max-statements: ["error", 50] */
```
Pessimistress: This file is identical to that from deck.gl.

examples/viewport-animation/src/app.js
```diff
@@ -39,34 +31,25 @@ export default class App extends Component {
...
-    // Remove existing animations
-    TWEEN.removeAll();
+  _onViewportChange = viewport => this.setState({
+    viewport: {...this.state.viewport, ...viewport}
+  });
 
-    const {viewport} = this.state;
+  _resize = () => this._onViewportChange({
+    width: this.props.width || window.innerWidth,
+    height: this.props.height || window.innerHeight
```
1chandu: This might trigger a `transition`, we should overwrite 'transitionDuration' to 0 here. (Might need same change in `viewport-animation` example of deck.gl but should be handled correctly in other examples. Try re-sizing the window.

Pessimistress:
The TransitionManager is diverged from deck.gl's version to fix some imminent bugs. I will consolidate the changes with @1chandu .

src/components/interactive-map.js
```diff
@@ -290,11 +312,14 @@ export default class InteractiveMap extends PureComponent {
         ref: this._eventCanvasLoaded,
         style: eventCanvasStyle
       },
-        createElement(StaticMap, Object.assign({}, this.props, {
-          visible: this.checkVisibilityConstraints(this.props),
-          ref: this._staticMapLoaded,
-          children: this._eventManager ? this.props.children : null
-        }))
+        createElement(StaticMap, Object.assign({}, this.props,
+          this._transitionManger && this._transitionManger.getTransionedViewport(),
```
1chandu: Is it ok if `getTransionedViewport()` returns `null` ?

src/components/navigation-control.js
```diff
@@ -42,7 +47,7 @@ export default class NavigationControl extends BaseControl {
 
   _updateViewport(opts) {
     const {viewport} = this.context;
-    const mapState = new MapState(Object.assign({}, viewport, opts));
+    const mapState = new MapState(Object.assign({}, viewport, LINEAR_TRANSITION_PROPS, opts));
```
1chandu: `LINEAR_TRANSITION_PROPS` has `transitionDuration` set, is it intentional to enable transition in this flow?

src/utils/map-controls.js
```diff
@@ -290,7 +298,7 @@ export default class MapControls {
     const isZoomOut = this.isFunctionKeyPressed(event);
 
     const newMapState = this.mapState.zoom({pos, scale: isZoomOut ? 0.5 : 2});
-    return this.updateViewport(newMapState);
+    return this.updateViewport(newMapState, LINEAR_TRANSITION_PROPS);
```
1chandu: 👍  quick transition on double-tap

src/utils/map-controls.js
```diff
@@ -98,9 +106,9 @@ export default class MapControls {
 
   /* Callback util */
   // formats map state and invokes callback function
-  updateViewport(newMapState, extraState = {}) {
+  updateViewport(newMapState, extraProps = {}, extraState = {}) {
```
1chandu: Does the code get more clean if you put last two props into single object ?

something like :
```
updateViewport(newMapState, {extraProps = NO_TRANSITION_PROPS, extraState = {}})

when calling updateViewport you can set only extraProps or extraState or both.
```

src/utils/transition-manager.js
```diff
@@ -0,0 +1,209 @@
+/* global setInterval, clearInterval */
```
1chandu: assuming no changes from deck.gl version.

src/utils/viewport-transition-utils.js
1chandu: assuming no changes from deck.gl version.

src/utils/transition-manager.js
1chandu: I see you already commented where changed, will take a look.

src/utils/transition-manager.js
1chandu: `_shouldIgnoreViewportChange` checks `_isTransitionInProgress` before it compares the viewport.

src/components/navigation-control.js
Pessimistress: Yes, this is trying to mimic Mapbox 's behavior

src/utils/transition-manager.js
Pessimistress: It still compares the two viewport even if transitionDuration is 0.

examples/viewport-animation/src/app.js
Pessimistress: Good point.
Now I think about it, should we remove width and height from interpolation? And automatically kill the transition in transition manager if the canvas size changed?

src/components/interactive-map.js
Pessimistress: Yes, Object.assign works with null and undefined.

src/utils/map-state.js
Pessimistress: @1chandu the linear interpolator needs to handle longitude and bearing change cross the 180 boundary.

src/utils/transition-manager.js
```diff
@@ -0,0 +1,209 @@
...
+
+  _shouldIgnoreViewportChange(nextProps) {
+    // Ignore update if it is due to current active transition.
+    if (this._isUpdateDueToCurrentTransition(nextProps)) {
+      return true;
+    }
+
+    // Ignore update if it is requested to be ignored
+    if (this._isTransitionInProgress() &&
+      this.props.transitionInterruption === TRANSITION_EVENTS.IGNORE) {
```
ibgreen: Nit: Why do we use this.props here instead of nextProps? Shouldn't the most recently supplied props determine the behavior?

src/utils/transition-manager.js
```diff
@@ -0,0 +1,209 @@
...
+      this._shouldIgnoreViewportChange(nextProps)) {
+      this.props = nextProps;
+      return;
+    }
+
+    if (isTransitionEnabled) {
+      const currentViewport = this.transitionContext.viewport || extractViewportFrom(this.props);
+      const endViewport = this.transitionContext.endViewport;
+
+      const startViewport = this._shouldSnapToEnd() ?
```
ibgreen: Nit: supply nextProps to shouldSnapToEnd so that it reads behavior from the most recent props?

src/utils/transition-manager.js
1chandu: The reason I initially added `onTransitionUpdate` was to not depend on `onViewportChange` callback to perform transition. Imagine a case if app doesn't provide `onVieportChange` or provides but it doesn't change component's state hence it doesn't trigger the render().

This will binds functionality of transition with application correctly implementing `onViewportChange()`, transition don't occur otherwise.

src/utils/transition-manager.js
1chandu: 1. I am not fully convinced we need these additional checks, `_isUpdateDueToCurrentTransition()` returns early when `transitionContext.viewport` is set to null, so this should handle both case you are checking, `!isTransitionEnabled && !isTransitionInProgress`. If not can you explain the specific scenario this code is handling in more detail. We can discuss offline if it is easier.

Also even if we do need these change, these should be moved inside `_shouldIgnoreViewportChange()` method.

src/utils/transition-manager.js
1chandu: When current props has `IGNORE`, it will ignore any new viewport changes, until the current transition is complete.

src/utils/transition-manager.js
```diff
@@ -0,0 +1,209 @@
...
+  onTransitionStart: () => {},
+  onTransitionInterrupt: () => {},
+  onTransitionEnd: () => {}
+};
+
+export default class TransitionManager {
+  constructor(props) {
+    this.props = props;
+
+    this.transitionContext = {
```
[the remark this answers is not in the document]
1chandu: Yeah, I had ran into an issue when I originally calling this just `context`, hence I changed its name to `transtionContext`. This object holds all params that are valid and change during a given transition and reset after it.

src/utils/map-state.js
1chandu: > 3. In a 'reactive" environment we should of course take care not to modify input values unless we have to, but it seems to me it should not matter for viewport equality as that (i.e. _areViewportsEqual) should be done on rounded values? math.gl exports an equals(a,b) that could be used.

I though I had code in interpolator methods that return end viewport when time = 1.0 (In transition manager I enforce interpolation to end with time = 1.0), but looks like I may have moved that code, I will double check and add it back if not present.

And yes I agree,  we should use `equals(a,b)` to avoid any rounding issues.

examples/viewport-animation/src/app.js
1chandu: We don't interpolate width and height.

Pessimistress: As discussed, I removed `width` and `height` from viewport comparison.

src/components/interactive-map.js
```diff
@@ -167,10 +186,13 @@ export default class InteractiveMap extends PureComponent {
       onStateChange: this._onInteractiveStateChange,
       eventManager
     }));
+
+    this._transitionManger = new TransitionManager(this.props);
```
abmai: typo: `transitionManger` => `transitionManager`; places below as well.

src/utils/map-controls.js
abmai: Good point - it probably does. But React convention is to keep the props and state objects separate in parameters, I think we should follow that as much as we can.

src/utils/transition-manager.js
```diff
@@ -0,0 +1,180 @@
+/* global requestAnimationFrame, cancelAnimationFrame */
```
abmai: Assuming that this file and the viewport utils below are copy-pasta from deck.gl - do we have tests for them?

PR #394, opened 2017-10-28
mapbox-gl up to 0.41.0

rilyu:


abmai:
Sorry for the delay here.

This removed paint classes, as well as the associating APIs: https://github.com/mapbox/mapbox-gl-js/pull/3643/files#diff-c6d02b0c73aa7597f4224ec35497cfe2L409

I know we have directed people to those APIs in the past when they asked about paint class support in `react-map-gl`, so this is potentially a breaking change for us as well.

abmai:
Checked internally and we're in the clear here.

amitassaraf:
@abmai Hey, any update on this? When installing with NPM I still get mapbox-gl@0.40.1 installed with the latest version of react-map-gl@3.1.3

Had to move to another wrapper due to this but the new wrapper marker hover scroll problems :(

abmai:
We haven't yet released a patch to include the 0.41.0 change, will make this a priority this week.

Issue #385, opened 2017-10-20
API access token is required ERROR despite never using Mapbox Layers (introduced in 3.1)


daumann:
I recently bumped react-map-gl from "3.0.4" to "3.1.0" and my app stopped working because it throws the error **An API access token is required to use Mapbox GL** during _initialization_.

I am not using any of mapbox layers/fonts/icons. Is an access token now mandatory simply for using this package?

ibgreen:
@daumann 3.1.0 may have been published prematurely. Recommend you pin to 3.0.x while we sort this out.

Pessimistress:
@daumann The missing API access token message should not have been printed as an error, but that should not have stopped the application from rendering. Can you provide more context of how you are using react-map-gl?

Pessimistress:
@daumann Pinging again. Your custom map tiles should still render in spite of the console error. Can you tell us more about the problem?

daumann:
@ibgreen @Pessimistress I now bumped the version to **3.1.1** and while the map is working as normal, I get this notokenwarning element rendered on top of my map:

![notokenwarning](https://user-images.githubusercontent.com/5303880/32624216-de1c82b4-c588-11e7-82a8-330be9244b82.png)

Here is the relevant element: https://github.com/uber/react-map-gl/blob/5cafff8f30640e903fc43c4b6c2b5b2c21f3b2e3/src/components/static-map.js#L178-L193

Is there a way to hide/ surpress it? On a conceptual level, shouldn't be a warning in the logs be enough?

This issue was introduced here: https://github.com/uber/react-map-gl/pull/370

ibgreen:
https://github.com/uber/react-map-gl/pull/400

Pessimistress:
Fixes published in 3.1.2 patch release.

daumann:
![5newerrors](https://user-images.githubusercontent.com/5303880/32660293-8e06862c-c622-11e7-8360-df041b23e75b.png)
Unfortunately with the new 3.1.2 patch release I encounter those 5 new errors in my log (the app works though). Any idea what caused this in the new release (from 3.1.1 to 3.1.2)?

I suspected mapStyle to be malformed, but as you can see in the above image I print out mapStyle right before the errors (which get triggered when react-map-gl component is rendered) and it has version sources and layers as properties.

Pessimistress:
@daumann Thanks for reporting this, it seems that these were suppressed by the previous token error message?
I can reproduce it. Should be an easy fix.

[...]

PR #400, opened 2017-11-08
Only show access token error if Mapbox data is used

Pessimistress:
Only warn about access token in console & canvas upon an unauthorized attempt to access Mapbox service.
This will eliminate the confusing message for non-mapbox-data users such as https://github.com/uber/react-map-gl/issues/385

### DIFF
--- a/src/components/static-map.js
+++ b/src/components/static-map.js
@@ -30,6 +30,7 @@ import Mapbox from '../mapbox/mapbox';
 
 /* eslint-disable max-len */
 const TOKEN_DOC_URL = 'https://uber.github.io/react-map-gl/#/Documentation/getting-started/about-mapbox-tokens';
+const NO_TOKEN_WARNING = 'A valid API access token is required to use Mapbox data';
 /* eslint-disable max-len */
 
 function noop() {}
@@ -69,7 +70,9 @@ export default class StaticMap extends PureComponent {
       this.componentWillReceiveProps = noop;
       this.componentDidUpdate = noop;
     }
-    this.state = {};
+    this.state = {
+      accessTokenInvalid: false
+    };
     autobind(this);
   }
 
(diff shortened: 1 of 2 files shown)


src/components/static-map.js
```diff
@@ -174,16 +178,25 @@ export default class StaticMap extends PureComponent {
     this._mapboxMap = ref;
   }
 
+  // Handle map error
+  _mapboxMapError(evt) {
+    if (evt.error && evt.error.status === 401 && !this.state.accessTokenInvalid) {
```
abmai: pull `401` out into a constant for better readability if it only signifies one thing? i.e.
```js
const UNAUTHORIZED_ERROR_CODE = 401;
```

src/mapbox/mapbox.js
```diff
@@ -39,6 +39,7 @@ const propTypes = {
   attributionControl: PropTypes.bool, /** Show attribution control or not. */
   preserveDrawingBuffer: PropTypes.bool, /** Useful when you want to export the canvas as a PNG. */
   onLoad: PropTypes.func, /** The onLoad callback for the map */
+  onError: PropTypes.func, /** The onError callback for the map */
```
abmai: Outside the scope of this, but do we want to allow external overrides / ability to hook into these?

Pessimistress: Mapbox's full list of events (removed mouse/gesture events):
- resize
- remove
- webglcontextlost
- webglcontextrestored
- load
- render
- error
- data
- styledata
- sourcedata
- dataloading
- styledataloading
- sourcedataloading

Now that we expose `getMap` as a public method, the user is able to register their own listeners if they need to. Aside from `load` and `error`, I've also used `render` and `dataloading`. Not sure if all of them are useful in the React context.

abmai: That makes sense. Would registering them on the Mapbox instance override our event listener for those events, though?

Not a blocking thing for this PR btw, feel free to land whenever you want.

PR #410, opened 2017-11-22
Upgrade to viewport-mercator-project@5.0

Pessimistress:


src/utils/map-state.js
```diff
@@ -348,13 +348,13 @@ export default class MapState {
   }
 
   _unproject(pos) {
-    const viewport = new PerspectiveMercatorViewport(this._viewportProps);
-    return pos && viewport.unproject(pos, {topLeft: false});
```
Pessimistress: This was a bug in the older versions of viewport-mercator-project. `topLeft` should be `true`.

package.json
```diff
@@ -41,7 +41,7 @@
     "math.gl": ">= 1.0.0-alpha.8",
     "mjolnir.js": ">=0.4.1",
     "prop-types": "^15.5.7",
-    "viewport-mercator-project": ">= 4.2.0-alpha.2"
+    "viewport-mercator-project": "^5.0.0-alpha.1"
```
abmai: Lol, jumping from one big alpha version to another.

Issue #415, opened 2017-11-28
Map with marker always captures scroll events


storybynumbers:
It seems that markers always capture mouse events, regardless of the settings. The following map renders as expected, but I cannot scroll past it. If I remove the marker, I can.

```jsx
<ReactMapGL
    dragPan={false}
    dragRotate={false}
    captureDoubleClick={false}
    mapStyle={_config.mapboxStyle}
    onViewportChange={this.handleViewport}
    scrollZoom={false}
    {...this.state.viewport}>

    <Marker latitude={44.427824} longitude={-73.213587} offsetLeft={154 / -2}
            captureScroll={false} captureDrag={false} captureClick={false} captureDoubleClick={false}
            offsetTop={-38}>
        <div className={css.marker}>
            <IconShop width={35}/>
            <h3>Marker</h3>
        </div>
    </Marker>

</ReactMapGL>
```

tlenclos:
Same issue here, the settings `scrollZoom` just stop the scroll on the map but still capture the wheel event.

tlenclos:
It seems that the event is always captured on the BaseControl component here event when `this.props.captureScroll` equals false https://github.com/uber/react-map-gl/blob/master/src/components/base-control.js#L66.

Adding a condition on the event handler fix the issue but I don't know if this is a clean fix. Any insights @Pessimistress ?
```javascript
const events = {};

if (this.props.captureScroll) {
  events.wheel = this._onEvent;
}
```

Pessimistress:
@storybynumbers @tlenclos Thanks for reporting this problem. I'll get a fix out asap.

Pessimistress:
3.1.5

PR #422, opened 2017-12-09
Controls should not register events that it doesn't capture

Pessimistress:
https://github.com/uber/react-map-gl/issues/415

### DIFF
--- a/src/components/base-control.js
+++ b/src/components/base-control.js
@@ -70,14 +70,6 @@ export default class BaseControl extends Component {
   }
 
   _onContainerLoad(ref) {
-    this._toggleEvents(this.props, ref);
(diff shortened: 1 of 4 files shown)


src/components/base-control.js
```diff
@@ -57,46 +64,40 @@ export default class BaseControl extends Component {
...
-      dblclick: this._onEvent
-    };
+    const {eventManager} = this.context;
+    let events = this._events;
 
-    if (ref) {
-      this.context.eventManager.on(events, ref);
-    } else {
-      this.context.eventManager.off(events);
+    if (events) {
```
abmai: Add comments to explain the logic in here?

PR #433, opened 2018-01-06
Remove immutable.js dependency

Pessimistress:
Saves ~55k off a bundle without tree-shaking.


package.json
```diff
@@ -55,6 +54,7 @@
     "eslint-config-uber-jsx": "^3.0.0",
     "eslint-plugin-react": "~6.7.0",
     "gl": "^4.0.3",
+    "immutable": "*",
```
balthazar: Some of our tests/examples are using immutable still? Would it be possible not to use `*`?

PR #454, opened 2018-02-03
Add this._map sanity check in getMap()

j0nas:
Since the `_map` reference gets set to null on `componentWillUnmount`, doing a `getMap()` in my own component on unmount crashes. Currently, the only way to circumvent that is to assert on the private _map field. This should fix that.


Pessimistress:
This has been included in the 3.2.2 patch release.

PR #469, opened 2018-03-02
Fix SVGOverlay link in API Reference

rimig:


Pessimistress:
Fixed in https://github.com/uber/react-map-gl/pull/653/files

Issue #474, opened 2018-03-07
Adding / Removing Sources while a layer style still exists.


cl0ckwork:
I've been playing around with this repo for a bit, really like the immutable aspects. I believe I've run into an issue with removing a`source` with a existing `layer(s)` on the `mapStyle` object.

In the [geojson example](https://uber.github.io/react-map-gl/#/Examples/geojson) a source and layer are added, on year change the source is updated. However, if one were to remove this example source, by taking it off the `sources` field in the `mapStyle`...

for example:
```javascript

_removeSource = () => {
    const newMapStyle = this.state.mapStyle.deleteIn(['sources', 'incomeByState']);
    this.setState({ mapStyle: newMapStyle })
  }
```
the `mapStyle` updates, but mapbox-gl does not visibly remove the layer, instead the layer remains visible. I believe this is because the `layers` are dependent on the `sources` (according to this [stackoverflow](https://stackoverflow.com/questions/42478356/removing-a-source-in-mapbox-gl-js-doesnt-remove-its-layers) answer), also while the dependent layer still exists mapbox-gl it appears as though prevents the source from being removed. Essentially `map.removeLayer` needs to be called before `map.removeSource`.

In react-map-gl, if a source is marked as `exit` in `style-utils/diffStyles`, only `map.removeSource(exit.id)` is called. Thus, the associated layers were not removed (which is why the map still renders them, from what I can gather).

so, if you were to apply the following in `style-utils/setDiffStyle` (at line 69)

```javascript
for (const exit of sourcesDiff.exit) {
    const layers = nextStyle.get('layers').toJS();
    for (const lyr of layers) {
      if (lyr.source === exit.id) {
        map.removeLayer(lyr.id);
      }
    }
    map.removeSource(exit.id);
  }
```
The map then updates with the removed layer. However, adding the source + layer back seems a bit troublesome (I am unable to do so, in my trivial example). 
```jsx
// loads the souce/layer back onto the mapStyle
<button onClick={() => this._loadData(data)}>ADD SOURCE</button>
```
This is likely because the layers weren't removed from the`mapStyle`, they are only removed from the actual `map` object. So `setDiffStyle` doesn't detect a layer change on the next go around.

Here's where it gets confusing.. if I remove the source and the layers that are associated from the `mapStyle`, everythings great. The the sources are removed, the layers are removed, and the map updates. However, when I add the source back under the same id, this shows up:

```javascript
Error: There is already a source with this ID
```
[...]

PR #477, opened 2018-03-10
handle source removal without layer removal

cl0ckwork:
Attempt to remedy issue #474 .

1. Throw an error/warning if the layers were not removed with their linked sources on mapStyle change.
2. If both layers and sources were updated then apply the removals in order (layer first, source second) using existing methodology with minor order change.

### DIFF
--- a/src/utils/style-utils.js
+++ b/src/utils/style-utils.js
@@ -49,38 +49,11 @@ export function setDiffStyle(prevStyle, nextStyle, map) {
   }
 
   const {sourcesDiff, layersDiff} = diffStyles(prevStyle, nextStyle);
-
-  // TODO: It's rather difficult to determine style diffing in the presence
-  // of refs. For now, if any style update has a ref, fallback to no diffing.
-  // We can come back to this case if there's a solid usecase.
-  if (layersDiff.updates.some(node => node.layer.get('ref'))) {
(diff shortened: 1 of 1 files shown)


src/utils/style-utils.js
```diff
@@ -114,3 +87,49 @@ function updateStyleSource(map, update) {
...
+    map.addLayer(update.layer.toJS(), update.before);
+  }
+
+  for (const exit of sourcesDiff.exit) {
+    map.removeSource(exit.id);
+  }
+}
+
+/* eslint-disable max-len */
+function checkForEqualLayerSourceChanges(sourceExit, nextLayers, callback) {
```
Pessimistress: Any particular reason to use a callback here?

Pessimistress:
@cl0ckwork do you have some time to work with me and get this landed? We'd like to publish a new minor version in 1-2 weeks.

src/utils/style-utils.js
cl0ckwork: Sorry for the delayed response! I left it there in-case the check needed to be used in the future to for a different reason. I don't see an impact to dropping the callback and either placing `applySourceLayerChanges` after the check inside `checkForEqualLayerSourceChanges` or making `checkForEqualLayerSourceChanges` sync and just returning a flag if things are good.

PR #522, opened 2018-06-04
Automatically load Mapbox stylesheet

Pessimistress:


renchap:
Is it any way to disable this?

I am lazy-loading my map component (using React code-splitting) and have `import "mapbox-gl/dist/mapbox-gl.css"` in my component JS file. But the initialize function is called before the CSS file is loaded as the `<style>` tag is added to the DOM when the code is first lazy-loaded, just before being executed.

I think a simple Prop to disable automatic loading when the user knows that the CSS is properly loaded would work.

PR #540, opened 2018-07-10
Dcoument compability with create-react-app and Typescript

zackhsi:


Pessimistress:
Thank you! Question since I've been out of loop on this discussion:

- Does CRA work out of the box now? Is any special config still required?
- Could you split out the mentioning of TypeScript into a separate bullet point?

zackhsi:
- It works out of the box!
- Done!

PR #546, opened 2018-07-18
Enables extra options to be passed to mapbox-gl constructor. Fixes #545.

### LINKED ISSUE #545: Passing maxTileCacheSize to Mapbox constructor?
Hi!

I'm trying to find a way of passing the maxTileCacheSize option to the MapboxGL constructor to work around mapbox/mapbox-gl-js#4052. It doesn't seem like this is currently possible. Any ideas? Thanks!

aendrew:
In some instances it's useful to be able to pass additional options to the Mapbox GL constructor — for instance, in [this particular project](https://ig.ft.com/gb-broadband-speed-map/) we put a lot of memory pressure on iPads and other high-res tablet devices and thus being able to specify `maxTileCacheSize` would have been helpful (as it is, we had to make a custom build of react-map-gl to access this functionality).

This adds a private API via props.mapOptions enabling such options to be passed to the Map constructor — I felt that making this a public API would be confusing in relation to the other props react-map-gl passes to Mapbox GL. As such I haven't written any tests or documentation — if, however, you think this should be a public API, do let me know and I'll do so.


aendrew:
@Pessimistress Done! :+1: Let me know if you want me to explain the prop further.

PR #554, opened 2018-07-23
Upgrade Mapbox-GL-JS to 0.47

manassra:
Contains a bug fix where webpack and JS uglification cause runtime errors for Mapbox in production: https://github.com/mapbox/mapbox-gl-js/pull/6981

### DIFF
--- a/package.json
+++ b/package.json
@@ -37,7 +37,7 @@
     "babel-runtime": "^6.23.0",
     "bowser": "^1.2.0",
     "immutable": "*",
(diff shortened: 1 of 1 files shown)


holic:
Spent the morning trying to debug where this problem was coming from. Thanks for fixing this!

When do fixes like this generally get pushed out to npm?

manassra:
@holic, we just published react-map-gl@3.3.2 with those changes!

PR #575, opened 2018-08-10
Fix npm start command for development

schnerd:
I might be missing something, but it seems like `npm start` in the root directory for development is flawed:

```
npm run start

> react-map-gl@3.3.0-alpha.5 start /Users/dschnurr/Uber/react-map-gl
> (cd examples/main && (path-exists node_modules || npm i) && npm run start-local)

sh: path-exists: command not found
```

Seems like it no longer looks for `path-exists` in root node_modules/.bin after the change directory command. It's probably easier to just use a bash folder existence check here instead of an npm package.

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
     "dist"
   ],
   "scripts": {
-    "start": "(cd examples/main && (path-exists node_modules || npm i) && npm run start-local)",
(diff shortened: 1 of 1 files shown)


schnerd:
> I just ended up having start open the website on the "Running the Examples" page.

Nice that could be a good model to move to in the near future.

schnerd:
Hmm looks like a made a merge commit. Would it make sense to only allow squash commits for this repo?

image

Issue #483, opened 2018-03-20
How to drag a marker component?


CraigglesO:
I've done a lot of reading, but there does not seem to be an easy way to do this.

I tried "onDrag" with the marker, but no response. 

suggestions?

Thanks,
Craig.

Pessimistress:
@CraigglesO 

Can you share a sample of your code? The `Marker` component does not have any interactive props but listeners added to its children should generally work.

Draggable marker is a common use case and I may add an example when I get around to it...

CraigglesO:
Sorry for getting back so late! Tomorrow I will make a basic mockup! Thanks for your time!

One more question, do you guys have an example for SVG overlay? or just some basic use case via a d-path?

Thanks,
Craig.

schnerd:
@CraigglesO @pocorschi Would like to hear more specifics about your use case so we can design the appropriate API here.

Here's an example of a possible API:
```js
class App extends Component {
  constructor(props) {
    super(props);
    this.state = {
      latitude: 37.785164,
      longitude: -100
    };
  }

  onMarkerDragStart = (event) => {
    const {latitude, longitude} = event;
    // Any functionality for when a drag starts
  }

  onMarkerDragEnd = (event) => {
    const {latitude, longitude} = event;
    // Any functionality for when a drag ends
    this.setState({latitude, longitude});
  }

  onMarkerDrag = (event) => {
    const {latitude, longitude} = event;
    // Any functionality when position of element being dragged changes
  }

  render() {
    return (
      <ReactMapGL
        latitude={this.state.latitude}
        longitude={this.state.longitude}
        width={500}
        height={500}
        mapStyle="mapbox://styles/mapbox/dark-v9"
        mapboxApiAccessToken={TOKEN} >
        <DraggableMarker
          latitude={this.state.latitude}
          longitude={this.state.longitude}
          onDragStart={this.onMarkerDragStart}
          onDragEnd={this.onMarkerDragEnd}
          onDrag={this.onMarkerDrag} >
          <PinIcon />
        </DraggableMarker>
      </ReactMapGL>
    );
  }
}
```

Would this suffice for your use cases?

Pessimistress:
@schnerd  https://github.com/mapbox/mapbox-gl-js/issues/3087

PR #576, opened 2018-08-11
Add drag and drop support for Marker

schnerd:
> Resolves issue #483

This PR adds support for drag and drop events to the `<Marker>` component.

kapture 2018-08-11 at 9 44 37

Simply add a `draggable` prop to your Marker, and subscribe to whatever events you need:

```js
<Marker 
  longitude={marker.longitude}
  latitude={marker.latitude}
  draggable
  onDragStart={this._onMarkerDragStart}
  onDrag={this._onMarkerDrag}
  onDragEnd={this._onMarkerDragEnd}
>
  <Pin size={20} />
</Marker>
```

This API design is intended to be similar to [Mapbox's draggable markers](https://www.mapbox.com/mapbox-gl-js/example/drag-a-marker/).

### DIFF
--- a/docs/components/marker.md
+++ b/docs/components/marker.md
@@ -35,11 +35,35 @@ Offset of the marker from the left in pixels, negative number indicates left.
 ##### `offsetTop` {Number} - default: `0`
 Offset of the marker from the top in pixels, negative number indicates up.
 
+##### `draggable` {Boolean} - default `false`
+Allows this marker component to be dragged around the map. (Use `onDragEnd` to capture the final position and update `longitude` and `latitude`).
+
+##### `onDragStart` {Function}
+Called when a draggable marker starts being dragged.
+
+Parameters
+- `event` - The pointer event.
+  + `event.lngLat` - The geo coordinates where the drag started, as `[lng, lat]`.
+
+##### `onDrag` {Function}
+Continuously called while a draggable marker is being dragged.
+
+Parameters
+- `event` - The pointer event.
+  + `event.lngLat` - The geo coordinates of the drag event, as `[lng, lat]`.
+
+##### `onDragEnd` {Function}
+Called when a draggable marker is released at its final position. This is usually a good time to capture `event.lngLat` and update the marker's `longitude` and `latitude` props.
+
+Parameters
+- `event` - The pointer event.
+  + `event.lngLat` - The geo coordinates where the drag ended, as `[lng, lat]`.
+
 ##### `captureScroll` {Boolean} - default: `false`
 Stop propagation of mouse wheel event to the map component. Can be used to stop map from zooming when this component is scrolled.
 
 ##### `captureDrag` {Boolean} - default: `true`
-Stop propagation of dragstart event to the map component. Can be used to stop map from panning when this component is dragged.
+Stop propagation of dragstart event to the map component. Can be used to stop map from panning when this component is dragged. Automatically true if `draggable` is `true`.
 
 ##### `captureClick` {Boolean} - default: `true`
 Stop propagation of click event to the map component. Can be used to stop map from calling the `onClick` callback when this component is clicked.
--- a/examples/draggable-markers/package.json
+++ b/examples/draggable-markers/package.json
@@ -1,0 +1,20 @@
+{
+  "scripts": {
+    "start": "webpack-dev-server --progress --hot --open",
+    "start-local": "webpack-dev-server --env.local --progress --hot --open"
+  },
+  "dependencies": {
+    "react": "^16.0.0",
+    "react-dom": "^16.0.0",
+    "react-map-gl": "^3.2.0"
+  },
+  "devDependencies": {
+    "babel-core": "^6.21.0",
+    "babel-loader": "^6.2.10",
+    "babel-preset-es2015": "^6.18.0",
+    "babel-preset-react": "^6.16.0",
+    "babel-preset-stage-2": "^6.18.0",
+    "webpack": "^2.4.0",
+    "webpack-dev-server": "^2.4.0"
+  }
+}
--- a/examples/draggable-markers/webpack.config.js
+++ b/examples/draggable-markers/webpack.config.js
@@ -1,0 +1,57 @@
+// NOTE: To use this example standalone (e.g. outside of repo)
+// delete the local development overrides at the bottom of this file
+
+// avoid destructuring for older Node version support
+const resolve = require('path').resolve;
+const webpack = require('webpack');
+
+// Otherwise modules imported from outside this directory does not compile.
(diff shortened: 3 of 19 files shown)


examples/draggable-markers/package.json
```diff
@@ -0,0 +1,20 @@
...
+  "devDependencies": {
+    "babel-core": "^6.21.0",
+    "babel-loader": "^6.2.10",
+    "babel-preset-es2015": "^6.18.0",
+    "babel-preset-react": "^6.16.0",
+    "babel-preset-stage-2": "^6.18.0",
+    "webpack": "^2.4.0",
+    "webpack-dev-server": "^2.4.0"
+  }
+}
```
schnerd: We should eventually look into not having all this boilerplate for each example.

src/components/base-control.js
```diff
@@ -58,11 +58,21 @@ export default class BaseControl extends Component {
...
 
     this._onContainerLoad = this._onContainerLoad.bind(this);
   }
 
+  componentWillUnmount() {
+    const {eventManager} = this.context;
+    if (eventManager && this._events) {
+      eventManager.off(this._events);
+    }
+  }
```
schnerd: @Pessimistress Seemed like it made sense to remove events if the control is unmounted – let me know if I'm missing something.

src/components/marker.js
```diff
@@ -48,11 +53,131 @@ const defaultProps = Object.assign({}, BaseControl.defaultProps, {
...
+
+    // panstart is already attached by parent class BaseControl,
+    // here we just add listeners for subsequent drag events
+    this._dragEvents = {
+      panmove: this._onDrag.bind(this),
+      panend: this._onDragEnd.bind(this),
+      pancancel: this._onDragCancel.bind(this)
+    };
+    eventManager.on(this._dragEvents);
+  }
```
schnerd: These events need to be bound to the map itself (instead of the marker), otherwise you get behavior like this when the mouse moves outside the pin:

draggable-panmove

And it makes sense to attach these events _after_ panstart, otherwise with 1000 markers you'd have 1000 onDrag event handlers being invoked after each panmove.

test/src/components/marker.spec.js
```diff
@@ -0,0 +1,71 @@
...
+  const result = ReactTestRenderer.create(tree);
+  const resultRoot = result.root;
+
+  t.ok(resultRoot.findByProps({className: 'mapboxgl-marker '}));
+  t.equal(resultRoot.findByProps({className: 'test-marker'}).children[0], 'hello');
+
+  result.unmount();
+
+  t.end();
+});
```
schnerd: Was going to write more tests for draggable functionality specifically, but it seems like react test renderer doesn't really do `refs`, make it difficult to test the desired behavior.

examples/main/stylesheets/main.scss
```diff
@@ -245,9 +245,6 @@ hr.short {
   bottom: 24px;
   color: $white-40;
 }
-.overlays {
-  cursor: crosshair;
-}
```
schnerd: Having `crosshair` as the default cursor across the examples & website seemed like a mistake. It overrides the default cursor behavior of the library (and doesn't look great IMO)

src/components/marker.js
```diff
@@ -31,7 +31,12 @@ const propTypes = Object.assign({}, BaseControl.propTypes, {
   // Offset from the left
```
schnerd: I'm not in love with how much code ended up being required in here, but I still think it's a manageable size and broken down into small methods.

docs/components/marker.md
```diff
@@ -35,11 +35,35 @@ Offset of the marker from the left in pixels, negative number indicates left.
...
+  + `event.lngLat` - The geo coordinates where the drag started, as `[lng, lat]`.
+
+##### `onDrag` {Function}
+Continuously called while a draggable marker is being dragged.
+
+Parameters
+- `event` - The pointer event.
+  + `event.lngLat` - The geo coordinates of the drag event, as `[lng, lat]`.
+
+##### `onDragEnd` {Function}
```
ibgreen: Is there a concept of DragCancel? (ESCAPE key etc)

examples/draggable-markers/package.json
ibgreen: > We should eventually look into not having all this boilerplate for each example.

It is our goal that each example folder can be independently copied out of this repo and work stand-alone, so we do want a minimal set of "boilerplate".

That said, we really want that boilerplate/configuration to be as minimal as possible.

Unfortunately the current fashion in javascript bundlers and compilers is to over-engineer for modularity and they do not seem to offer umbrella modules (not a single unneeded development plugin must installed, even if that requires the user to add 20 lines to package.json and has no effect the final bundle size :).

Two options here are probably

* move to `babel-preset-env` which should remove two lines
* explore `create-react-app`, which now seems to be supported by `mapbox-gl`

examples/draggable-markers/webpack.config.js
```diff
@@ -0,0 +1,57 @@
...
+      include: [resolve('.')],
+      exclude: [/node_modules/],
+      use: [{
+        loader: 'babel-loader',
+        options: BABEL_CONFIG
+      }]
+    }]
+  },
+
+  resolve: {
```
ibgreen: I think this resolve is no longer needed and could be removed from examples

src/components/marker.js
ibgreen: > I'm not in love with how much code ended up being required in here, but I still think it's a manageable size and broken down into small methods.

How generic is this code? If another component wanted to also implement dragging would it be highly reusable. Could it make sense to organize it as

`Marker` > `DraggableControl` > `BaseControl`

It could look more appealing to have these methods in a separate class that is focused on dragging, rather than mixed with the Marker specific functionality.

docs/components/marker.md
schnerd: So technically there is a `pancancel` event in hammer which could theoretically warrant an onDragCancel call, but I couldn't figure out how to trigger it (dragging outside viewport, etc didn't do anything).

We could add support for escape key, though that would require another explicit keyup handler. If we're mostly trying to stay aligned with [mapbox draggable markers](https://www.mapbox.com/mapbox-gl-js/example/drag-a-marker/), they currently don't have a cancel event so it might be fine to omit for now.

src/components/marker.js
schnerd: I like this idea, let me try restructuring it that way.

examples/draggable-markers/webpack.config.js
schnerd: https://github.com/uber/react-map-gl/pull/577

PR #581, opened 2018-08-20
Provide umd bundle

TrySound:
In this diff I bundled UMD with rollup. Size snapshot plugin shows
information about different kinds of sizes + treeshakability for esm
format which I'm gonna introduce in the next PR.

UMD helps to test library in sandboxes + shows complete size of the
library, with bundled dependencies.


Pessimistress:
Any concern if we do this with webpack? I'd prefer not to inflate the dev dependencies if possible.

PR #598, opened 2018-09-22
Use class properties for binding

Pessimistress:


TrySound:
This is actually the only one syntax which makes react components easier to support. All others does not make sense for reusable components.

PR #599, opened 2018-09-22
Provide static browser/node targets

TrySound:
Ref https://github.com/uber/react-map-gl/issues/557 https://github.com/uber-web/mjolnir.js/pull/22


TrySound:
Can we publish then major mjolnir release?

TrySound:
So can this be merged? I will start working on rollup builds

Pessimistress:
@TrySound do we need to update the mjolnir.js dependency?

TrySound:
Done

digitalkaoz:
yes, a merge and a prelease would be nice :)

TrySound:
@ibgreen @Pessimistress I have a few ideas how to improve the size of this library. Will start working on it after merge of this PR.

Pessimistress:
Published 4.0.0-alpha.2

digitalkaoz:
@Pessimistress @TrySound i get this error now:

```
node_modules/react-map-gl/dist/esm/components/static-map.js

Module not found: Can't resolve '../utils/mapboxgl' in 'node_modules/react-map-gl/dist/esm/components
```

but its there :/

```js
export default null;
//# sourceMappingURL=mapboxgl.js.map
```

so whats wrong here?

TrySound:
Is it webpack?

TrySound:
And what is the target?

TrySound:
@digitalkaoz Could you try to add `./` before paths in browser section? I probably missed this nuance.

digitalkaoz:
Indeed that works! But why does the Browser section behave different than
the other sections?

Bogdan Chadkin <notifications@github.com> schrieb am Do., 4. Okt. 2018,
17:57:

TrySound:
I don't know. Probably by spec.

TrySound:
Perhaps this is done to be able to replace packages with another ones.

TrySound:
Sent a fix
https://github.com/uber/react-map-gl/pull/610

cigzigwon:
I'm guessing this is the reason my rollup component lib won't load the map? I'm making a compoennt lib consuming this lib and it will never load. You also need to implement the worker for mapbox-gl v2 or it won't compile correctly in webpack. I'm having trouble in rollup probably because the mapbox-gl lib should be a peer dep. I think no matter what I do I'm stuck.

PR #610, opened 2018-10-04
Fix browser targets

TrySound:
Ref https://github.com/uber/react-map-gl/pull/599

I made a mistake here. Sorry for that.

### DIFF
--- a/package.json
+++ b/package.json
@@ -18,9 +18,9 @@
   "module": "dist/esm/index.js",
   "esnext": "dist/es6/index.js",
   "browser": {
-    "dist/es5/utils/mapboxgl.js": "dist/es5/utils/mapboxgl.browser.js",
-    "dist/esm/utils/mapboxgl.js": "dist/esm/utils/mapboxgl.browser.js",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -18,9 +18,9 @@
   "module": "dist/esm/index.js",
   "esnext": "dist/es6/index.js",
   "browser": {
-    "dist/es5/utils/mapboxgl.js": "dist/es5/utils/mapboxgl.browser.js",
-    "dist/esm/utils/mapboxgl.js": "dist/esm/utils/mapboxgl.browser.js",
-    "dist/es6/utils/mapboxgl.js": "dist/es6/utils/mapboxgl.browser.js"
+    "./dist/es5/utils/mapboxgl.js": "./dist/es5/utils/mapboxgl.browser.js",
```
Pessimistress: Do we actually need `mapboxgl.browser.js`? Does `mapbox-gl` works here?

TrySound: Yes, `mapbox-gl` is imported in browser version (for electron renderer and web), null is exported for node and electron main.

Issue #561, opened 2018-08-03
v4.0 Goals


Pessimistress:
### Core Components
- [x] Remove Immutable from dependency
- [x] Remove implementation of style diffing: use Mapbox's own style diffing feature
- [x] Update or remove the usage of the legacy React context API
- [x] Support relative `width` and `height` props

### Interaction
- [x] Offer flags for pan/zoom/rotate in callbacks
- [x] Offer APIs to access more pointer events

### Build System
- [x] Upgrade babel to v7
- [x] Remove manual binding of class methods

### Other
- [ ] Improve testing

TrySound:
I would start using babel-runtime. Currently helpers are duplicated in every module.

tstirrat15:
@Pessimistress is there any busywork-type work that I can do to help move towards this release?

sos0:
Relative 👏 width 👏 and 👏  height 👏 🎉 🎈

Pessimistress:
4.0.0-beta.1 is published

archansel:
Hi, Does v4 support react below v16? I tried using v4.0.0-beta.3 and got `_react.createContext` error. My react version is `15.6.1`

tstirrat15:
@archansel No. One of the goals of v4 is to move towards the new React context API (released in v16.3), which is incompatible with the old API.

Pessimistress:
4.0.0 is published!

TrySound:
@Pessimistress Where is tag and changelog?

Pessimistress:
@archansel Some of the v15 API that v3 was using will be removed in the next major release of React. Although unfortunate, when we have to choose, we'd like the latest react-map-gl to support 17 rather than 15.

Pessimistress:
@TrySound 4.0-release branch

PR #611, opened 2018-10-08
WIP Update the usage of the legacy React context API

XiaokaiUber:
For #561

### DIFF
--- a/examples/additional-overlays/choropleth-overlay.js
+++ b/examples/additional-overlays/choropleth-overlay.js
@@ -1,3 +1,4 @@
+/* eslint-disable no-unused-vars */
 // Copyright (c) 2015 Uber Technologies, Inc.
 
 // Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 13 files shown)


examples/additional-overlays/choropleth-overlay.js
```diff
@@ -114,26 +113,38 @@ export default class ChoroplethOverlay extends Component {
...
-          width: `${width}px`,
-          height: `${height}px`,
-          position: 'absolute',
-          pointerEvents: 'none',
-          opacity: this.props.globalOpacity,
-          left: 0,
-          top: 0
-        } }/>
+      createElement(StaticContext.Consumer,
+        null, sContext => {
```
TrySound: Why not jsx here? Looks inconsistent.

Pessimistress: Legacy build system. I will upgrade it in another PR.

Pessimistress: Hmm didn't see that it was an example. Examples already can transpile JSX. There should not be any usage of context components on the application level.

PR #613, opened 2018-10-10
Upgrade to new React context API

Pessimistress:
React plans to remove the legacy context API in v17: https://reactjs.org/docs/context.html

Note: additional-overlay example is broken, will fix in a follow up PR.

@XiaokaiUber

### DIFF
--- a/src/components/base-control.js
+++ b/src/components/base-control.js
@@ -52,13 +52,13 @@ export default class BaseControl extends PureComponent {
   constructor(props) {
     super(props);
 
-    this.context = {};
+    this._context = {};
     this._events = null;
     this._containerRef = null;
   }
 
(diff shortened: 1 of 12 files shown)


src/components/base-control.js
```diff
@@ -39,24 +40,19 @@ const defaultProps = {
...
  * is almost always triggered by a viewport change, we almost definitely need to
  * recalculate the marker's position when the parent re-renders.
  */
-export default class BaseControl extends Component {
+export default class BaseControl extends PureComponent {
 
   constructor(props) {
     super(props);
 
+    this.context = {};
```
TrySound: This may be broken because `context` field is used by old context api. I would use another name.

PR #614, opened 2018-10-11
Support relative map size

Pessimistress:
For #561 

### Change list
- Accept CSS strings for width/height props
- Add `onResize` callback to `StaticMap` and `InteractiveMap`

Tested DOM/map/overlay synchronization with the map control and deck.gl examples

@TrySound

### DIFF
--- a/src/components/static-map.js
+++ b/src/components/static-map.js
@@ -23,6 +23,7 @@ import PropTypes from 'prop-types';
 import {normalizeStyle} from '../utils/style-utils';
 
 import WebMercatorViewport from 'viewport-mercator-project';
+import AutoSizer from 'react-virtualized-auto-sizer';
 
 import Mapbox from '../mapbox/mapbox';
 import mapboxgl from '../utils/mapboxgl';
@@ -41,7 +42,21 @@ function noop() {}
 
 const UNAUTHORIZED_ERROR_CODE = 401;
 
+const CONTAINER_STYLE = {
+  position: 'absolute',
+  width: '100%',
+  height: '100%',
+  overflow: 'hidden'
+};
+
 const propTypes = Object.assign({}, Mapbox.propTypes, {
+  /** The dimensions of the map **/
+  width: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
+  height: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
+
+  /** Callback when map size changes **/
+  onResize: PropTypes.func,
+
   /** The Mapbox style. A string url or a MapboxGL style Immutable.Map object. */
   mapStyle: PropTypes.oneOfType([
     PropTypes.string,
@@ -63,7 +78,8 @@ const propTypes = Object.assign({}, Mapbox.propTypes, {
 const defaultProps = Object.assign({}, Mapbox.defaultProps, {
   mapStyle: 'mapbox://styles/mapbox/light-v8',
   preventStyleDiffing: false,
-  visible: true
+  visible: true,
+  onResize: noop
 });
 
 export default class StaticMap extends PureComponent {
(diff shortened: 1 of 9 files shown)


src/components/static-map.js
```diff
@@ -98,22 +118,8 @@ export default class StaticMap extends PureComponent {
...
-      width: this.props.width,
-      height: this.props.height
-    });
-  }
-
-  componentDidUpdate() {
-    // Since Mapbox's map.resize() reads size from DOM
-    // we must wait to read size until after render (i.e. here in "didUpdate")
-    this._updateMapSize(this.state, this.props);
+    this._updateMapProps(newProps);
```
TrySound: Can this be moved to `componentDidUpdate`?

src/components/static-map.js
```diff
@@ -192,51 +207,59 @@ export default class StaticMap extends PureComponent {
...
+          key: 'map-mapbox',
+          ref: this._mapboxMapLoaded,
+          style: mapStyle,
+          className
+        }),
+        createElement(AutoSizer, {
+          key: 'autosizer',
+          disableWidth: Number.isFinite(width),
+          disableHeight: Number.isFinite(height)
+        }, this._renderOverlays.bind(this)),
```
TrySound: I would do this in constructor or just use class properties.

Pessimistress: This is subtle... We need `AutoSizer` to rerender when either the props of `StaticMap` change or the container is resized. But it's a pure component. `bind` makes a new child every render pass and triggers that rerender.

src/mapbox/mapbox.js
```diff
@@ -250,6 +254,14 @@ export default class Mapbox {
 
     this._create(props);
 
+    // Hijack dimension properties
+    // This eliminates the timing issue between calling resize() and DOM update
+    const {container} = props;
+    container.__defineGetter__('offsetWidth', () => this.width);
+    container.__defineGetter__('clientWidth', () => this.width);
+    container.__defineGetter__('offsetHeight', () => this.height);
+    container.__defineGetter__('clientHeight', () => this.height);
```
schnerd: Clever. 

Would Object.defineProperty be better here? Sounds like defineGetter is deprecated.
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/__defineGetter__

test/src/components/map.spec.js
```diff
@@ -155,7 +155,8 @@ TEST_CASES.forEach(testCase => {
   });
 });
 
-test('Interactive map renders children on first render', t => {
+// children are no longer rendered on first pass due to auto sizer
+test.skip('Interactive map renders children on first render', t => {
```
schnerd: Might as well delete the test?

wafisher:
How can I have the component fill its space and _keep_ filling its space when resized? I could attach a function to `resize` but what would it do, set `height: '100%', width: '100%'` again, which wouldn't cause a re-render.

wafisher:
Just solved my issue. The problem was this code:

```
<ReactMapGL
  {...this.state.viewport}
  [...other props...]
/>
```

The `onViewportChange` would give real numbers (not "100%") for width and height, which would then set the component with those numbers, not CSS styles. I changed by keeping width/height out of viewport and setting explicitly on `ReactMapGL` component.

invisibleroads:
Here is how we solved it, thanks to @wafisher's comment that `onViewportChange` tries to explicitly set width and height. 

```
class Map extends Component {
  state = { 
    viewport: {
    },
  } 

  onViewportChange = viewport => { 
    const {width, height, ...etc} = viewport
    this.setState({viewport: etc})
  } 

  render () { 
    const { 
      viewport,
    } = this.state
    return ( 
      <ReactMapGL
        width='100%'
        height='100%'
        {...viewport}
        mapboxApiAccessToken={process.env.REACT_APP_MAPBOX_TOKEN}
        onViewportChange={viewport => this.onViewportChange(viewport)}
      />
    ) 
  } 
}
```
@salah93

PR #615, opened 2018-10-12
Refactor/remove bower dependency

Scarysize:
This removes `bowser` as a (somewhat heavy) dependency. The prefix for the cursor `grab` and `grabbing` CSS attributes are not necessary anymore.

- Chrome, Firefox & Edge work with the un-prefixed versions
- Safari won't work with with either

AFAIK this is only used in `<InteractiveMap />`:

https://github.com/uber/react-map-gl/blob/28c38848f0653ed1b2d0b6be363bb6e318c7ac47/src/components/interactive-map.js#L119-L121

which is called here:

https://github.com/uber/react-map-gl/blob/28c38848f0653ed1b2d0b6be363bb6e318c7ac47/src/components/interactive-map.js#L312-L317

An alternativ solution might be to just have a CSS class on the element: 

https://github.com/uber/react-map-gl/blob/28c38848f0653ed1b2d0b6be363bb6e318c7ac47/src/components/interactive-map.js#L324

this would allow for fallback values via CSS:

```
.event-canvas {
  /* ... */
  cursor: grab
  cursor: -moz-grab
  cursor: -webkit-grab
}
```


Pessimistress:
@Scarysize if you could sign the CLA I'll include this change in the next release.

Scarysize:
Signed the CLA and solved the merge conflict. Thanks for the fast response! Any idea when the next release will be? We can use the github branch as a npm dependency for now, but an official release would great.

Pessimistress:
Published 3.3.6

Scarysize:
Thanks a lot for resolving this so quickly, appreciate it!

PR #619, opened 2018-10-14
Bug fixes for auto-resize mode

Pessimistress:
- Fix: onHover and onClick errors if width/height are not numbers
- Fix: `mapbox.resize()` error before map is loaded

### DIFF
--- a/src/components/interactive-map.js
+++ b/src/components/interactive-map.js
@@ -282,7 +282,10 @@ export default class InteractiveMap extends PureComponent {
       }
 
       if (this.props.onHover) {
-        const viewport = new WebMercatorViewport(this.props);
+        const viewport = new WebMercatorViewport(Object.assign({}, this.props, {
+          width: this._width,
+          height: this._height
+        }));
         event.lngLat = viewport.unproject(pos);
         event.features = features;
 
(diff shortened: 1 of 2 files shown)


src/components/interactive-map.js
```diff
@@ -282,7 +282,10 @@ export default class InteractiveMap extends PureComponent {
       }
 
       if (this.props.onHover) {
-        const viewport = new WebMercatorViewport(this.props);
+        const viewport = new WebMercatorViewport(Object.assign({}, this.props, {
+          width: this._width,
```
abmai: This will always override width/height with the internal values?

src/mapbox/mapbox.js
```diff
@@ -149,7 +149,11 @@ export default class Mapbox {
   // (e.g. until "componentDidUpdate")
   resize() {
     this._map.resize();
-    this._map._render();
+    try {
```
abmai: Is it possible to instead check if map style is loaded before calling `_render`?

src/components/interactive-map.js
Pessimistress: Yes. Internal values will always reflect the numeric dimension of the map.

abmai:
👍

Issue #612, opened 2018-10-09
double-click zooming follows a strange arc path


vbud:
We've been trying to understand why double-click zooming behaves so much different in react-map-gl than in mapbox-gl. Observe the difference by double-click zooming on these two examples:
https://www.mapbox.com/mapbox-gl-js/example/simple-map/
https://uber.github.io/react-map-gl/#/Examples/dynamic-styling

The mapbox-gl example translates between point A and point B as you would expect. The react-map-gl example starts and ends at the same points, but it follows a strange arcing path to get there.

Is this a bug? Or something we could override?

Pessimistress:
Published 3.3.7

vbud:
Thank you so much @Pessimistress!

PR #625, opened 2018-10-18
Handle zoom center during linear transition

Pessimistress:
https://github.com/uber/react-map-gl/issues/612

### DIFF
--- a/src/utils/transition/linear-interpolator.js
+++ b/src/utils/transition/linear-interpolator.js
@@ -34,6 +34,13 @@ export default class LinearInterpolator extends TransitionInterpolator {
     const startViewportProps = {};
     const endViewportProps = {};
 
+    if (this.around) {
+      Object.assign(endViewportProps, endProps, {
+        around: this.around,
+        aroundLngLat: new WebMercatorViewport(endProps).unproject(this.around)
+      });
+    }
+
     for (const key of this.propNames) {
       const startValue = startProps[key];
       const endValue = endProps[key];
@@ -41,15 +48,6 @@ export default class LinearInterpolator extends TransitionInterpolator {
 
       startViewportProps[key] = startValue;
       endViewportProps[key] = getEndValueByShortestPath(key, startValue, endValue);
-    }
-
-    if (this.around) {
-      Object.assign(endViewportProps, {
-        width: endProps.width,
-        height: endProps.height,
-        around: this.around,
-        aroundLngLat: new WebMercatorViewport(endProps).unproject(this.around)
-      });
     }
 
     return {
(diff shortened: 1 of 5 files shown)


src/utils/transition/linear-interpolator.js
```diff
@@ -11,17 +13,34 @@ const VIEWPORT_TRANSITION_PROPS = ['longitude', 'latitude', 'zoom', 'bearing', '
...
-   * @param {Array} transitionProps - list of props to apply linear transition to.
+   * @param {Array} [transitionProps] - list of props to apply linear transition to.
+   * @param {Array} [opts] - additional options.
    */
-  constructor(transitionProps = VIEWPORT_TRANSITION_PROPS) {
+  constructor(transitionProps, opts) {
     super();
-    this.propNames = transitionProps;
+
+    if (Array.isArray(transitionProps)) {
```
1chandu: Feels hacky to differentiate between one arg vs two args, can we make constructor take an object? If this leads to a bigger change leave a comment here.

Pessimistress: This fix needs to go in as a patch. We can change the API in the next major release.

Pessimistress: Though I can make it take only one argument -- either an array (old) or an object (new)

src/utils/transition/linear-interpolator.js
```diff
@@ -42,6 +61,18 @@ export default class LinearInterpolator extends TransitionInterpolator {
     for (const key of this.propNames) {
       viewport[key] = lerp(startProps[key], endProps[key], t);
     }
+
+    if (endProps.around) {
+      // zoom around provided point
+      const [longitude, latitude] = new WebMercatorViewport(Object.assign({}, endProps, viewport))
+        .getMapCenterByLngLatPosition({
```
1chandu: Can we apply this logic and calculate lng/lat during  https://github.com/uber/react-map-gl/blob/837588c9272e254f7c86941d02ca2705f058c6a4/src/utils/map-controls.js#L364

1chandu: This might guarantee the screen/lng-lat values match only at ends, but not sure if the arc style transition gets fixed or not.

Pessimistress: The start/end of the transition was already calculated correctly. The problem is that the transition of longitude/latitude is not linear during the transition. When we treated them as linear we get the weird arc.

1chandu: I see, thanks for the details, looks good to me, already approved.

Issue #569, opened 2018-08-08
Click event isn't fired on the original map


vkammerer:
Since https://github.com/uber/react-map-gl/pull/565 landed in 3.3.4, the original map layer doesn't capture clicks anymore

```javascript
const map = this.mapRef.getMap();
map.on('click', e => {
  console.log(e);
  // nothing gets logged
});
```

This makes it impossible to use certain features, such as [handling clicks on clusters](https://www.mapbox.com/mapbox-gl-js/example/cluster/) for example.

Is there a way to get the clicks to be passed to the original map layer?

Pessimistress:
Why do you need to register your own event listener? Does the `onClick` prop not work?

vkammerer:
My use case is that I want to be able to [handle clicks on clusters](https://www.mapbox.com/mapbox-gl-js/example/cluster/), but I guess there could be many others..

Pessimistress:
This module is a React wrapper for mapbox-gl so you are encouraged to use the React API wherever possible. There is no guarantee for any mapbox-gl native examples to work.

In your use case, please use the [`onClick` prop](https://uber.github.io/react-map-gl/#/Documentation/api-reference/interactive-map?section=properties) to access the features under the pointer.

vkammerer:
> There is no guarantee for any mapbox-gl native examples to work

Ok wow, this is a very strong statement. Thank you for the clarification. I believe this will be a major blocker for many teams, as it means that once you opt in for this library, you cannot benefit from the original API.

This should definitely be stated at the top of the README to warn users.

Pessimistress:
To be clear, here is how you can capture a click event and recreate that example:

```js
_onClick(event) {
  // either
  const feature = event.features.find(f => f.layer.id === 'clusters');
  // or
  const point = [event.center.x, event.center.y];
  const feature = this.mapRef.queryRenderedFeatures(point, { layers: ['clusters'] })[0];

  if (feature) {
    // look up cluster expansion zoom
  }
}

render() {
  return <MapGL onClick={this._onClick} ... />;
}
```

> this is a very strong statement.

Not really. Front and foremost, the goal of this library is to provide an interface that conforms with the React programming conventions. Similarly, you cannot go find a vanilla HTML/JavaScript code snippet and expect it to work as-is with React.

[...]

PR #628, opened 2018-10-20
Remove click delay in map controls when `captureDoubleClick` is on

Pessimistress:
Reported in https://github.com/uber/react-map-gl/issues/569

Avoid registering the `dblclick` event as it adds a 300ms delay before firing `click`.

### DIFF
--- a/src/components/base-control.js
+++ b/src/components/base-control.js
@@ -112,9 +112,9 @@ export default class BaseControl extends PureComponent {
   _onPointerUp = (evt) => {
     if (this.props.captureDoubleClick) {
       const {eventManager} = this._context;
-      const doubletap = eventManager.manager.get('doubletap');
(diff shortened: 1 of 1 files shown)


src/components/base-control.js
```diff
@@ -109,14 +109,23 @@ export default class BaseControl extends PureComponent {
     }
   }
 
-  _onClick = (evt) => {
-    if (this.props.captureClick) {
-      evt.stopPropagation();
+  _onPointerUp = (evt) => {
+    if (this.props.captureDoubleClick) {
+      const {eventManager} = this._context;
+      const doubletap = eventManager.manager.get('doubletap');
```
abmai: Why not go all-in with the destructuring here since we're already doing it above? Otherwise, should we check for null when accessing `doubletap.options.eanble`?

```suggestion
      const {options: {enable}} = eventManager.manager.get('doubletap');
```

Issue #567, opened 2018-08-08
TRANSITION_EVENTS.JOIN ?


cyrilchapon:
In here we'd like to achieve the following :

The user enters an adress :
- We perform an ajax query to reverse-geocode it (we get a point)
- We fly to that point with a transition
- **while** we do this, we perform another ajax query to get house boundaries of the user (we get a bbox)
- If the transition is still going, we "adjust" the ongoing transition to "finish" it with a sort fitBounds on the bbox
- If the transition is finished, we start a new one

Currently, I'm afraid the first use-case "adjusting" a transitition is not possible, am I right ?

I tried the following :

- Go for a transition (with the "set and forget" strategy)

```
    this.setState(prevState => ({
      viewport: {
        ...prevState.viewport,
        longitude: locationGeometry.coordinates[0],
        latitude: locationGeometry.coordinates[1],
        transitionDuration: MAP_LARGE_ANIMATION_DURATION,
        transitionInterpolator: new FlyToInterpolator(),
        transitionEasing: MAP_LARGE_ANIMATION_TRANSITION,
        transitionInterruption: TRANSITION_EVENTS.BREAK
      }
    }))
```

Then (1s later) change the viewport, without resetting transition props
```
    this.setState(prevState => {
      const parcelFitViewport = //Calculate viewport

      const {
        longitude,
        latitude,
        zoom
      } = parcelFitViewport

      return ({
        viewport: {
          ...prevState.viewport,
          longitude,
          latitude,
          zoom: zoom,
          // This is false in that use-case
          ...(prevState.initialZoomFinished ? {
            transitionDuration: MAP_SMALL_ANIMATION_DURATION,
            transitionInterpolator: new FlyToInterpolator(),
            transitionEasing: MAP_LARGE_ANIMATION_TRANSITION,
            transitionInterruption: TRANSITION_EVENTS.SNAP_TO_END
          } : {})
        }
      })
    })
```

But that doesn't work at all. The new viewport change **respects** the transition props, even if I don't re-set them. I also tried the 3 of the other transitionInteruptions strategies, but none of them accomplish what I want.

The troubles are :
- When I `setState(fitBoundsViewport)`, this entirely start a new transition (with the amount of time I set before), so if the first transition is 3000ms, with 2500 already elapsed, then the total transition time is roughly 5500ms if you follow me
- Therefore, I thought of "calculating the remaining time of the ongoing transition", and setting this time for the new one... but :
[...]

PR #631, opened 2018-10-22
added new interruption mode: UPDATE

XiaokaiUber:
For #567

#### Change List
- added new interruption mode which allows the transition to keep going on its current path smoothly when the destination is updated

### DIFF
--- a/docs/components/interactive-map.md
+++ b/docs/components/interactive-map.md
@@ -181,10 +181,11 @@ Easing function that maps a value from `[0, 1]` to `[0, 1]`. Default to `t => t`
 
 ##### `transitionInterruption` {Number}
 
-What to do if an ongoing transition is interrupted by another transition. There are 3 options:
+What to do if an ongoing transition is interrupted by another transition. There are 4 options:
 - `TRANSITION_EVENTS.BREAK` - Start new transition from the current view. This is the default.
 - `TRANSITION_EVENTS.SNAP_TO_END` - Jump to the end of the previous transition before starting the new transition.
 - `TRANSITION_EVENTS.IGNORE` - Complete the previous transition and ignore the new viewport change.
+- `TRANSITION_EVENTS.UPDATE` - Continue the ongoing transition but change its path from the current view toward the end of the interrupting transition.
 
 You may import the constants as follows:
 
@@ -210,7 +211,7 @@ Callback that is fired when the map's viewport properties should be updated.
 onViewportChange(viewState, interactionState, oldViewState);
 ```
 
-Arguments: 
+Arguments:
 
 - `viewState` {Object} The next viewport properties, including: `width`, `height`, `latitude`, `longitude`, `zoom`, `bearing`, `pitch`, `altitude`, `maxZoom`, `minZoom`, `maxPitch`, `minPitch`, `transitionDuration`, `transitionEasing`, `transitionInterpolator`, `transitionInterruption`.
 - `interactionState` {Object} The current interaction that caused this viewport change. See `onInteractionStateChange` for possible fields.
--- a/docs/components/interactive-map.md
+++ b/docs/components/interactive-map.md
@@ -185,7 +185,7 @@ What to do if an ongoing transition is interrupted by another transition. There 
 - `TRANSITION_EVENTS.BREAK` - Start new transition from the current view. This is the default.
 - `TRANSITION_EVENTS.SNAP_TO_END` - Jump to the end of the previous transition before starting the new transition.
 - `TRANSITION_EVENTS.IGNORE` - Complete the previous transition and ignore the new viewport change.
-- `TRANSITION_EVENTS.UPDATE` - Continue the ongoing transition but change its path from the current view toward the end of the interrupting transition.
+- `TRANSITION_EVENTS.UPDATE` - Continue the ongoing transition but change its destination to the new viewport.
 
 You may import the constants as follows:
 
--- a/src/utils/transition-manager.js
+++ b/src/utils/transition-manager.js
@@ -8,7 +8,8 @@ const noop = () => {};
 export const TRANSITION_EVENTS = {
   BREAK: 1,
   SNAP_TO_END: 2,
-  IGNORE: 3
+  IGNORE: 3,
+  UPDATE: 4
 };
 
 const DEFAULT_PROPS = {
@@ -27,7 +28,8 @@ const DEFAULT_STATE = {
   animation: null,
   propsInTransition: null,
   startProps: null,
-  endProps: null
+  endProps: null,
+  completion: 0
 };
 
 export default class TransitionManager {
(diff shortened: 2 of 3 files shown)


src/utils/transition-manager.js
```diff
@@ -175,15 +195,16 @@ export default class TransitionManager {
     const {startTime, duration, easing, interpolator, startProps, endProps} = this.state;
 
     let shouldEnd = false;
-    let t = (currentTime - startTime) / duration;
+    let t = (currentTime - startTime) / duration + this.state.completion;
```
Pessimistress: This is incorrect. When we update the transition from A -> B to A -> C, what should happen at the break point is:

image

Your current solution jumps:
image

src/utils/transition-manager.js
```diff
@@ -132,15 +134,19 @@ export default class TransitionManager {
       isRotating: startProps.bearing !== endProps.bearing ||
         startProps.pitch !== endProps.pitch
     };
-
+    const currentTime = Date.now();
+    const completion = this.state.interruption === TRANSITION_EVENTS.UPDATE ?
+                      (currentTime - this.state.startTime) / startProps.transitionDuration : 0;
+    const totalCompletion = this.state.totalCompletion + completion;
```
Pessimistress: What happens if `this.state.totalCompletion` is non-zero and you interrupt it with BREAK mode?

src/utils/transition-manager.js
```diff
@@ -132,15 +134,19 @@ export default class TransitionManager {
...
     this.state = {
       // Save current transition props
-      duration: endProps.transitionDuration,
+      totalCompletion,
+      duration: endProps.transitionDuration * (1 - totalCompletion),
```
Pessimistress: UPDATE mode needs to use duration, easing and interpolators from the previous transition.

src/utils/transition-manager.js
```diff
@@ -173,17 +179,17 @@ export default class TransitionManager {
...
     let t = (currentTime - startTime) / duration;
     if (t >= 1) {
       t = 1;
       shouldEnd = true;
+      this.state.totalCompletion = 0;
     }
-    t = easing(t);
-
+    const tc = this.state.totalCompletion;
+    t = (easing(t * (1 - tc) + tc) - easing(tc)) / (1 - tc);
```
Pessimistress: This does not end with 1 when t is 1. Please add test cases.

src/utils/transition-manager.js
```diff
@@ -132,22 +138,30 @@ export default class TransitionManager {
       isRotating: startProps.bearing !== endProps.bearing ||
         startProps.pitch !== endProps.pitch
     };
-
+    const currentTime = Date.now();
+    let newDuration = endProps.transitionDuration;
+    let newEasing = endProps.transitionEasing;
+    let newInterpolator = endProps.transitionInterpolator;
+    if (this.state.interruption === TRANSITION_EVENTS.UPDATE) {
```
Pessimistress: This should be moved into `processViewportChange`, where `TRANSITION_EVENTS .SNAP_TO_END` is handled

test/src/utils/transition-manager.spec.js
```diff
@@ -116,3 +118,24 @@ test('TransitionManager#callbacks', t => {
...
+  t => t * t,
+  t => t * t * t];
+const interruptions = [0.2, 0.5, 0.8];
+const values = [0, 0.5, 1];
+
+test('cropEasingFunction', function (t) {
+  easingFunctions.forEach(func => {
+    interruptions.forEach(x0 => {
+      var newEasing = cropEasingFunction(func, x0);
+      values.forEach(val => {
```
Pessimistress: Explicitly test `newEasing(0) === 0` and `newEasing(1) === 1`

docs/components/interactive-map.md
```diff
@@ -181,10 +181,11 @@ Easing function that maps a value from `[0, 1]` to `[0, 1]`. Default to `t => t`
 
 ##### `transitionInterruption` {Number}
 
-What to do if an ongoing transition is interrupted by another transition. There are 3 options:
+What to do if an ongoing transition is interrupted by another transition. There are 4 options:
 - `TRANSITION_EVENTS.BREAK` - Start new transition from the current view. This is the default.
 - `TRANSITION_EVENTS.SNAP_TO_END` - Jump to the end of the previous transition before starting the new transition.
 - `TRANSITION_EVENTS.IGNORE` - Complete the previous transition and ignore the new viewport change.
+- `TRANSITION_EVENTS.UPDATE` - Continue the ongoing transition but change its path from the current view toward the end of the interrupting transition.
```
Pessimistress: `change its destination to the new viewport.`

src/utils/transition-manager.js
```diff
@@ -5,10 +5,16 @@ import MapState from './map-state';
 
 const noop = () => {};
 
+export function cropEasingFunction(easing, x0) {
```
Pessimistress: Add comment on what this function does

src/utils/transition-manager.js
```diff
@@ -65,9 +73,17 @@ export default class TransitionManager {
...
       const startProps = Object.assign({}, currentProps,
         this.state.interruption === TRANSITION_EVENTS.SNAP_TO_END ?
-        this.state.endProps : this.state.propsInTransition
+          this.state.endProps : this.state.propsInTransition
       );
 
+      const currentTime = Date.now();
+      if (this.state.interruption === TRANSITION_EVENTS.UPDATE) {
+        const x0 = (currentTime - this.state.startTime) / this.state.duration;
+        nextProps.transitionDuration = this.state.duration - (currentTime - this.state.startTime);
```
Pessimistress: Do not mutate an argument object - this may cause unexpected side effects for whoever calls this method.

test/src/utils/transition-manager.spec.js
```diff
@@ -1,5 +1,10 @@
 import test from 'tape-catch';
 import TransitionManager from 'react-map-gl/utils/transition-manager';
+import {equals} from 'math.gl';
+import {cropEasingFunction} from '../../../src/utils/transition-manager';
```
Pessimistress: Import from `react-map-gl` alias

test/src/utils/transition-manager.spec.js
```diff
@@ -116,3 +121,108 @@ test('TransitionManager#callbacks', t => {
...
+        height: 100,
+        longitude: -122.45,
+        latitude: 37.78,
+        zoom: 12, pitch: 0,
+        bearing: 0,
+        transitionDuration: 2000,
+        transitionEasing: t => t,
+        transitionInterpolator: new LinearInterpolator()}
+    ],
+    expect: [
```
Pessimistress: Use keys instead of index, e.g.

```
{
  TRANSITION_EVENTS.BREAK: {
    duration: false,
    transitionInterpolator: true
  }
}
```

Issue #642, opened 2018-10-31
onViewportChange Warning


joseomar10:
I'm testing an example of the library with the version "4.0.0-beta.3",  but I have this warning to method onViewportChange: 

Warning: Cannot update during an existing state transition (such as within `render`). Render methods should be a pure function of props and state.

![screen shot 2018-10-31 at 10 41 18 am](https://user-images.githubusercontent.com/8315391/47800207-d0a14f00-dcf9-11e8-97e4-0548036ef0de.png)

allenchuang:
@Pessimistress This issue is still happening....

https://github.com/uber/react-map-gl/issues/872

leonardoviveiros:
Still happening in version `5.0.11`

JonathanJohnson23:
Yup Still a bug

RELNO:
This is still a bug @ `"react-map-gl": "^5.1.3"`

mapsgeek:
i still get this when i set the `height` and `width` to `100%` or a `100vh,100vw` @`^6.1.11`

PR #643, opened 2018-10-31
Do not call onResize during render

Pessimistress:
For #642

### DIFF
--- a/src/components/static-map.js
+++ b/src/components/static-map.js
@@ -154,7 +154,6 @@ export default class StaticMap extends PureComponent {
       this._width = width;
       this._height = height;
       this._updateMapProps(this.props);
-      this.props.onResize({width, height});
     }
   }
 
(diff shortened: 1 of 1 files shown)


pkayokay:
`OnResize` still seems to be called once on render

```
        <ReactMapGL
          {...this.state.viewport}
          height="100%"
          width="100%"
          attributionControl={false}
          mapboxApiAccessToken={this.props.popmenuConfig.mapboxPublicKey}
          mapStyle="mapbox://styles/mapbox/streets-v9"
          ref={this.mapRef}
          onLoad={() => {
            if (this.props.locations.length >= 2) {
              this.setBounds();
            }
          }}
          onResize={() => {
            console.log('onResize');
          }}
        >
          {this.props.children}
        </ReactMapGL>
```

console shows "onResize" upon site refresh

PR #651, opened 2018-11-04
Set up flow

Pessimistress:


package.json
```diff
@@ -29,6 +29,7 @@
   "scripts": {
     "start": "(cd examples/main && (if [ ! -d 'node_modules' ]; then npm i; fi) && npm run start-local)",
     "build": "./scripts/build.sh && ./scripts/collect-metrics.sh",
+    "flow": "flow check",
```
TrySound: I recommend to use another name like `typecheck`. Running `yarn flow` in development process will take too much time with this script. The original one runs the server.

src/components/base-control.js
```diff
@@ -84,19 +89,24 @@ export default class BaseControl extends PureComponent {
     }
   }
 
-  _onScroll = (evt) => {
+  _context: any = {};
+  _events: any = null;
+  /* global HTMLDivElement */
```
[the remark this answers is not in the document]
TrySound: Yeah, flowtype will help. I found only these two rules useful though.
```js
    'flowtype/define-flow-type': 'error',
    'flowtype/use-flow-type': 'error',
```

flow-typed/npm/mapbox-gl_vx.x.x.js
```diff
@@ -0,0 +1,18 @@
...
+ *
+ * Fill this stub out by replacing all the `any` types.
+ *
+ * Once filled out, we encourage you to share your work with the
+ * community by sending a pull request to:
+ * https://github.com/flowtype/flow-typed
+ */
+
+declare module 'mapbox-gl' {
+  declare module.exports: any;
```
akre54: Doesn't mapbox.gl export its flowtypes?

src/components/interactive-map.js
```diff
@@ -94,12 +99,8 @@ const propTypes = Object.assign({}, StaticMap.propTypes, {
...
 
   // A map control instance to replace the default map controller
-  // The object must expose one property: `events` as an array of subscribed
-  // event names; and two methods: `setState(state)` and `handle(event)`
-  controller: PropTypes.shape({
-    events: PropTypes.arrayOf(PropTypes.string),
-    handleEvent: PropTypes.func
-  })
+  // The object must expose a method: `setOptions(opts)`
+  controller: PropTypes.instanceOf(MapController)
```
akre54: Follow-up: should add a babel transform to convert flow types to proptypes for the dist file? That way we get type safety by default but don't break API

src/components/static-map.js
```diff
@@ -204,14 +251,18 @@ export default class StaticMap extends PureComponent {
...
 
-  _renderOverlays(dimensions) {
+  _renderOverlays(dimensions : {
+    width?: number,
+    height?: number
+  }) {
     const {
-      width = this.props.width,
-      height = this.props.height
+      width = Number(this.props.width),
```
akre54: These will convert to `NaN` if width or height is not supplied. Looks fine in this context but something to be aware of if we do a `typeof width === 'number'` check later on

src/mapbox/mapbox.js
```diff
@@ -111,33 +157,47 @@ function checkPropTypes(props, component = 'component') {
...
     }
 
+    this.mapboxgl = props.mapboxgl;
+
     if (!Mapbox.initialized && console.debug) { // eslint-disable-line
       Mapbox.initialized = true;
 
       // Version detection using babel plugin
       /* global __VERSION__ */
+      // $FlowFixMe
```
akre54: If the error is that `__VERSION__` is undefined, just `declare var __VERSION__` somewhere (either in this file or a header file somewhere)

Pessimistress: Then lint is not happy (__VERSION__ defined but not used) because __VERSION__ is replaced with the actual version string by the babel plugin version-inline.

flow-typed/npm/mapbox-gl_vx.x.x.js
Pessimistress: Doesn't look like it https://github.com/mapbox/mapbox-gl-js/tree/master/build

src/components/static-map.js
Pessimistress: The callback will only be missing width/height if they are disabled on AutoSizer: https://github.com/uber/react-map-gl/blob/master/src/components/static-map.js#L306

Overall I struggled a lot with flow when using type overrides. e.g. StaticMap and InteractiveMap support `width: number | string` but Mapbox and MapController only accept `width: number`. Flow does not like it when I pass props from one to another using `Object.assign`.

flow-typed/npm/mapbox-gl_vx.x.x.js
akre54: https://github.com/mapbox/mapbox-gl-js/pull/7079 looks like it'll be coming soon!

src/mapbox/mapbox.js
akre54: Weird. Oh well!

src/components/interactive-map.js
schnerd: FWIW, last time I looked into auto-generating proptypes from flowtypes it had a lot of rough edges, see this: https://github.com/uber-web/baseui/issues/47

Maybe it has improved since then. Safest bet is just to manually define propTypes, and surface area of react-map-gl's public API is relatively small.

PR #656, opened 2018-11-07
export InteractiveContext at the top level

vbud:
I just tried react-map-gl v4 - thanks for all the great work!

One thing that would be helpful: it is useful to get access to the `eventManager` on `context` like we were able to prior to react-map-gl v4. Exporting this at the top level is super helpful so we don't have to import from deep in the directory structure.

If you'd prefer a different name for the export let me know.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -45,3 +45,4 @@ export {default as MapController} from './utils/map-controller';
 
 // Experimental Features (May change in minor version bumps, use at your own risk)
 export {StaticContext as _MapContext} from './components/static-map';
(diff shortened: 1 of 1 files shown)


Pessimistress:
The experimental `_MapContext` export is intended for using Marker, Popup etc. without a map. If you wish to implement your own control components, can you give https://uber.github.io/react-map-gl/#/Documentation/advanced/custom-components a try?

vbud:
I see, you want to provide access to that context via `BaseControl`. I'll try that out.

PR #663, opened 2018-11-23
Cache the interactive context to prevent context thrashing

TeaSeaLancs:
Defining a new `interactiveContext` every render causes React to propagate that value across the context provider/consumer every render, which causes a large amount of unnecessary overhead.

This PR caches the interactiveContext and updates it when variables that affect it are changed in state.


TeaSeaLancs:
Hi Uber team, after receiving the CLA signing requirement and discussing it with my company, i'm going to close this pull request and submit it again from my own personal repo fork as opposed to a company-owned repo fork. Apologies for the inconvenience!

Issue #669, opened 2018-12-04
Handling Events in Custom MapController


tomrussell:
I'm trying to work with the MapController component to construct a DrawControl as described in https://github.com/uber/react-map-gl/issues/450#issuecomment-378590697, but using the updated API in v4

My simple controller looks like this:

```js
class DrawController extends MapController {
  constructor({ enabled }) {
    super();
    this.enabled = enabled;
    this.events = ['click'];
  }
  
  handleEvent = (event) => {
    if (this.enabled) {
      if (event.type === 'click') {
        console.log('Clicked');
      }
    }
    return super.handleEvent(event);
  }
}
```

HandleEvent is called but the final line returns an error when returning to the parent handleEvent function

```
Uncaught TypeError: Cannot read property 'call' of undefined
    at DrawController._this.handleEvent (DrawController.js:16)
    at EventRegistrar._emit (event-registrar.js:115)
    at EventRegistrar.handleEvent (event-registrar.js:87)
    at Manager.emit (hammer.js:2525)
    at TapRecognizer.emit (hammer.js:2130)
    at TapRecognizer.tryEmit (hammer.js:1572)
    at TapRecognizer.<anonymous> (hammer.js:2109)
    at boundFn (hammer.js:197)
```

Pessimistress:
Can you make sure that `MapController` is imported directly from `react-map-gl`? https://uber.github.io/react-map-gl/#/Documentation/introduction/upgrade-guide

guisers:
I am having the same problem.

Seems that having an arrow function inside of a parent class does not allow children to use `super` when implementing the parent arrow function as an arrow function. On the other hand, implementing the parent arrow function as a function results in the child function not being called at all.

I think `handleEvent` must be changed to a function.

tomrussell:
@guisers I had tried implementing `handleEvent` in my controller as a function instead, in doing so it was never called

guisers:
Yes, what I mean is that the parent class MapController in react-map-gl needs this change. There doesn't appear to be anything we can do 😬

Pessimistress:
Fix is published in 4.0.5

PR #674, opened 2018-12-05
Fix map controller class extensibility

Pessimistress:
For #669 

- Do not use arrow function for `handleEvent`
- Add example for testing

### DIFF
--- a/src/utils/map-controller.js
+++ b/src/utils/map-controller.js
@@ -73,7 +73,10 @@ export default class MapController {
   };
   _events: any = {};
   _transitionManager: TransitionManager = new TransitionManager();
-  _handleEvent: Function;
+
+  constructor() {
+    (this:any).handleEvent = this.handleEvent.bind(this);
+  }
 
   /**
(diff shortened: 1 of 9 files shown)


src/utils/map-controller.js
```diff
@@ -204,14 +205,18 @@ export default class MapController {
   }
 
   toggleEvents(eventNames: Array<string>, enabled: boolean) {
+    // Bind the method here instead of in the constructor, so that inherited classes
+    // do not have to do binding themselves
```
jcready: Not sure why you need to do this here. If you [bind in the constructor as was the case before](https://github.com/uber/react-map-gl/pull/651/files#diff-37b4265bef26974d812b2b2c97039331L59) then any class that extends this one should call `super()` it its constructor which should trigger the binding.

Pessimistress: Hmm I seem to have confused this with a legacy autobinder issue we had. Will revert to the good old way of binding.

PR #693, opened 2019-01-18
Fix dragging marker with offset

### LINKED ISSUE #691: Marker is jumping at start of drag and only works with offset
I'm having trouble with dragging with an offset. At the start of drag the marker jumps up and to the left. Oddly enough, when I remove the offsets now, the marker isn't dragging at all.
```javascript
<ReactMapGL
  ref={map => (this.mapRef = map)}
  {...viewport}
  mapboxApiAccessToken={MAPBOX_ACCESS_TOKEN}
>
  <Marker
    latitude={marker.latitude}
    longitude={marker.longitude}
    offsetLeft={-10}
    offsetTop={-25}
    draggable
    onDragEnd={this.onMarkerDragEnd}
  >
    <img
      width="20px"
      height="25px"
      src="https://image.flaticon.com/icons/svg/33/33622.svg"
    />
  </Marker>
</ReactMapGL>
```

why

If it matters, this is embedded in an electron app, but that really shouldn't from my knowledge of electron.

Thanks!

schnerd:
Fixes #691 

Think I overlooked this initially because the `<Pin>` component used in examples uses custom `translate` rules instead of the Marker offset props. I've updated the example and fixed the root issue.

### DIFF
--- a/examples/draggable-markers/src/control-panel.js
+++ b/examples/draggable-markers/src/control-panel.js
@@ -13,7 +13,7 @@ export default class ControlPanel extends PureComponent {
     const {events = {}} = this.props;
     const lngLat = events[eventName];
     return (
-      <div>
+      <div key={eventName}>
         <strong>{eventName}:</strong>{' '}
         {lngLat ? lngLat.map(round5).join(', ') : <em>null</em>}
       </div>
@@ -26,9 +26,9 @@ export default class ControlPanel extends PureComponent {
       <Container>
         <h3>Draggable Marker</h3>
         <p>Try dragging the marker to another location.</p>
-        <p>
+        <div>
           {eventNames.map(this.renderEvent)}
-        </p>
+        </div>
         {/* TODO add a "View Code" link here when we know the release */}
       </Container>
     );
--- a/src/components/draggable-control.js
+++ b/src/components/draggable-control.js
@@ -20,16 +20,25 @@
 import PropTypes from 'prop-types';
 import BaseControl from './base-control';
 
+import type {MjolnirEvent} from 'mjolnir.js';
+
 const propTypes = Object.assign({}, BaseControl.propTypes, {
   draggable: PropTypes.bool,
   onDrag: PropTypes.func,
   onDragEnd: PropTypes.func,
-  onDragStart: PropTypes.func
+  onDragStart: PropTypes.func,
+  // Offset from the left
+  offsetLeft: PropTypes.number,
+  // Offset from the top
+  offsetTop: PropTypes.number
 });
 
 const defaultProps = Object.assign({}, BaseControl.defaultProps, {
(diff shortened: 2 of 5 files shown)


examples/draggable-markers/src/control-panel.js
```diff
@@ -26,9 +26,9 @@ export default class ControlPanel extends PureComponent {
       <Container>
         <h3>Draggable Marker</h3>
         <p>Try dragging the marker to another location.</p>
-        <p>
+        <div>
           {eventNames.map(this.renderEvent)}
-        </p>
+        </div>
```
schnerd: Fixes a couple react warnings in the example

src/components/draggable-control.js
```diff
@@ -20,17 +20,26 @@
...
 const propTypes = Object.assign({}, BaseControl.propTypes, {
   draggable: PropTypes.bool,
   onDrag: PropTypes.func,
   onDragEnd: PropTypes.func,
-  onDragStart: PropTypes.func
+  onDragStart: PropTypes.func,
+  // Offset from the left
+  offsetLeft: PropTypes.number,
+  // Offset from the top
+  offsetTop: PropTypes.number
```
schnerd: Moving the `offset*` props back into draggable control felt better than overriding _getDragLngLat in Marker. If there are other classes extending DraggableControl in the future, they may also want to support offsets.

src/components/marker.js
```diff
@@ -58,18 +49,33 @@ export default class Marker extends DraggableControl {
...
+
+    const [x, y] = this._getPosition();
 
     const containerStyle = {
       position: 'absolute',
-      left: x + offsetLeft,
-      top: y + offsetTop
+      left: x,
+      top: y,
+      cursor: draggable ? (dragPos ? 'grabbing' : 'grab') : 'auto'
```
schnerd: @Pessimistress This isn't perfect because sometimes the mouse moves faster than the marker, but it seems to work reasonably well. Let me know if you there's a better way of achieving this.

schnerd:
Not sure why `/travis-ci/push` failed... tests seem unrelated to my changes.

src/components/marker.js
Pessimistress: You can set the cursor style on the entire canvas to fix this. Without doing it the "React way" you can access the root element through the pointer events...

schnerd: Gotcha. Sounds like something that might require its own PR, and the small change here should be a decent improvement for now.

timendez:
@schnerd just FYI the example on the page still allows you to move the pin if you click and hold to the bottom left of the pin:

gif

Pessimistress:
@schnerd https://github.com/uber/react-map-gl/blob/master/docs/developer-guide.md#website

Issue #700, opened 2019-01-24
Cannot pan map after using zoom navigation control


macobo:
Writing this to save others a couple of hours.

We use react-map-gl with redux, storing viewstate in it. We ran into an issue with a pretty vanilla setup where the map stopped allowing panning after clicking on zoom navigation control.

Our component render:

```javascript
    <ReactMapGL
      onViewportChange={this.props.updateViewport}
      {...this.props.viewport}
    >
      <NavigationControl onViewportChange={this.props.updateViewport} />
```

We discovered this was because the payload passed to `this.props.updateViewport` contained a noop `onViewportChange` function.

![screenshot from 2019-01-24 18-17-43](https://user-images.githubusercontent.com/148820/51691896-62bca900-2004-11e9-990a-ab92a35fdcff.png)

This promptly got stored in our store. Due to `{...this.props.viewport}` being last this overwrote our own onViewportChange function - causing it to do nothing.

An easy fix was to move the viewport spread as the first argument:

```javascript
    <ReactMapGL
      {...this.props.viewport}
      onViewportChange={this.props.updateViewport}
    >
      <NavigationControl onViewportChange={this.props.updateViewport} />
```

Honestly no clue why this doesn't break with react setState.

macobo:
Perfect, thank you! Feel free to close after the next release. :+1:

Pessimistress:
Fix published in 4.0.10.

PR #702, opened 2019-01-25
Remove excessive fields from NavigationControl's callback argument

Pessimistress:
For https://github.com/uber/react-map-gl/issues/700

### DIFF
--- a/src/utils/map-controller.js
+++ b/src/utils/map-controller.js
@@ -28,9 +28,9 @@ import type {MjolnirEvent} from 'mjolnir.js';
 const NO_TRANSITION_PROPS = {
   transitionDuration: 0
 };
-const LINEAR_TRANSITION_PROPS = {
+export const LINEAR_TRANSITION_PROPS = {
   transitionDuration: 300,
-  transitionEasing: t => t,
+  transitionEasing: (t: number) => t,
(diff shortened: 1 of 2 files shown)


src/utils/map-controller.js
```diff
@@ -28,9 +28,9 @@ import type {MjolnirEvent} from 'mjolnir.js';
 const NO_TRANSITION_PROPS = {
   transitionDuration: 0
 };
-const LINEAR_TRANSITION_PROPS = {
+export const LINEAR_TRANSITION_PROPS = {
   transitionDuration: 300,
-  transitionEasing: t => t,
+  transitionEasing: (t: number) => t,
   transitionInterpolator: new LinearInterpolator(),
   transitionInterruption: TRANSITION_EVENTS.BREAK
```
rolyatmax: Will the `TransitionManager`'s default props need to match these, too? If so, is it worth importing these into `utils/transition-manager`?

Pessimistress: Technically no. The default props of `TransitionManager` is the default behavior when user only supply `transitionDuration`, which can be anything we want. The controller's `LINEAR_TRANSITION_PROPS` intend to match the behavior of Mapbox.

Issue #699, opened 2019-01-23
Customized map control will lose when set `reuseMaps` to true in StaticMap


hijiangtao:
## Problem description

I found `reuseMaps` will keep all the things the same but except the added controls. Once, I add a map control when map is loaded (such as a language control `A`), then I switch to another page, which also use `react-map-gl`, I found map in the new page will go back to its original control without `A`.

## Reproduce steps

1. Include `component.js` in two different pages, such as `/pageA` and `/pageB`;

```
// component.js
import React from 'react';
import { StaticMap } from 'react-map-gl';

const addControlHandler = (event) => {
  event.target.setLayoutProperty('country-label-lg', 'text-field', ['get', 'name_zh']);
}

const Map = (props) => {
  return (
    <StaticMap
          onLoad={addControlHandler}
          reuseMaps
          mapStyle="mapbox://styles/mapbox/dark-v9"
          preventStyleDiffing
          mapboxApiAccessToken={MAPBOX_TOKEN}
        />
  );
}

export default Map;

// pageA.js
import Map from './component.js';
const App = (props) => {
	return (
		<Map />
	)
}
```

2. Add a link to pageB in pageA;
3. Open pageA, and click the link;
4. You will see the pageB without the added control in pageA;

If I use `reuseMaps`, the `onLoad` will never be fired again in the new page, so I can't add control manually, too.

hijiangtao:
Why not pass `Mapbox.savedMap` into the target property of event, since user may need to manually operate something if map control will get lost at this situation?

I have a judgement like this in my project, would it save me from the unexpected error?

```
// code in this issue
const addControlHandler = (event) => {
  event.target.setLayoutProperty('country-label-lg', 'text-field', ['get', 'name_zh']);
}

// actual code in my project
const addControlHandler = (event) => {
  const map = event && event.target;
  if (map) {
	map.setLayoutProperty('country-label-lg', 'text-field', ['get', 'name_zh']);
  }
}
```

hijiangtao:
And I thought it means currently I can only set `reuseMaps` to false to forbid such *no-update-triggered-at-second-time* situation?

Pessimistress:
> Why not pass Mapbox.savedMap into the target property of event

No reason, already flagged as a bug. I will try to put up a fix this weekend. If this is blocking you you're welcome to open a PR.

Pessimistress:
[...]

Issue #683, opened 2018-12-27
Is there an API I can use for calling setLayoutProperty method on the map instance?


hijiangtao:
The mapbox API provides `setLayoutProperty` method for developers to change the layout property of a map instance, such as the `language`:

According to mapbox [API](https://www.mapbox.com/mapbox-gl-js/example/language-switch/), I can use it in this way:

```
var map = new mapboxgl.Map({
    container: 'map',
    style: 'mapbox://styles/mapbox/light-v9',
    center: [16.05, 48],
    zoom: 2.9
});

// and the new property value.
map.setLayoutProperty('country-label-lg', 'text-field', ['get', 'name_zh']);
```

I am developing project with react-map-gl and deck.gl, I use `StaticMap` exactly in my project and need to change the display language of maps.

But after looking through the docs of react-map-gl and `StaticMap` [file](https://github.com/uber/react-map-gl/blob/cb173b0ac405c42200ddfdaad5a6bed96d665fbf/src/components/static-map.js), I couldn't find a method to achieve that. 

I found that `StaticMap` provides a method called `getMap()`:

https://github.com/uber/react-map-gl/blob/cb173b0ac405c42200ddfdaad5a6bed96d665fbf/src/components/static-map.js#L174-L176

But when I use this method to make changes to map instance, there's nothing happened.

```
const map = this.refMap.current.getMap();

map.on(
	'load', 
	() => map.setLayoutProperty('country-label-lg', 'text-field', ['get', 'name_zh-Hans'])
);
...

render() {
	return (
		<StaticMap ref={e => {this.refMap = e;}} >
	);
}
```

hijiangtao:
I tried again and found the load event listener callback function had never been called, is it right to use `getMap()` method to get the map instance?

Mapbox GL JS API give a way to listen load event like this:

```
var map = new mapboxgl.Map({
    container: 'map',
    style: 'mapbox://styles/mapbox/light-v9',
    center: [-90.96, -0.47],
    zoom: 8
});

map.on('load', function () {});
```

Pessimistress:
- Yes, [getMap](https://uber.github.io/react-map-gl/#/Documentation/api-reference/static-map?section=methods) is a public API. Though your code seems to be confusing the legacy [callback ref](https://reactjs.org/docs/refs-and-the-dom.html#callback-refs) and [createRef](https://reactjs.org/docs/refs-and-the-dom.html#creating-refs) from React.
[...]

PR #704, opened 2019-01-25
Bug fix: implement onLoad event calling when reuseMaps is set to true

hijiangtao:
## Description and resolution

According to the discussion in #699 , there's a TODO inside mapbox.js, which is calling onLoad event again when map instance is re-created again with its' property `reuseMaps` set to true.

However, we need to handle two situations here:

1. We need to `setStyle` with customized `mapStyle`: In this situation, we need to wait until style is loaded, and then call `onLoad` event handler; we choose `.once` API here to listen `style.load` event;
2. We don't need to `setStyle` with customized `mapStyle`: In this situation, we can directly generate an `event` object and pass it to `props.onLoad` method and call it right now;

## Tests and documentation

* Test: `npm run test` passed with 222 tests.
* Documentation: No need to update document since there's not API change with this commit;

### DIFF
--- a/examples/main/constants/toc.js
+++ b/examples/main/constants/toc.js
@@ -77,15 +77,14 @@ export const standaloneExamples = [
     component: ZoomToBounds
   },
   {
-<<<<<<< HEAD
     path: 'reuse-map',
     name: 'ReuseMaps',
     component: ReuseMap
-=======
+  },
+  {
     path: 'heatmap',
     name: 'Heatmap',
     component: Heatmap
->>>>>>> upstream/master
   }
 ];
 
--- a/examples/main/constants/toc.js
+++ b/examples/main/constants/toc.js
@@ -16,6 +16,7 @@ import Interaction from '../../interaction/src/app';
 import Layers from '../../layers/src/app';
 import ViewportAnimation from '../../viewport-animation/src/app';
 import ZoomToBounds from '../../zoom-to-bounds/src/app';
+import ReuseMap from '../../reuse-map/src/app';
 
 export const BASIC_EXAMPLES = 'basicExamples';
 export const STANDALONE_EXAMPLES = 'standalonExamples';
@@ -73,6 +74,11 @@ export const standaloneExamples = [
     path: 'zoom-to-bounds',
     name: 'Zoom To Bounds',
     component: ZoomToBounds
+  },
+  {
+    path: 'reuse-map',
+    name: 'ReuseMaps',
+    component: ReuseMap
   }
 ];
 
--- a/examples/main/constants/toc.js
+++ b/examples/main/constants/toc.js
@@ -16,7 +16,6 @@ import Interaction from '../../interaction/src/app';
 import Layers from '../../layers/src/app';
 import ViewportAnimation from '../../viewport-animation/src/app';
 import ZoomToBounds from '../../zoom-to-bounds/src/app';
-import ReuseMap from '../../reuse-map/src/app';
 import Heatmap from '../../heatmap/src/app';
 
 export const BASIC_EXAMPLES = 'basicExamples';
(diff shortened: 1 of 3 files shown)


src/mapbox/mapbox.js
```diff
@@ -236,13 +236,18 @@ export default class Mapbox {
...
-      // Update style
+      // Update style and call onload again
       if (props.mapStyle) {
         this._map.setStyle(props.mapStyle);
-      }
 
-      // TODO - need to call onload again, need to track with Promise?
-      props.onLoad();
+        // call onload event handler after style fully loaded when style needs update
+        this._map.once('style.load', props.onLoad);
```
Pessimistress: Do we know for sure that this event will fire even if `props.mapStyle` is the same as the last style set?

hijiangtao:
I also unified inputs of `props.onLoad` event to one object (with `type` property to be `load`), though in some situation, the event type maybe `styledata` rather than `load`.

That's because developers define `onLoadHandler` method inside `onLoad` property of react-ma-p-gl, so they would always want `event` object that passed to their `onLoadHandler` should be the same, which should always point to a `load` event.

It may not be a perfect solution but I think it's the best one right now, a better solution is to have a more precise event types (pretty the same as https://github.com/uber/react-map-gl/issues/683#issuecomment-457804805 ), then we can handler `load` event and `styledata` event separately.

A pseudocode should like this inside `_create` method (suppose there's a handler called styleLoad in props):

```
      ...

	  // Step3: update style and call onload again
      if (props.mapStyle) {
        this._map.setStyle(props.mapStyle, {
          diff: true
        });

        // call onload event handler after style fully loaded when style needs update
        (this._map.isStyleLoaded() ?
          props.styleLoad() : this._map.once('styledata', props.styleLoad))();
      }

	  props.onLoad({
        type: 'load',
        target: this._map
      });

```

src/mapbox/mapbox.js
```diff
@@ -236,13 +236,23 @@ export default class Mapbox {
...
+      });
+
       if (props.mapStyle) {
-        this._map.setStyle(props.mapStyle);
+        this._map.setStyle(props.mapStyle, {
+          diff: true
+        });
+
+        // call onload event handler after style fully loaded when style needs update
+        (this._map.isStyleLoaded() ?
```
Pessimistress: This syntax is rather odd. Just use `if`?

hijiangtao: I got a warning of **Method '_create' has too many statements (26). Maximum allowed is 25.** if I use if statement, that’s why I choose this one syntax.

Pessimistress: You can add this line to the top of this method to disable the lint rule:

```js
/* eslint-disable max-statements */
```

I will refactor it after the patch.

hijiangtao: Yes, that's correct since it will overwrite the global settings. I can handle it by adding another commit.

hijiangtao:
4 more implements with:

1. Enrich `examples/reuse-map` with `onLoad` event;
2. Implement TOC since reusemap example had no entrance currently;
3. Use `if` statement rather than a ternary expression;
4. Resolve conflicts from upstream;

examples/main/constants/toc.js
```diff
@@ -76,9 +77,15 @@ export const standaloneExamples = [
...
   {
+<<<<<<< HEAD
+    path: 'reuse-map',
+    name: 'ReuseMaps',
+    component: ReuseMap
+=======
     path: 'heatmap',
     name: 'Heatmap',
     component: Heatmap
+>>>>>>> upstream/master
```
hijiangtao: Sorry for this, I'll resolve it by adding another commit.

Pessimistress:
Reusemap example is not on the website because it does not meet the visual quality bar. Can you remove it for now so that we can release?

Pessimistress:
> is there any specific guidelines of the visual quality bar (I'm interested)?

Since this is a wrapper library, we try to put up examples that replicate existing [mapbox-gl-js examples](https://docs.mapbox.com/mapbox-gl-js/examples) or at least along the same lines. The reuse-map example in particular was made for performance profiling/debugging purposes so we never bothered to make it look good.

Issue #707, opened 2019-01-27
Race condition between styles loading and InteractiveMap.interactiveLayerIds


macobo:
We've now seen this error in production a few times. 

![screenshot from 2019-01-27 11-00-05](https://user-images.githubusercontent.com/148820/51798941-3763d500-2223-11e9-8f08-eb870af8c542.png)

This happens if:

1. You have `interactiveLayerIds` set
2. You move the cursor on the map before mapbox is done loading styles

I currently don't have a jsfiddle prepared, but can prepare it if needed.

No visible breakage afterwards, the application runs as intended, but the uncaught error is clogging up our error handling tools.

### Proposed fix

Since the problem comes from the library querying the map before the styles are loaded, a possible fix is to ignore all mouse events until `onLoad` has fired. 

Happy to provide a fix, though I'm struggling a bit with writing good tests for this in your setup (used to jest + enzyme).

Pessimistress:
Fix published in 4.0.10.

rgazelot:
Hello guys, actually the issue is still there according to what I experienced. 
![Screenshot 2021-01-19 at 12 20 57](https://user-images.githubusercontent.com/1488251/105028100-e3c2a580-5a50-11eb-94c8-29aeabe0cc7e.png)

Exact same behaviour, you move your mouse over the map before the map style is rendered. I'm using the v5.2.12, but maybe I missed something?

AntonRaichuk:
Seeing the same issue as @rgazelot 


@Pessimistress could you take a look?

koooge:
I got the same error message with 7.0.19.

workaround:

```js
const [interactiveLayerIds, setInteractiveLayerIds] = useState<string[]>([])

<ReactMapboxGl
  interactiveLayerIds={interactiveLayerIds}
  onLoad={() => setInteractiveLayerIds(['someLayerId'])}
  ...
>
```

PR #708, opened 2019-01-27
Don't query interactive layer ids before map has loaded

macobo:
This can cause an false-positive error such as `The layer 'foobar' does
not exist in the maps style and cannot be queries for features`.

This needs tests, though the author of this PR doesn't know how to
create these without enzyme.

Solves #707

### DIFF
--- a/src/components/interactive-map.js
+++ b/src/components/interactive-map.js
@@ -166,6 +166,7 @@ type InteractiveMapProps = StaticMapProps & {
 };
 
 type State = {
+  isLoaded: boolean,
   isDragging: boolean,
   isHovering: boolean
 };
@@ -204,6 +205,8 @@ export default class InteractiveMap extends PureComponent<InteractiveMapProps, S
   }
 
   state : State = {
+    // Whether mapbox styles have finished loading
+    isLoaded: false,
     // Whether the cursor is down
     isDragging: false,
     // Whether the cursor is over a clickable feature
@@ -347,6 +350,11 @@ export default class InteractiveMap extends PureComponent<InteractiveMapProps, S
     return event;
   }
 
+  _onLoad = (event : MapEvent) => {
+    this.setState({isLoaded: true});
+    this.props.onLoad(event);
+  }
+
   _onEvent = (callbackName : string, event : MapEvent) => {
     const func = this.props[callbackName];
     if (func) {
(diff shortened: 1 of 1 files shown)


src/components/interactive-map.js
```diff
@@ -347,6 +350,11 @@ export default class InteractiveMap extends PureComponent<InteractiveMapProps, S
     return event;
   }
 
+  _onLoad = () => {
+    this.setState({stylesLoaded: true});
+    this.props.onLoad();
```
Pessimistress: Need to forward the event object to the callback.

src/components/interactive-map.js
```diff
@@ -204,6 +205,8 @@ export default class InteractiveMap extends PureComponent<InteractiveMapProps, S
   }
 
   state : State = {
+    // Whether mapbox styles have finished loading
+    stylesLoaded: false,
```
Pessimistress: Just call it `isLoaded`?

src/components/interactive-map.js
macobo: I couldn't find documentation what the event object here is, but you're right that there is an event being triggered here.

Interestingly enough, the [typescript type definitions here](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/react-map-gl/index.d.ts#L43) also aren't aware there's an event object.

Will fix though, thanks for catching this!

macobo:
Thanks for the review! I think this is ready for another round @Pessimistress

src/components/interactive-map.js
```diff
@@ -390,7 +398,7 @@ export default class InteractiveMap extends PureComponent<InteractiveMapProps, S
     if (!this.state.isDragging) {
       const {onHover, interactiveLayerIds} = this.props;
       let features;
-      if (interactiveLayerIds || onHover) {
+      if (this.state.isLoaded && (interactiveLayerIds || onHover)) {
```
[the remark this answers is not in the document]
macobo: Thanks. Opted to always normalize the event.

Pessimistress:
Thank you for contributing!

(Looks like CI is failing for unrelated reasons)

macobo:
Can confirm, `npm test` runs without fail on local machine.

Thanks for the quick reviews @Pessimistress. Feel free to close any relevant issues now.

This library is such pleasure to use. Thanks for building and maintaining it! Now if only we could figure out proper drawing controls at my workplace we could get rid of google maps for good...

Issue #673, opened 2018-12-05
InteractiveContext


tilgovi:
I have an interest in using either `InteractiveContext` or `DraggableControl` (or both). I can import them with a deep import from the source right now, but I want to have clear expectations.

Is the intention that these are private right now or was it an oversight to not export them?

Pessimistress:
We can potentially export `DraggableControl`.

What is your use case for `InteractiveContext`? If you need access to `eventManager`, I'd recommend extending the [BaseControl class](https://uber.github.io/react-map-gl/#/Documentation/advanced/custom-components).

The `_MapContext` (aka `StaticContext`) export is merely targeting the use case where we need to use Marker/Popup/NavigationControl without a map.

tilgovi:
`InteractiveContext` is probably not necessary. I'm currently accessing `eventManager`, but that may be because I'm not yet using `BaseControl` for everything.

Using `DraggableControl` is attractive because it handles dragging, but perhaps `Marker` is the right thing to use for that.

tilgovi:
It's not clear to me how to bind events from `BaseControl` other than to access `this._context.eventmanager`, which feels like private API. If instead `InteractiveContext` were public, I could use that directly.

I'm particularly interested in experimenting with `InteractiveContext` and the new [React Hooks API](https://reactjs.org/docs/hooks-reference.html). I could see passing`InteractiveContext` to [`useContext`](https://reactjs.org/docs/hooks-reference.html#usecontext) in order to bind events.

I like that react-map-gl has a good default `MapController` and I like mjolnir.js, but it is limiting to centralize all event handling in `InteractiveMap` props and to be forced to use class-based components in order to extend `BaseControl`.

tilgovi:
I've modified the issue to remove `DraggableControl`. The `InteractiveContext` would be the most useful export.

Pessimistress:
@tilgovi Your request is valid. It's just that from API design perspective I really don't want to export two contexts - let me think about how they can be merged.

tilgovi:
Thank you!

PR #718, opened 2019-02-05
Consolidate InteractiveContext and StaticContext into one

Pessimistress:
For https://github.com/uber/react-map-gl/issues/673

The experimental export `MapContext` now allows consumers to receive all context props.

### DIFF
--- a/src/components/map-context.js
+++ b/src/components/map-context.js
@@ -1,0 +1,19 @@
+import {createContext} from 'react';
+
+export default createContext({
+  /* Map context */
+
+  // Viewport
+  viewport: null,
+  // mapboxgl.Map instance
+  map: null,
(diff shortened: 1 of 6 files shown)


src/components/map-context.js
```diff
@@ -0,0 +1,17 @@
+import {createContext} from 'react';
+
+export default createContext({
```
xintongxia: I am curious why these but not others needed to be in `MapContext`?

xintongxia:
Add documentation.

src/components/map-context.js
Pessimistress: Examples for "others"? These are information our existing controls need to render properly. We do add new props to the context if they are identified as useful.

Issue #711, opened 2019-01-31
Update to docs: Display Maps Without A Mapbox Token


thomasgwatson:
Recently went through the process of adding a non-Mapbox base layer to a ReactMapGL instance.

At a minimum, I wish the docs had directed me to https://docs.mapbox.com/mapbox-gl-js/style-spec/ instead of https://docs.mapbox.com/mapbox-gl-js/style-spec/#sources-vector 

It is a small difference in url (its the same page!) but the data structure of the overall spec is so much more important to get that one of its subparts (just the source)

In addition, there were a few other pointers that would be helpful

-don't forget to convert to immutable data structure
-Your style depends on what schema your vector tile is in
-other authorization patterns (reference the transformRequest API)

I want to submit a pull-request but before I do, just filing this in to have an issue to reference and would love to get a quick read on which (if any) of the extra pointers I should include. Thanks!

macobo:
We've done the same at eAgronom. Note that immutable.js is not needed at all, you can as well pass normal objects as mapStyle

thomasgwatson:
@macobo I haven't tested but was going off the comments in: https://github.com/uber/react-map-gl/issues/119

"mapStyle is expected to be an ImmutableJS object or a url string. Wrapping your style first in an ImmutableJS object with Immutable.fromJS(mapStyle) should resolve the issue." Has that changed?

Also... @xintongxia added the docs tag but... not sure if there is interest in a pull request about this?

Pessimistress:
Please do!

Re: immutable - immutable.js dependency was removed in v4.0. You can still use it to avoid the awkward spreading operation, but it's no longer required.

PR #722, opened 2019-02-08
Updated docs for using non-mapbox sourced tiles

thomasgwatson:
As described in #711 fleshed out the docs for the use of non-mapbox tile sources.


thomasgwatson:
Ahh, I signed the CLA but I'm concerned I won't be able to link it to my commit, since the machine I pushed on didn't have my email set in its git config :\

Pessimistress:
@thomasgwatson Can you try push again with the correct email?

thomasgwatson:
@Pessimistress apologies, had to pop another PR #730 so I wouldn't have to overwrite commit history! Should work now

Issue #721, opened 2019-02-06
Calls setState after the component is unmounted


jaydenseric:
`<ReactMapGL />` internally calls `setState` after the component is unmounted, triggering a React warning like this:

```
Warning: Can't perform a React state update on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.
    in InteractiveMap
```

You can reproduce this issue by putting a basic map like this on a page:

```jsx
import ReactMapGL from 'react-map-gl'

<ReactMapGL
  width="100px"
  height="100px"
  mapboxApiAccessToken="…"
/>
```

Then from another page, quickly navigate to the page and back again within a few seconds.

Pessimistress:
Fixed in 4.0.11

PR #723, opened 2019-02-08
deregister map event listeners when destroy map

xintongxia:
for #721

### DIFF
--- a/src/mapbox/mapbox.js
+++ b/src/mapbox/mapbox.js
@@ -224,6 +224,13 @@ export default class Mapbox {
   }
 
   // PRIVATE API
+  _fireLoadEvent = () => {
+    this.props.onLoad({
+      type: 'load',
+      target: this._map
+    });
+  };
+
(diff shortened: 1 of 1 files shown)


src/mapbox/mapbox.js
```diff
@@ -224,6 +226,18 @@ export default class Mapbox {
   }
 
   // PRIVATE API
+  // delayed fireLoadEvent in case map is unmounted right after
+  _fireLoadEvent = () => {
+    this._timer = window.setTimeout(() => {
```
xintongxia: I do not think we can avoid that error completely, as this `mapbox` will call the `props.onLoad`, and have no control what the user will do in `props.onLoad`.

Also when `props.onLoad` is already triggered in `_fireLoadEvent`, I do not know what is the appropriate way to cancel this `props.onLoad` call. So I add a timer here to delay the call in case the map is unmounted before loaded.

Issue #672, opened 2018-12-05
Mapbox' GeolocateControl does not trigger onViewportChange


felixerdy:
I implemented Mapbox' GeolocateControl like this: 
```js
componentDidMount() {
    this.map.getMap().addControl(new mapboxgl.GeolocateControl({
      positionOptions: {
        enableHighAccuracy: true,
      },
      trackUserLocation: true,
    }))
}
```

where `this.map` is the reference to `ReactMapGL`. 
```js
<ReactMapGL
          ref={el => (this.map = el)}
          {...this.props.viewport}
          onViewportChange={viewport => this.props.setViewport(viewport)}
...
```

It all works fine and I can visualize the users location. The transition to the users location however does not trigger `onViewportChange`. Panning / zooming by mouse works and triggers `onViewportChange`of course.

Pessimistress:
Unfortunately, Mapbox's native controls do not trigger any React callbacks. We can potentially add a new React wrapper for this control, hopefully when I have time to work on this.

Relevant ticket: https://github.com/uber/react-map-gl/issues/484

Pessimistress:
An React version of GeolocateControl is included in v4.1.

Issue #484, opened 2018-03-21
Track User Location


webdevsyd:
Hi, is there a way to location the current coordinates of the user? Im referring to this example
https://www.mapbox.com/mapbox-gl-js/example/locate-user/

Thanks!

Pessimistress:
@webdevsyd You can't use Mapbox's geolocate control directly, but it is not hard to make your own:

```js
_locateUser() {
  // https://developer.mozilla.org/en-US/docs/Web/API/Geolocation/Using_geolocation
  navigator.geolocation.getCurrentPosition(position => {
    this.updateViewport({
      longitude: position.coords.longitude,
      latitude: position.coords.latitude
    });
  });
}
```

xintongxia:
`GeolocateControl` is included since version 4.1.3

PR #724, opened 2019-02-09
add geolocate control

xintongxia:
For #484 and #672

### DIFF
--- a/docs/components/geolocate-control.md
+++ b/docs/components/geolocate-control.md
@@ -1,0 +1,75 @@
+# Geolocate Control
+
+This is a React equivalent of Mapbox's [GeolocateControl](https://www.mapbox.com/mapbox-gl-js/api/#geolocatecontrol).
+
+```js
+import React, { Component } from "react";
+import ReactMapGL, {GeolocateControl} from "react-map-gl";
+
+class Map extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      viewport: {
+        width: 800,
+        height: 600,
+        longitude: -122.45,
+        latitude: 37.78,
+        zoom: 14
+      }
+    }
+  }
+  
+  _updateViewport = (viewport) => {
+    this.setState({viewport});
+  }
+  
+  render() {
+    const {viewport} = this.state;
+    return (
+      <ReactMapGL {...viewport} onViewportChange={updateViewport}>
+        <div style={{ position: "absolute", right: 0 }}>
+          <GeolocateControl 
+            positionOptions={{enableHighAccuracy: true}}
+            trackUserLocation={true}
+            onViewportChange={this._updateViewport}
+          />
+        </div>
+      </ReactMapGL>
+    );
+  }
+}
+```
+
+## Properties
+
+##### `onViewportChange` {Function}
+
+Callback when the viewport needs to be updated. See [InteractiveMap](/docs/components/interactive-map.md).
+
+
+Accepts all the options of [Mapbox GeolocatControl](https://docs.mapbox.com/mapbox-gl-js/api/#geolocatecontrol).
+
+##### `positionOptions` {Object} - default: `{enableHighAccuracy:false, timeout:6000}`
+
+A Geolocation API [PositionOptions](https://developer.mozilla.org/en-US/docs/Web/API/PositionOptions) object.
+
+##### `fitBoundsOptions` {Object} - default: `{maxZoom: 15}`
+
+A  [fitBounds](https://docs.mapbox.com/mapbox-gl-js/api/#map#fitbounds) options object to use when the map is panned and zoomed to the user's location. The default is to use a  maxZoom of 15 to limit how far the map will zoom in for very accurate locations.
+
+##### `trackUserLocation` {Boolean} - default: `false`
+
+If  true the Geolocate Control becomes a toggle button and when active the map will receive updates to the user's location as it changes.
+
+##### `showUserLocation` {Boolean} - default: `true`
+
+By default a dot will be shown on the map at the user's location. Set to false to disable.
+
+## Styling
+
+Like its Mapbox counterpart, this control relies on the mapbox-gl stylesheet to work properly. Make sure to add the stylesheet to your page.
+
+## Source
+
+[geolocate-control.js](https://github.com/uber/react-map-gl/tree/master/src/components/geolocate-control.js)
--- a/src/components/geolocate-control.js
+++ b/src/components/geolocate-control.js
@@ -185,8 +185,9 @@ export default class GeolocateControl extends BaseControl {
   };
 
   _renderMarker = () => {
+    const {showUserLocation} = this.props;
     const {markerPosition} = this.state;
-    if (!markerPosition) {
+    if (!showUserLocation || !markerPosition) {
       return null;
     }
 
@@ -207,7 +208,7 @@ export default class GeolocateControl extends BaseControl {
       return null;
     }
 
-    const {className, showUserLocation} = this.props;
+    const {className} = this.props;
     return createElement(
       'div',
       {
@@ -216,7 +217,7 @@ export default class GeolocateControl extends BaseControl {
         onContextMenu: e => e.preventDefault()
       },
       [
-        showUserLocation && this._renderMarker(),
+        this._renderMarker(),
         this._renderButton('geolocate', 'Geolocate', this._onClickGeolocate)
       ]
     );
--- a/src/components/geolocate-control.js
+++ b/src/components/geolocate-control.js
@@ -73,10 +73,18 @@ export default class GeolocateControl extends BaseControl {
     });
   }
 
+  componentDidUpdate() {
+    // MapboxGeolocateControl needs manipulate the style of Marker's container
+    const markerRef = this._markerRef.current;
+    if (this._mapboxGeolocateControl && markerRef) {
+      this._mapboxGeolocateControl._dotElement = markerRef._containerRef.current;
+    }
+  }
+
   componentWillUnmount() {
     // re-implement MapboxGeolocateControl's _onRemove
     // clear the geolocation watch if exists
-    const geolocationWatchID = this._mapboxGeolocateControl._geolocateButton;
+    const geolocationWatchID = this._mapboxGeolocateControl._geolocationWatchID;
     if (geolocationWatchID !== undefined) {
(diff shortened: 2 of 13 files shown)


docs/components/geolocate-control.md
```diff
@@ -0,0 +1,42 @@
...
+
+This is a React equivalent of Mapbox's [GeolocateControl](https://www.mapbox.com/mapbox-gl-js/api/#geolocatecontrol).
+
+```js
+import React, { Component } from "react";
+import ReactMapGL, {GeolocateControl} from "react-map-gl";
+
+class Map extends Component {
+  render() {
+    const { viewport, updateViewport } = this.props;
```
Pessimistress: Make this a functional example by storing viewport in component state

docs/components/geolocate-control.md
```diff
@@ -0,0 +1,42 @@
...
+        </div>
+      </ReactMapGL>
+    );
+  }
+}
+```
+
+## Properties
+
+Accepts all the options of [Mapbox GeolocatControl](https://docs.mapbox.com/mapbox-gl-js/api/#geolocatecontrol)
```
Pessimistress: List these props for clarity.
Mapbox may add new options to this control in future releases and we still need to update the component to support them.

src/components/geolocate-control.js
```diff
@@ -0,0 +1,285 @@
...
+  // Callbacks fired when the user interacted with the map. The object passed to the callbacks
+  // contains viewport properties such as `longitude`, `latitude`, `zoom` etc.
+  onViewStateChange: PropTypes.func,
+  onViewportChange: PropTypes.func
+});
+
+const defaultProps = Object.assign({}, BaseControl.defaultProps, {
+  className: '',
+  // mapbox geolocate options
+  positionOptions: {
```
Pessimistress: Is this something we have to copy, or can we use mapbox's default by not passing it in?

src/components/geolocate-control.js
```diff
@@ -0,0 +1,285 @@
...
+        if (this.props[prop]) {
+          controlOptions[prop] = this.props[prop];
+        }
+      }
+    );
+
+    this._mapboxGeolocateControl = new MapboxGeolocateControl(controlOptions);
+
+    // replace mapbox internal prop
+    this._mapboxGeolocateControl._map = this._context.map;
```
Pessimistress: Does it work if we don't let the control access the map? This risks it changing the map state without our knowledge.

src/components/geolocate-control.js
```diff
@@ -0,0 +1,285 @@
...
+    this._mapboxGeolocateControl._watchState = WATCH_STATE.OFF;
+
+    // replace mapbox internal UI elements
+    this._mapboxGeolocateControl._geolocateButton = this._geolocateButtonRef.current;
```
Pessimistress: We should render the UI components in React.

src/components/geolocate-control.js
```diff
@@ -0,0 +1,285 @@
...
+    this._mapboxGeolocateControl._updateCamera = this._updateCamera;
+
+    this._mapboxGeolocateControl._setup = true;
+  };
+
+  componentWillReceiveProps(nextProps) {
+    if (nextProps.trackUserLocation) {
+      this.setState({watchState: WATCH_STATE.OFF}, () => {
+        const {map} = this._context;
+        map.on('movestart', event => {
```
Pessimistress: This is too much hacking. How much of the native control's code are we actually reusing? Is it worth it to bring it in this way?

src/components/geolocate-control.js
xintongxia: Inside mapbox's `geolocate-control`, all the functions related to handle locating user, `trigger, _onSuccess, _onError`, changed the `classList` of `_geolocateButton`, so I need replace the  mapbox's `_geolocateButton` with our own react button element, then mapbox can manipulate the css of `_geolocateButton`. 

I did this because I think we should use Mapbox's `trigger, _onSuccess, _onError` instead of duplicate the code.

src/components/geolocate-control.js
xintongxia: We reuse most of the Mapbox's code. The main different parts is about set UI elements and update viewport.

However, this specific logic happened in their `_setupUI`, which we can not use. Because we need replace their code with react elements.

src/components/geolocate-control.js
```diff
@@ -0,0 +1,272 @@
...
+  }
+
+  _checkGeolocationSupport() {
+    // not necessary to check again
+    const {supportsGeolocation} = this.state;
+    if (supportsGeolocation !== undefined) {
+      return;
+    }
+
+    if (window.navigator.permissions !== undefined) {
```
Pessimistress: Separate the browser API handling from component state update.

```js
isGeolocationSupported().then(result => {
  this.setState(...);
  this._setupMapboxGeolocateControl(result);
});
```

src/components/geolocate-control.js
```diff
@@ -0,0 +1,272 @@
...
+      });
+    } else {
+      this.setState(
+        {supportsGeolocation: Boolean(window.navigator.geolocation)},
+        () => this._setupMapboxGeolocateControl(this.state.supportsGeolocation)
+      );
+    }
+  }
+
+  trigger = () => {
```
Pessimistress: Call it `onClick` for consistency

src/components/geolocate-control.js
```diff
@@ -0,0 +1,204 @@
...
+  // Callbacks fired when the user interacted with the map. The object passed to the callbacks
+  // contains viewport properties such as `longitude`, `latitude`, `zoom` etc.
+  onViewStateChange: PropTypes.func,
+  onViewportChange: PropTypes.func
+});
+
+const defaultProps = Object.assign({}, BaseControl.defaultProps, {
+  className: '',
+  // mapbox geolocate options
+  showUserLocation: true,
```
Pessimistress: All props not marked `isRequired` should be listed here.

src/components/geolocate-control.js
```diff
@@ -0,0 +1,204 @@
...
+  }
+
+  _containerRef: { current: null | HTMLDivElement } = createRef();
+  _geolocateButtonRef: { current: null | HTMLDivElement } = createRef();
+  _markerRef: { current: null | HTMLDivElement } = createRef();
+
+  componentDidMount() {
+    isGeolocationSupported().then(result => {
+      this.setState({supportsGeolocation: result}, () => {
+        this._setupMapboxGeolocateControl(this.state.supportsGeolocation);
```
Pessimistress: No need to stack these, just call `this._setupMapboxGeolocateControl(result)`

src/components/geolocate-control.js
```diff
@@ -0,0 +1,204 @@
...
+    const controlOptions = {};
+    ['positionOptions', 'fitBoundsOptions', 'trackUserLocation', 'showUserLocation'].forEach(
+      prop => {
+        if (prop in this.props) {
+          controlOptions[prop] = this.props[prop];
+        }
+      }
+    );
+
+    this._mapboxGeolocateControl = new MapboxGeolocateControl(controlOptions);
```
Pessimistress: Anything we need to do on unmount?

src/components/geolocate-control.js
```diff
@@ -0,0 +1,204 @@
...
+
+    // replace mapbox internal prop
```
Pessimistress: Add a comment here that the following code re-implements `MapboxGeolocateControl._setupUI`

src/components/geolocate-control.js
```diff
@@ -0,0 +1,204 @@
...
+export default class GeolocateControl extends BaseControl {
+  static propTypes = propTypes;
+  static defaultProps = defaultProps;
+
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      supportsGeolocation: undefined,
+      markerPosition: {
```
Pessimistress: Marker should not be rendered before the position is found. You will be able to see this placeholder at low zoom levels.

src/components/geolocate-control.js
```diff
@@ -69,12 +68,21 @@ export default class GeolocateControl extends BaseControl {
...
-      });
+      this.setState({supportsGeolocation: result});
+      this._setupMapboxGeolocateControl(result);
     });
   }
 
+  componentWillUnmount() {
+    // re-implement MapboxGeolocateControl's _onRemove
+    // clear the geolocation watch if exists
+    const geolocationWatchID = this._mapboxGeolocateControl._geolocateButton;
```
Pessimistress: `this._mapboxGeolocateControl._geolocationWatchID`?

src/utils/geolocate-utils.js
```diff
@@ -4,7 +4,7 @@ let supported;
 export function isGeolocationSupported() {
   // not necessary to check again
   if (supported !== undefined) {
-    return new Promise().then(() => supported);
+    return new Promise((resolve, reject) => resolve(supported));
```
Pessimistress: Use `Promise.resolve`

src/components/geolocate-control.js
```diff
@@ -180,6 +192,12 @@ export default class GeolocateControl extends BaseControl {
       captureDrag: false,
       captureDoubleClick: false
     });
+
+    // MapboxGeolocateControl needs manipulate the style of Marker's container
+    const markerRef = this._markerRef.current;
+    this._mapboxGeolocateControl._dotElement = markerRef && markerRef._containerRef.current;
```
Pessimistress: This needs to be in `componentDidUpdate`

src/components/geolocate-control.js
```diff
@@ -0,0 +1,224 @@
...
+      type: 'button',
+      title: label,
+      onClick: callback,
+      children
+    });
+  };
+
+  _renderMarker = () => {
+    const {markerPosition} = this.state;
+    if (!markerPosition) {
```
Pessimistress: Need to check `showUserLocation` too?

xintongxia: checked in the `render()`.

Maybe I could move that logic to `_renderMarker`.

jaydenseric:
It seems this feature has not been published yet, even though it was merged into the `master` branch 5 days before the latest release?

`GeolocateControl` is not in the exports:

https://unpkg.com/react-map-gl@4.0.13/dist/es5/index.js

sdll:
@Pessimistress, @xintongxia, let me know if there is anything I can help with to expedite the integration of this feature with the upstream.

PR #730, opened 2019-02-15
Updated docs for using non-mapbox sourced tiles

thomasgwatson:
Repushing this with proper author on commit! Soz for the double PR!


thomasgwatson:
#722 clone, but with proper authorship

PR #733, opened 2019-02-20
Add onNativeClick callback

Pessimistress:
For https://github.com/uber/react-map-gl/issues/569#issuecomment-465193236

- Fix regression where `onClick` is fired with delay even if `doubleClickZoom` is `false`
- Add `onNativeClick` prop per discussion

react-map-click

### DIFF
--- a/src/components/base-control.js
+++ b/src/components/base-control.js
@@ -74,8 +74,9 @@ export default class BaseControl extends PureComponent<ControlProps> {
       this._events = {
         wheel: this._onScroll,
         panstart: this._onDragStart,
+        anyclick: this._onClick,
         click: this._onClick,
-        pointerup: this._onPointerUp
+        dblclick: this._onDblClick
       };
       eventManager.on(this._events, ref);
     }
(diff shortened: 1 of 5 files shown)


src/components/base-control.js
```diff
@@ -104,18 +105,9 @@ export default class BaseControl extends PureComponent<ControlProps> {
     }
   }
 
-  _onPointerUp = (evt: MjolnirEvent) => {
+  _onDblClick = (evt: MjolnirEvent) => {
     if (this.props.captureDoubleClick) {
-      const {eventManager} = this._context;
```
Pessimistress: This hack is no longer necessary with the new click handler in `InteractiveMap`

src/components/base-control.js
```diff
@@ -104,18 +105,9 @@ export default class BaseControl extends PureComponent<ControlProps> {
...
-      const {options: {enable}} = eventManager.manager.get('doubletap');
-      // Temporarily disable doubletap
-      if (enable) {
-        eventManager._toggleRecognizer('doubletap', false);
-        /* global setTimeout */
-        setTimeout(() => {
-          eventManager._toggleRecognizer('doubletap', true);
-        }, 0);
```
xintongxia: I am surprised that even with `0` timeout works.

Issue #735, opened 2019-02-21
Rendering problem with Mapbox custom layer


ThisIsBarney:
**Version**: 4.0.11
**Problem**: When using custom layer in Mapbox, after I dragged the map, the render method (shown below) will **no longer be triggered** until I start to drag the map again.
```
class CustomLayer {
  constructor() {
    this.id = 'custom_layer';
    this.type = 'custom';
    this.renderingMode = '3d';

    // xxxxxxx
  }

  onAdd = (map: mapboxgl.Map, gl: WebGLRenderingContext) => {
    // some initialization
  }

  render = (gl: WebGLRenderingContext, matrix: number[]) => {
    // nothing will appear here after drag the map
  }
}

onViewportChange = (viewport) => {
    this.setState({
      viewport: {
        ...this.state.viewport, ...viewport
      },
    })
  }

render() {
    const {
      viewport,
    } = this.state;
    return (
      <div>
        <MapGL
          {...viewport}
          mapStyle={MAP_STYLE}
          onViewportChange={this.onViewportChange}
          ref={ref => { this.mapbox = ref; }}
        >
        </MapGL>
      </div>
    );
  }
```

BTW, there is no problem with Version 4.0.10.

Pessimistress:
@ThisIsBarney I am unable to reproduce this behavior using your code sample. `render` is still triggered. Can you provide more context of your use case? Do you have animations in the custom layer?

Between 4.0.10 and 4.0.11 we introduced a change where we redraw the map synchronously inside the React render cycle. More context [here](https://github.com/mapbox/mapbox-gl-js/issues/7893).

ThisIsBarney:
@Pessimistress Yes, I have animations in the custom layer to update 3d Objects. I use three.js in custom layer. The code of animation is like:
```
animate = () => {
  // do some transformations of the three.js objects
  
  window.requestAnimationFrame(this.animate);
}
```
`render` behaves correct if I don't drag the map, but after I dragged the map and stopped dragging, the `render` stops. The `animate` function still get called every frame when `render` stops.

Pessimistress:
Are you using `triggerRepaint` to update the frame?

ThisIsBarney:
@Pessimistress YES, I use `triggerRepaint` in render function of the custom layer class

Pessimistress:
OK that makes sense now. I can fix this.
Though I still think using `triggerRepaint` inside a render function is weird. Where is the `animate` function living then?

[...]

PR #736, opened 2019-02-22
Fix synchronous redraw bug

Pessimistress:
For https://github.com/uber/react-map-gl/issues/735

### DIFF
--- a/src/mapbox/mapbox.js
+++ b/src/mapbox/mapbox.js
@@ -221,12 +221,13 @@ export default class Mapbox {
     const map = this._map;
     // map render will throw error if style is not loaded
     if (map.isStyleLoaded()) {
-      map._render();
       // cancel the scheduled update
       if (map._frame) {
         map._frame.cancel();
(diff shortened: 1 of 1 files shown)


src/mapbox/mapbox.js
```diff
@@ -221,12 +221,13 @@ export default class Mapbox {
...
     // map render will throw error if style is not loaded
     if (map.isStyleLoaded()) {
-      map._render();
       // cancel the scheduled update
       if (map._frame) {
         map._frame.cancel();
         map._frame = null;
       }
+      // the order is important - render() may schedule another update
+      map._render();
```
xintongxia: Correct me if my understanding is wrong. 

`map._render() will trigger another update` means in Mapbox's `render()` function will request another animation frame, which will be assigned to `map._frame`.

Pessimistress: Mapbox's `render` does not, but in the issue linked above, a custom layer does.

PR #753, opened 2019-03-13
Support JSX syntax

xintongxia:


xintongxia:
This will be covered by another PR for upgrading dev tools.

PR #754, opened 2019-03-14
add release script

xintongxia:
replace all links with specified branch in docs and examples


Pessimistress:
The script should also update the website links: https://github.com/uber/deck.gl/blob/master/scripts/update-release-branch.sh#L17

PR #761, opened 2019-03-15
Enhance GeolocateControl component

### LINKED ISSUE #759: [GeolocateControl]: How to get `markerPosition`?
The `GeolocateControl` component can update and track the user's position. But I also need to perform some operation with it's data.

Is there any way to get access to the `GeolocateControl` component's `this.state.markerPosition`?

MunifTanjim:
- [x] add callback for `mapbox-gl-js` geolocatecontrol's `geolocate` event _(resolves #759)_
- [x] support `BACKGROUND` `_watchState` if the Map is panned while `trackUserLocation` is `true`

**Note to Reviewers**:
1. Should this component also add support for the other events related to GeolocateControl?
    - `error`
    - `trackuserlocationstart`
    - `trackuserlocationend`
2. Should this component support a prop (`triggerOnMount`) to activate tracking without first clicking the button?


src/components/geolocate-control.js
```diff
@@ -125,6 +135,22 @@ export default class GeolocateControl extends BaseControl {
     this._mapboxGeolocateControl._updateCamera = this._updateCamera;
 
     this._mapboxGeolocateControl._setup = true;
+
+    // when the camera is changed (and it's not as a result of the Geolocation Control) change
+    // the watch mode to background watch, so that the marker is updated but not the camera.
+    if (this._mapboxGeolocateControl.options.trackUserLocation) {
+      this._context.eventManager.on('panstart', () => {
```
Pessimistress: Many user interactions may lead to camera change. I think a more reliable event here is:

```js
this._context.map.on('move', ...)
```

MunifTanjim: I didn't see the `move`/`movestart` event in the [MapController class docs](https://uber.github.io/react-map-gl/#/Documentation/api-reference/map-controller).

If they do, can you give me any hints about when they fire?

Pessimistress: I'm suggesting that you listen to an event on the Mapbox Map instance, instead of an event of the EventManager.

The EventManager event `panstart` is an input event. It refers to when the user starts dragging on the canvas. Exactly *how* that gesture manipulates the camera is up to the implementation of the MapController.

The Mapbox `movestart` event is a camera event. It refers to any change to the current viewport.

Input events such as `pan`, `wheel`, and `pinch` can all cause the camera to move. But `movestart` will fire in all of the above circumstances.

MunifTanjim: @Pessimistress 
I did some tests and EventManager's `panstart` is a more appropriate event in this case.

In the `mapbox-gl-js` library, `map.fitBounds()` function can pass data to their internal events. In the `_updateCamera` function, they use that feature to pass `geolocateSource: true` flag to the `movestart` events:

https://github.com/mapbox/mapbox-gl-js/blob/90d2d48a5a0dfff4f3365a407deb087bba7a02a0/src/ui/control/geolocate_control.js#L193-L195

And in the `movestart` listener, they check that flag to filter out the automatic camera events (without user interaction), so the listener function essentially runs only when the user moves the map:

https://github.com/mapbox/mapbox-gl-js/blob/90d2d48a5a0dfff4f3365a407deb087bba7a02a0/src/ui/control/geolocate_control.js#L297-L298

We can't pass data to mapbox-gl-js's internal events (as we don't use `map.fitBounds()`), so we can't set the `geolocateSource: true` flag in `movestart` events. So, it's not possible to use 'movestart` in this case.

But listening EventManager's `panstart` works as expected. No need to filter out automatic camera events and only runs when user pans the map.

MunifTanjim:
The build is failing because of an unrelated error. :roll_eyes:

jgibson02:
Could support for `error`, `trackuserlocationstart`, and `trackuserlocationend` be added?

src/components/geolocate-control.js
Pessimistress: Ok that makes sense. The change itself looks good to me; could you also add the new prop to the documentation of this component?

MunifTanjim: done

PR #769, opened 2019-04-02
Custom mapbox server

sumarlidason:
enables configuration of the mapbox endpoint

ref https://github.com/mapbox/kepler.gl/issues/1

### DIFF
--- a/src/mapbox/mapbox.js
+++ b/src/mapbox/mapbox.js
@@ -95,7 +95,7 @@ const defaultProps = {
 type MapboxGL = {
   version: string,
   accessToken: string,
-  config: Object,
+  baseApiUrl: string,
   Map: Function
 };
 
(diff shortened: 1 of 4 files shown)


src/mapbox/mapbox.js
```diff
@@ -37,6 +37,7 @@ const propTypes = {
   gl: PropTypes.object, /** External WebGLContext to use */
 
   mapboxApiAccessToken: PropTypes.string, /** Mapbox API access token for Mapbox tiles/styles. */
+  mapboxApiUrl: PropTypes.string,
```
Pessimistress: Add this to the `StaticMap` and `InteractiveMap` components documentation as well?

sumarlidason: @Pessimistress thanks for 👀; updated 👍

src/mapbox/mapbox.js
```diff
@@ -346,6 +350,7 @@ export default class Mapbox {
 
     // Creation only props
     this.mapboxgl.accessToken = props.mapboxApiAccessToken || defaultProps.mapboxApiAccessToken;
+    this.mapboxgl.config.API_URL = props.mapboxApiUrl;
```
xintongxia: there is another prop `baseApiUrl` documented in [mapbox-gl-js](https://docs.mapbox.com/mapbox-gl-js/api/#baseapiurl), do you know the difference between `baseApiUrl` and `this.mapboxgl.config.API_URL`?

sumarlidason: @xintongxia you are right, that method has been available since v0.52. I'll update it.

PR #772, opened 2019-04-13
Fix synchronous redraw bug

Pessimistress:
Apparently `map.isStyleLoaded()` can return `false` after the map is loaded. Change the safety check to align with that in the mapbox-gl.

This should *really* fix the synchronization issue between canvas and other components.

Also adding an option to revert to v3 behavior for those who are experiencing issue with this change.

### DIFF
--- a/src/mapbox/mapbox.js
+++ b/src/mapbox/mapbox.js
@@ -48,7 +48,7 @@ const propTypes = {
   ]), /** The Mapbox style. A string url to a MapboxGL style */
 
   visible: PropTypes.bool, /** Whether the map is visible */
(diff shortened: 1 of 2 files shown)


src/mapbox/mapbox.js
```diff
@@ -48,6 +48,7 @@ const propTypes = {
   ]), /** The Mapbox style. A string url to a MapboxGL style */
 
   visible: PropTypes.bool, /** Whether the map is visible */
+  asyncRender: PropTypes.bool, /** Whether the map should manage its own render cycle */
```
xintongxia: Maybe say `Mapbox` instead of map?

PR #777, opened 2019-04-24
Update DeckGL overlay examples

draperunner:
This PR does the following for the DeckGL overlay examples:
* upgrades deck.gl to v7
* uses viewState prop 
* adds a dummy mapboxApiAccessToken prop

I used this example to add DeckGL to my ReactMapGL map, but spent quite a while to figure out that I needed to put the viewport in the `viewState` prop. Apparently the example was using an older version of deck.gl.


xintongxia:
@draperunner 

Thanks for your contribution.

Could you help resolve the conflicts?

PR #798, opened 2019-05-28
Fix transition manager tests

Pessimistress:


test/src/utils/transition-manager.spec.js
```diff
@@ -404,6 +406,9 @@ test('TransitionManager#TRANSITION_EVENTS', t => {
         testCase.shouldChange[mode].transitionEasing,
         'transitionEasing match'
       );
+
+      // Clean up
+      transitionManager._endTransition();
```
xintongxia: Why this is needed to be called in tests? Is this a step that whoever uses TransitionManager's also need to do?

test/src/utils/transition-manager.spec.js
```diff
@@ -395,7 +397,7 @@ test('TransitionManager#TRANSITION_EVENTS', t => {
       // testing easing function
       let testEasingFunc = testCase.input[ti].transitionEasing;
       if (mode === TRANSITION_EVENTS.UPDATE) {
-        const completion = (time[1] - time[0]) / testCase.input[ti].transitionDuration;
+        const completion = 100 / testCase.input[ti].transitionDuration;
```
xintongxia: Where did you get this number `100`?

Pessimistress: https://github.com/uber/react-map-gl/pull/798/files#diff-a02c4dd2c586d5f96eeeb6581028fb59R376

test/src/utils/transition-manager.spec.js
Pessimistress: We provided an external timer so the animation never ends. The test cannot end if there is a pending animation frame requested.

PR #816, opened 2019-06-04
Fix style diff warning when reusing map

Pessimistress:
Always set style from scratch when creating a new `Mapbox` instance from a recycled `Map`


src/mapbox/mapbox.js
```diff
@@ -271,7 +271,7 @@ export default class Mapbox {
     // Step3: update style and call onload again
     if (props.mapStyle) {
       this._map.setStyle(props.mapStyle, {
-        diff: true
+        diff: false
```
xintongxia: add your comment here.

Issue #768, opened 2019-04-02
allow Popup to overflow outside map


huchenme:
Currently Popup got hide inside map, if a popup is big enough, user will not able to see popup outside map

Demo: https://codesandbox.io/s/531p2ynork

max-programming:
Wasn't this resolved? Is there no way to have the popup outside somehow?

PR #821, opened 2019-06-07
Fix overlay styles

xintongxia:
for https://github.com/uber/react-map-gl/issues/768

### DIFF
--- a/examples/controls/src/app.js
+++ b/examples/controls/src/app.js
@@ -81,6 +81,7 @@ export default class App extends Component {
         mapStyle="mapbox://styles/mapbox/dark-v9"
         onViewportChange={this._updateViewport}
         mapboxApiAccessToken={TOKEN}
+        style={{overflow: 'hidden'}}
       >
         {CITIES.map(this._renderCityMarker)}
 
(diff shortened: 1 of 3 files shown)


src/components/static-map.js
```diff
@@ -275,7 +275,10 @@ export default class StaticMap extends PureComponent<StaticMapProps, State> {
         createElement('div', {
           key: 'map-overlays',
           className: 'overlays',
-          style: CONTAINER_STYLE,
+          style: {
```
Pessimistress: Make another constant

examples/controls/src/app.js
```diff
@@ -81,6 +81,7 @@ export default class App extends Component {
         mapStyle="mapbox://styles/mapbox/dark-v9"
         onViewportChange={this._updateViewport}
         mapboxApiAccessToken={TOKEN}
+        style={{overflow: 'hidden'}}
```
Pessimistress: Why is this needed?

src/components/static-map.js
```diff
@@ -284,7 +289,7 @@ export default class StaticMap extends PureComponent<StaticMapProps, State> {
 
   render() {
     const {className, width, height, style, visibilityConstraints} = this.props;
-    const mapContainerStyle = Object.assign({position: 'relative'}, style, {
+    const mapContainerStyle = Object.assign({position: 'relative', overflow: 'inherit'}, style, {
```
Pessimistress: This is not going to work if `StaticMap` is used standalone.

xintongxia:
This is hacky and cannot resolve the issue. This PR will make both markers and popups overflow on a basemap

Issue #676, opened 2018-12-11
Popup compresses on right side of screen


wafisher:
When using a popup, the popup shrinks its content as it approaches the right side of the screen, but this doesn't happen on the left. (See here: [Wrapping div.mov.zip](https://github.com/uber/react-map-gl/files/2667840/Wrapping.div.mov.zip))

The reason is that `top` and `left` are specified, the div expands its width to the right side of the screen as necessary and then the popup is transformed via CSS to its final position.

I fixed this with `style={{whiteSpace: 'nowrap'}}` on the content `<div>` but I wonder if this is intended behavior or if there should be something in the library to force the popup not to shrink its content.

PR #823, opened 2019-06-12
Fix compressed popup when reaching right bound of map

xintongxia:
For https://github.com/uber/react-map-gl/issues/676

### DIFF
--- a/src/components/popup.js
+++ b/src/components/popup.js
@@ -40,6 +40,8 @@ const propTypes = Object.assign({}, BaseControl.propTypes, {
   offsetLeft: PropTypes.number,
   // Offset from the top
   offsetTop: PropTypes.number,
+  // Custom CSS for the container
+  style: PropTypes.object,
   // Size of the tip
   tipSize: PropTypes.number,
   // Whether to show close button
@@ -61,6 +63,7 @@ const defaultProps = Object.assign({}, BaseControl.defaultProps, {
   altitude: 0,
   offsetLeft: 0,
   offsetTop: 0,
+  style: {},
   tipSize: 10,
   anchor: 'bottom',
   dynamicPosition: true,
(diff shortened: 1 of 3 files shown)


Pessimistress:
If I'm reading this correctly, it doesn't fix anything, just offers an option to fix it?

xintongxia:
> If I'm reading this correctly, it doesn't fix anything, just offers an option to fix it?

I added a prop `minWidth`

Pessimistress:
That is a breaking change and we can't publish it in a patch release.
- Does mapbox apply minWidth to their popups, if so, what's the value?
- Add to what's new and upgrade guide

src/components/popup.js
```diff
@@ -131,14 +135,17 @@ export default class Popup extends BaseControl<PopupProps, *, HTMLDivElement> {
...
     const style = {
       position: 'absolute',
-      left: x + offsetLeft,
-      top: y + offsetTop,
-      transform: `translate(${-anchorPosition.x * 100}%, ${-anchorPosition.y * 100}%)`,
+      transform: `
+        translate(${-anchorPosition.x * 100}%, ${-anchorPosition.y * 100}%)
+        translate(${left}px, ${top}px)
+      `,
+      maxWidth,
```
xintongxia: https://github.com/mapbox/mapbox-gl-js/blob/master/src/ui/popup.js#L406

src/components/popup.js
```diff
@@ -40,6 +40,8 @@ const propTypes = Object.assign({}, BaseControl.propTypes, {
   offsetLeft: PropTypes.number,
   // Offset from the top
   offsetTop: PropTypes.number,
+  // A number that sets the CSS property of the popup's maxWidth in pixels
+  maxWidth: PropTypes.number,
```
xintongxia: https://github.com/mapbox/mapbox-gl-js/blob/master/src/ui/popup.js#L31

PR #870, opened 2019-08-20
Fix react-map-gl-draw dependencies

xintongxia:
resolve error for nebula's dependency `@turf/difference`

https://github.com/Turfjs/turf/issues/1383#issuecomment-434486339

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -15,6 +15,7 @@
     "lint": "eslint src --ignore-pattern workers"
   },
   "dependencies": {
+    "@turf/area": "^6.0.1",
     "@turf/bbox": "^6.0.1",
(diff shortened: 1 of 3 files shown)


website/package.json
```diff
@@ -56,6 +59,9 @@
     "webpack-hot-middleware": "^2.17.0",
     "webworkify-webpack-dropin": "^1.1.9"
   },
+  "resolutions": {
+    "@turf/difference": "6.0.1"
+  },
```
cmonacaps: 👍

cmonacaps:
@xintongxia Is there a workaround while this is under review?  Thank you.

PR #875, opened 2019-08-29
Add babel react preset and migrate react components to jsx

MoradiDavijani:
I added `@babel/preset-react` and refactored React components to use JSX syntax instead of `createElement`.

Related issue: #871


MoradiDavijani:
Any opinion on this? :)

PR #913, opened 2019-10-22
fixes Issue 192: fixed geolocate control bug

sudoStatus200:
Hi, 
I created an issue just a moment ago and solution was very easy, So I  have created a pull request after adding an if condition to peice of code which was causing issue. 

Thanks,


xintongxia:
Thanks. Could you run `npm run lint fix` locally and update the PR?

sudoStatus200:
@xintongxia  YEah sure

PR #916, opened 2019-10-24
Fix for issue #908: source and layer cp - source clean up

### LINKED ISSUE #908: Error, when show/hide geojson layer with Source and Layer components
Hi,

I got problem when using Source and Layer components. In my case i need to show or hide geojson layer.
 My component looks like this:
```
import React from "react";
import { Layer, Source } from "react-map-gl";
import data from "data/GEOJSON/YVER-64_MT.geojson";

const fillLayer = {
  id: "fill",
  type: "fill",
  "paint": {
    "fill-color": [ "get", "fill" ],
    "fill-opacity": [ "get", "stroke-opacity" ],
  },
};

const lineLayer = {
  id: "line",
  type: "line",
  "paint": {
    "line-color": [ "get", "stroke" ],
    "line-opacity": [ "get", "fill-opacity" ],
    "line-width": [ "get", "stroke-width" ],
  },
};

const GeoJsonLayer = () => {
  return (
    <Source type="geojson" data={data}>
      <Layer {...fillLayer} />
      <Layer {...lineLayer} />
    </Source>
  );
};

export default GeoJsonLayer;
```

And depends on parent component state display/hide this GeoJosnLayer component.

I got error when hiding:
```
index.js:1437 Error: Source "jsx-source-2" cannot be removed while layer "fill" is using it.
    at i.removeSource (mapbox-gl.js:20157)
    at r.removeSource (mapbox-gl.js:24188)
    at Source.componentWillUnmount (source.js:44)
    at callComponentWillUnmountWithTimer (react-dom.development.js:21760)
    at HTMLUnknownElement.callCallback (react-dom.development.js:363)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:412)
    at invokeGuardedCallback (react-dom.development.js:465)
    at safelyCallComponentWillUnmount (react-dom.development.js:21767)
    at commitUnmount (react-dom.development.js:22252)
    at unmountHostComponents (react-dom.development.js:22728)
    at commitDeletion (react-dom.development.js:22764)
    at commitMutationEffects (react-dom.development.js:25274)
    at HTMLUnknownElement.callCallback (react-dom.development.js:363)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:412)
    at invokeGuardedCallback (react-dom.development.js:465)
    at commitRootImpl (react-dom.development.js:24997)
    at unstable_runWithPriority (scheduler.development.js:820)
    at runWithPriority$2 (react-dom.development.js:12280)
    at commitRoot (react-dom.development.js:24872)
    at finishSyncRender (react-dom.development.js:24249)
    at performSyncWorkOnRoot (react-dom.development.js:24223)
    at react-dom.development.js:12334
    at unstable_runWithPriority (scheduler.development.js:820)
    at runWithPriority$2 (react-dom.development.js:12280)
    at flushSyncCallbackQueueImpl (react-dom.development.js:12329)
    at flushSyncCallbackQueue (react-dom.development.js:12317)
    at discreteUpdates$1 (react-dom.development.js:24363)
    at discreteUpdates (react-dom.development.js:1485)
    at dispatchDiscreteEvent (react-dom.development.js:5970) "
```
Is there different approach to hide/show layer without this error ?

MKaHead:
Should fix #908

### DIFF
--- a/src/components/layer.js
+++ b/src/components/layer.js
@@ -100,7 +100,13 @@ export default class Layer<Props: LayerProps> extends PureComponent<Props> {
   }
 
   componentWillUnmount() {
-    this._map.removeLayer(this.id);
+    if (!this._map || !this._map.getStyle()) {
+      return;
+    }
+
+    if (this._map.getLayer(this.id)) {
+      this._map.removeLayer(this.id);
+    }
   }
 
   id: string;
--- a/src/components/source.js
+++ b/src/components/source.js
@@ -52,9 +52,9 @@ export default class Source<Props: SourceProps> extends PureComponent<Props> {
       return;
     }
 
-    const { id } = this.props;
+    const {id} = this.props;
     if (this._map.getSource(id)) {
-      const { layers } = this._map.getStyle();
+      const {layers} = this._map.getStyle();
       if (layers) {
         for (let i = 0; i <= layers.length; i++) {
           if (layers[i].source === id) {
@@ -65,7 +65,7 @@ export default class Source<Props: SourceProps> extends PureComponent<Props> {
     }
 
     this._map.removeSource(id);
-  };
+  }
 
   id: string;
   type: string;
--- a/src/components/source.js
+++ b/src/components/source.js
@@ -48,8 +48,24 @@ export default class Source<Props: SourceProps> extends PureComponent<Props> {
   }
 
   componentWillUnmount() {
-    this._map.removeSource(this.id);
-  }
+    if (!this._map || !this._map.getStyle()) {
+      return;
+    }
+
+    const { id } = this.props;
+    if (this._map.getSource(id)) {
+      const { layers } = this._map.getStyle();
+      if (layers) {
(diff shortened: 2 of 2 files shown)


Pessimistress:
Thank you for the PR. Some thoughts about the behavior:

My understanding of the issue: when both `<Source>` and `<Layer>` nodes are removed, we have to call `removeSource` and `removeLayer` in a particular order, however there is no guarantee which `componentWillUnmount` is called first. Your fix is that all layers referencing the to-be-removed source will be removed before the source itself. As a side effect, the following scenario might occur: If an app removes a `<Source>` node but not the `<Layer>` nodes referencing it, removal will be successful. If the app later adds the `<Source>` back in, the `<Layer>` nodes will be present in the React tree but not added back.

On principle, we want to maintain the alignment between the React representation and the underlying map state as well as we can. While it's always possible to manually call the native mapbox APIs and mutate the map state without going through the react-map-gl components, react-map-gl itself should avoid creating such discrepancy.

Issue #921, opened 2019-10-28
Locate User example not working when used in CRA


woss:
when i try to add the same code to a component and then call it within my react app i keep getting the following error,
```
mapbox-gl.js:33 Uncaught TypeError: Cannot read property 'getMaxBounds' of undefined
    at o._isOutOfMapMaxBounds (mapbox-gl.js:33)
    at o._onSuccess (mapbox-gl.js:33)
```

Map settings are the same as in the example.

In general all other (that i am using)  features are working.

Any help would be appreciated

Pessimistress:
This error seems to be inside mapbox-gl. Can you check which mapbox-gl version is installed in your project?

woss:
These are installed components
```
yarn list v1.19.1
├─ react-map-gl@5.1.0-alpha.1
├─ @mapbox/geojson-area@0.2.2
├─ @mapbox/geojson-rewind@0.4.0
├─ @mapbox/geojson-types@1.0.2
├─ @mapbox/jsonlint-lines-primitives@2.0.2
├─ @mapbox/mapbox-gl-supported@1.4.1
├─ @mapbox/point-geometry@0.1.0
├─ @mapbox/tiny-sdf@1.1.1
├─ @mapbox/unitbezier@0.0.0
├─ @mapbox/vector-tile@1.3.1
├─ @mapbox/whoots-js@3.1.0
└─ mapbox-gl@1.5.0
```

I wonder why copying 100% of the code from your example to separate file and creating the Storybook story fails.

niikNak:
Yeah I'm having the same issue. Have the same mapbox-gl version.

niikNak:
Downgrading to mapbox-gl@1.4.1 seemed to fix the issue. 
I'll be honest, I'm a bit new to reporting issues on public repos. Would it be a good idea to report the issue to the mapbox-gl-js team as well?

woss:
well, it would, but I am using this lib and their examples. Maybe there is nothing to do with the mapbox itself rather than passing the correct data from wrapped component to actual mapbox component.

VictorBaudot:
> Downgrading to mapbox-gl@1.4.1 seemed to fix the issue.

I personally still have the error after downgrading to 1.4.1

woss:
for me, it's gone. 
now I realised that location is off by close to 1km, investigating that 😁

TashTenner:
Same issue here - is downgrading the only way to solve it?

Pessimistress:
[...]

PR #934, opened 2019-11-05
Fix geolocate control crash

xintongxia:
For #921


src/components/geolocate-control.js
```diff
@@ -141,6 +141,10 @@ export default class GeolocateControl extends BaseControl<
 
     this._mapboxGeolocateControl = new mapboxgl.GeolocateControl(controlOptions);
 
+    if (this._context) {
```
Pessimistress: What if map is not initialized yet? Check before geolocating?

PR #953, opened 2019-12-06
Fix missing control icons when using mapbox-gl 1.6.0 (#952)

### LINKED ISSUE #952: Control icons missing with mapbox-gl-js 1.6.0
The icons for map controls are missing when using react-map-gl 5.1.3 with mapbox-gl-js 1.6.0. I expect this affects most earlier versions of react-map-gl too. The controls are functional but blank.

The 1.6.0 release of mapbox-gl-js changed the CSS and elements used for these controls from:

```html
<button class="mapboxgl-ctrl-icon mapboxgl-ctrl-fullscreen">
</button>
```

to

```html
<button class="mapboxgl-ctrl-fullscreen">
  <span class="mapboxgl-ctrl-icon" aria-hidden="true"></span>
</button>
```

with the icons now rendered into the new `<span>`s. The relevant commit is https://github.com/mapbox/mapbox-gl-js/commit/d59b3da4.

I'd be happy to submit a patch, but I wasn't sure how you'd like to handle backwards compatibility with older mapbox-gl-js versions. Something like the following combination of the old and new HTML might work, but I haven't tested the extra `<span>` for adverse effects on older versions.

```html
<button class="mapboxgl-ctrl-icon mapboxgl-ctrl-fullscreen">
  <span class="mapboxgl-ctrl-icon" aria-hidden="true"></span>
</button>
```

tsibley:
Adds necessary span.mapboxgl-ctrl-icon to each button but leaves the .mapboxgl-ctrl-icon class on the button too for backwards compat with versions less than 1.6.0.

I tried to test the examples with controls after force installing mapbox-gl 1.5.1 and 1.6.0 and it appeared to work in both. However, I'm not 100% confident of the mapbox-gl versions that actually got bundled by Webpack though, as the dep interactions are quite complex (and I don't know of a way to check the loaded version in the JS console).

Resolves #952.

### DIFF
--- a/src/components/fullscreen-control.js
+++ b/src/components/fullscreen-control.js
@@ -113,7 +113,9 @@ export default class FullscreenControl extends BaseControl<
         type="button"
         title={label}
         onClick={callback}
-      />
+      >
+        <span className="mapboxgl-ctrl-icon" aria-hidden="true" />
+      </button>
     );
   }
 
--- a/src/components/geolocate-control.js
+++ b/src/components/geolocate-control.js
@@ -211,7 +211,9 @@ export default class GeolocateControl extends BaseControl<
         type="button"
         title={label}
         onClick={callback}
-      />
+      >
+        <span className="mapboxgl-ctrl-icon" aria-hidden="true" />
+      </button>
     );
   };
 
--- a/src/components/navigation-control.js
+++ b/src/components/navigation-control.js
@@ -96,8 +96,16 @@ export default class NavigationControl extends BaseControl<
   _renderCompass() {
     const {bearing} = this._context.viewport;
     return (
-      <span className="mapboxgl-ctrl-compass-arrow" style={{transform: `rotate(${-bearing}deg)`}} />
+      <span
+        className="mapboxgl-ctrl-icon"
+        aria-hidden="true"
+        style={{transform: `rotate(${-bearing}deg)`}}
+      />
     );
+  }
+
+  _renderIcon() {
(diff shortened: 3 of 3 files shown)


Pessimistress:
@tsibley Thank you for fixing this. As far as I can tell this solution does not work with the older versions of mapbox-gl. Could you add a check for mapbox version similar to https://github.com/uber/deck.gl/blob/c55cbff9a733476e8786d0e9eff16b212057cce6/modules/mapbox/src/deck-utils.js#L105 and switch between the two class names?

tsibley:
@Pessimistress I should be able to do that. Can you give me pointers for reliably running the examples against a locally build react-map-gl and different exact versions of mapbox-gl? I'd like to be able to make sure a new patch that conditions the class on the version actually works. :-)

Pessimistress:
> Can you give me pointers for reliably running the examples against a locally build react-map-gl and different exact versions of mapbox-gl?

- add `"resolutions": {"mapbox-gl": "<version>"}` to the root package.json
- `yarn`
- cd into an example folder, run `yarn start-local`

Pessimistress:
@tsibley follow up on this - do you think you will be able to update the PR soon? I’m happy to pick it up from here if you’d like.

tsibley:
I won’t be able to pick this up until at least the week of 6 Jan. You’re more than welcome to pick it up before then!

Pessimistress:
Closing in favor of #967. Thank you for your help in addressing this issue!

PR #959, opened 2019-12-17
Allow customizing labels in NavigationControl and GeolocateControl

ahie:


ahie:
> Looks good. Can you also add these new props to the documentation?

Done.

Issue #951, opened 2019-12-04
Add ScaleControl component


Pessimistress:
https://docs.mapbox.com/mapbox-gl-js/api/#scalecontrol

```jsx
<MapGL ...>
  <ScaleControl maxWidth={100} unit="metric" />
</MapGL>
```

PR #966, opened 2019-12-23
Add ScaleControl

Pessimistress:
For #951


kylebarron:
Can there be a note in the docs noting that `ScaleControl` is included in >= v5.1.6? I've been struggling to include it and just realized I'm only using `5.1.3`.

Issue #872, opened 2019-08-23
onViewportChange Warning


lucdandrel:
Hello all, 
I'm using the latest version of React-Map-GL (5.0.10), and I've an warning in devtools, from onViexportChange. 

Any idea why ? 

Thanks ! 

<img width="428" alt="Capture d’écran 2019-08-23 à 10 46 14" src="https://user-images.githubusercontent.com/17253963/63579583-4af5d500-c593-11e9-9b3d-19b0a8c5463b.png">

My component : 
![carbon](https://user-images.githubusercontent.com/17253963/63579384-d0c55080-c592-11e9-9bfd-0db55f7aceac.png)

lucdandrel:
I've this error only in dev environnement not in production

allenchuang:
I'm getting the same issue. Seems like `_onViewportChange` is being called on render. Is this expected behavior?

mrnguyenpro123:
Me too. It always calls but viewport the same

jlubeck:
Same here

Pessimistress:
Try 5.1.6

PR #968, opened 2019-12-30
Fix React state update warning

Pessimistress:
For https://github.com/uber/react-map-gl/issues/872

- Avoid calling `onViewportChange` during `render` (move to `componentDidUpdate`)
- Fix an issue during initial `MapController.setOptions` -- we were always calling `onViewportChange` because `oldViewport` was empty.

### DIFF
--- a/src/utils/map-controller.js
+++ b/src/utils/map-controller.js
@@ -135,7 +135,7 @@ export default class MapController {
   /* Callback util */
   // formats map state and invokes callback function
   updateViewport(newMapState: MapState, extraProps: any = {}, extraState: any = {}) {
-    const oldViewport = this.mapState ? this.mapState.getViewportProps() : {};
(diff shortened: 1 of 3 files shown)


src/utils/map-controller.js
```diff
@@ -176,12 +176,14 @@ export default class MapController {
     this.onViewportChange = onViewportChange;
     this.onStateChange = onStateChange;
 
-    if (!this.mapStateProps || this.mapStateProps.height !== options.height) {
+    const dimensionChanged = !this.mapStateProps || this.mapStateProps.height !== options.height;
```
xintongxia: Don't need check whether width changed here?

Pessimistress: Width does not affect normalization. A map may need to be scaled and shifted vertically to fill the height.

PR #970, opened 2020-01-02
Add clusters example

Pessimistress:
Reproduces Mapbox's [Create and style clusters](https://docs.mapbox.com/mapbox-gl-js/example/cluster/) example.

### DIFF
--- a/examples/clusters/src/app.js
+++ b/examples/clusters/src/app.js
@@ -1,0 +1,78 @@
+import React, {Component} from 'react';
+import {render} from 'react-dom';
+import MapGL, {Source, Layer} from 'react-map-gl';
+
+import ControlPanel from './control-panel';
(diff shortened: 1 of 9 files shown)


examples/clusters/src/app.js
```diff
@@ -0,0 +1,78 @@
...
+  _onViewportChange = viewport => this.setState({viewport});
+
+  _onClick = event => {
+    const feature = event.features[0];
+    const clusterId = feature.properties.cluster_id;
+
+    const mapboxSource = this._sourceRef.current.getSource();
+
+    mapboxSource.getClusterExpansionZoom(clusterId, (err, zoom) => {
+      if (err) return;
```
xintongxia:
```
if (err) {
  return;
}
```

Issue #996, opened 2020-01-16
MapState is not exported from index.js


HasanAyan:
Hi, I would like to do various transitions from child components of InteractiveMap. 

I looked at the source code of NavigationControl and it is using `context.ViewportChange` ,`context.ViewStateChange`,` context.viewport` and `MapState`

To give you more context, I am trying to create a ClickZoomableLayer which will render Layer component with given props, bind to `anyclick` event on context.eventManager, and fly to clicked feature's position.

However, it seems MapState class is not exported from `index.js` and I can't import it from `'/utils/map-state.js'` since typecsript is complaining about it. Is there a workaround?

HasanAyan:
I have opened a new pull request to fix this issue 

Export MapState from index.js so it is available for import by consumers #997

Pessimistress:
What features are you trying to use on `MapState`? You can simply call `onViewStateChange` with the updated view state `{longtidue: <number>, latitude: <number>, zoom: <number>, transitionInterpolator: new FlyToInterpolator(), transitionDuration: <number>}`.

HasanAyan:
alright @Pessimistress , I tried it and worked but I am not sure about the parameters this function takes. 
what is the use of interactionState and newViewState ?

Is this correct way to do it?

```
            <ReactMapGL
              {...viewportState}
              // onViewportChange={viewport => {
              //   setViewportState(state => ({ ...state, ...viewport }));
              // }}
              onViewStateChange={viewStateChangeInfo => {
                setViewportState(state => ({
                  ...state,
                  ...viewStateChangeInfo.viewState
                }));
              }}/>
```


and when I want to focus on something;

```
export const fly = (
  mapContext: MapContextProps,
  params: Partial<ViewportProps>
) => {
  mapContext.onViewStateChange &&
    mapContext.onViewStateChange({
      viewState: {
        ...params,
        transitionDuration: 300,
        transitionInterpolator: new FlyToInterpolator(),
        transitionInterruption: TRANSITION_EVENTS.BREAK
      }
    } as any);
};
```

Pessimistress:
That looks fine. `interactionState` is used by some applications to determine what interaction triggered the view state change (pan, rotate, zoom, transition etc.). If your app doesn't need it then you don't have to send anything.

PR #997, opened 2020-01-16
Export MapState from index.js so it is available for import by consumers

HasanAyan:
Currently, although MapState type is included in the typings, index.js doesn't actually export MapState and when a typescript using consumer imports it, editor does not complain but the build fails due to missing export statement. On the other hand, importing it directly from 'react-map-gl/src/utils/map-state' is not possible since typescript complains about types being missing.

So I added the missing import statement to the index.js file

I have also opened an issue to describe why it is necessary.

**MapState is not exported from index.js** #996


HasanAyan:
Closing as it seems I don't need it actually

PR #1038, opened 2020-02-24
Migrate to import * as React from 'react'

TrySound:
Ref https://github.com/facebook/react/pull/18102

This is the recommended way to import react now.


TrySound:
Would be good to have oculus installed locally. Now precommit hook prevents committing without some abstract oculus tools.

Issue #898, opened 2019-09-24
popup text is blurry


lolcookie:
I noticed that when you use popups the text is blurry, this is even noticeable on
https://uber.github.io/react-map-gl/#/Examples/markers-popups

Is this fixable at all?

xintongxia:
<img width="617" alt="Screen Shot 2019-09-29 at 2 56 38 PM" src="https://user-images.githubusercontent.com/2927808/65840096-5f7f6880-e2c9-11e9-8b20-502be5ce5094.png">

From my observation, it it not blurry, as shown above. Or are you considering this blurry?

lolcookie:
![ReactMapGL-1](https://user-images.githubusercontent.com/9392694/65935677-85814600-e44c-11e9-8496-22822cfe6eeb.png)

This may only be an issue with low dpi displays 🤔

rilexus:
I have the same issue

pzi:
See related issue on the Mapbox repo: https://github.com/mapbox/mapbox-gl-js/pull/3258

heygrady:
The problem is here: https://github.com/uber/react-map-gl/blob/master/src/components/popup.js#L141-L142

```js
// this would greatly improve the issue
const style = {
      position: 'absolute',
      transform: `
        translate(${-anchorPosition.x * 100}%, ${-anchorPosition.y * 100}%)
        translate(${Math.round(left)}px, ${Math.round(top)}px)
      `,
      display: undefined,
      zIndex: undefined
    };
```

The core issue is positioning on decimal pixels. If you are on an older laptop (like a 2012 MacBook Air) you will have a 1x display, so half-pixels cause a huge problem. But even on brand new laptops that have 2x and 3x displays the partial pixels will cause the overlays to look blurry.

To "really" fix the issue you'd need to know the `offsetHeight` and `offsetWidth` as well as the `window.devicePixelRatio` to determine if the transform would cause the element to be rendered "blurry".

```js
const pixelRatio = window.devicePixelRatio || 1

const crispPixel = (size) => Math.round(size * pixelRatio) / pixelRatio

const crispPercentage = (el, percentage, dimension = 'width') => {
  const origSize = dimension === 'width' ? el.offsetWidth : el.offsetHeight
  return (crispPixel(percentage / 100 * origSize) / origSize) * 100
}

const el = this._containerRef.current

const style = {
      position: 'absolute',
      transform: `
[...]

PR #1039, opened 2020-02-25
Fix issue where translate by partial pixels causes blurry rendering

heygrady:
https://github.com/uber/react-map-gl/issues/898#issuecomment-580899897

The core issue is positioning on decimal pixels. If you are on an older laptop (like a 2012 MacBook Air) you will have a 1x display, so half-pixels cause a huge problem. Even on brand new laptops (that have 2x and 3x displays) the partial pixels will cause the overlays to look blurry.

To really fix the issue you'd need to know the `offsetHeight` and `offsetWidth` as well as the `window.devicePixelRatio` to determine if the transform would cause the element to be rendered "blurry".

### DIFF
--- a/src/components/marker.js
+++ b/src/components/marker.js
@@ -21,6 +21,7 @@
 import React from 'react';
 import PropTypes from 'prop-types';
 import DraggableControl from './draggable-control';
+import {crispPixel} from '../utils/crisp-pixel';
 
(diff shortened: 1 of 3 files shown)


Pessimistress:
CI is failing due to unrelated reasons. Looks like environmental variable settings are not applied to branches from forked repos.

PR #1041, opened 2020-02-27
Fix crispPercentage by adding an offset cache

heygrady:
Fix crispPercentage by adding an offset cache to avoid repeatedly calling el.offsetWidth

Follow up to https://github.com/uber/react-map-gl/pull/1039

It was bugging me that it calls `offsetWidth`/`offsetHeight` "too often". This keeps a temporary cache in a `WeakMap`, which should improve performance in re-render scenarios.


Pessimistress:
Does this make it impossible to update the dimensions of a popup? I can see apps changing a popup's layout if the user interacts with its content.

Pessimistress:
Closing for inactivity. This will be irrelevant in v7.

PR #1045, opened 2020-03-05
Fix invalid syntax in scale control docs

rubenmoya:
This PR fixes a syntax error found in the [Scale Control docs](https://uber.github.io/react-map-gl/docs/api-reference/scale-control)


Pessimistress:
Thank you for your submission! This seems to be a duplicate of #1025.

Issue #681, opened 2018-12-23
Flow types


normanrz:
It's great that you have added flow types to the codebase. What would be the best way to consume these types from my react application?

Pessimistress:
Flow types will be available in v5.3/v6.1.

PR #1108, opened 2020-05-21
Add @flow to index.js

NachoCasta:
Hi!
I just started using react-map-gl for a project I'm doing, where I'm also using flow. In order to do so, I had to create a stub for react-map-gl declaring all types to be any.
However, I just realized that the library is already flow typed, which is awesome, but flow still complains that I'm exporting from an untyped module.
Looking into #681, it seems I'm not the only one who noticed this. I'm not sure if this would solve the issue or what is the proper way to expose these types to the consumers but if it is not I would be happy to help if someone points me in the right direction.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,3 +1,4 @@
+// @flow
 // Copyright (c) 2015 Uber Technologies, Inc.
 
 // Permission is hereby granted, free of charge, to any person obtaining a copy


Pessimistress:
I don't believe this will work. We may need to do something like https://www.jackfranklin.co.uk/blog/npm-flowjs-javascript/

Pessimistress:
Looks like we just need to create an additional index.js.flow file: https://github.com/facebook/flow/issues/1996#issuecomment-228925018

Issue #904, opened 2019-10-03
Use .getMap() outside of Map Component 


l3git9:
Hi,

Working on a project that is getting quite large and I was wondering if there is a simple solution to using the map outside of the component. At the moment I am using a ref in order to use getMap() within the map component, but I have a search function that needs to flyTo the location when the user clicks on the result and I would like to be able to do so within another component. Is that possible or is everything going to need to funnel back into the Map component? I've looked everywhere for an example and have found nothing so I believe that I won't to in the manner that I want to do it. If I can't that's fine, but it would suck if there is and I just absolutely pass it up only to make it more complicated.

mayteio:
I use context inside React to get access to the map instance:

```javascript
// MapContext.js
import React, {createContext, useState} from 'react';

export const MapContext = createContext();

export function MapProvider(props){
  const [map, setMap] = useState();
  return <MapContext.Provider value={[map, setMap]} {...props} />
}

export function useMap(){
  const context = useContext(MapContext);
  if(context === undefined) throw Error('You forgot to wrap your app with <MapProvider />');
  return context;
}

// App.js
import React from 'react';
import {MapProvider} from './MapContext';

export default function App(){
  const [, setMap] = useMap();
  return (
    <MapProvider>
      <Example />
      <ReactMapGL
        ...
        ref={ref => ref && setMap(ref)}
      />
    </MapProvider>
  )
}

// Example.js
import React from 'react';
import {useMap} from './MapContext';

export default function Example(){
  const [map] = useMap();
  const onClick = () => {
    map && map.flyTo(...);.
  }

  useEffect(() => alert('map ready!'), [map])

  return <button onClick={onClick}>Fly like a bird</button>
}
```

Note; I have switched between using `useRef` and `useState` to store the context. They both seem to have tradeoffs; [useRef is probably the intended usage according to dan](https://twitter.com/dan_abramov/status/1105930134186532864?lang=en), but this makes it difficult to listen to it actually being set, i.e. with `useEffect`. [Depending on ref.current is problematic though](https://github.com/facebook/react/issues/14387#issuecomment-503616820).

useState on the other hand will trigger an effect re-run. Mutations to the map object should not trigger a re-render because it's the same object, just mutated.

I believe the correct way is to [use a callbackRef](https://reactjs.org/docs/hooks-faq.html#how-can-i-measure-a-dom-node), however I am yet to implement this for `react-map-gl`.

Maybe a feature request to expose a useMap hook? 🙏

[...]

PR #1109, opened 2020-05-21
Feature/maprefprovider

mayteio:
Hey @Pessimistress, I'm opening this early so we can discuss API et al before I go too deep in a solution that isn't compatible with the direction of the package.

The main benefit of this hook, in my experience, is to be able to access the mapbox instance anywhere in your app. At present (the way MapContext) is used, it's only possible to access the mapbox ref from a child of either `<InteractiveMap />` or `<StaticMap />` (as MapContext is only provided within those components).

Looking forward to further hook implementation, I've also found it beneficial to be able to do other things like set the viewport from elsewhere in the app. I've outlined how I approach this here: https://github.com/visgl/react-map-gl/issues/904#issuecomment-539689479. 

Essentially there are three components:
1. `<MapProvider />` (name is arbitrary at this state, MapHookProvider, MapRefProvider, etc)
2. `useMapRef` hook that consumes this provider (or useViewport, etc)
3. A wrapped `<StaticMap />` that uses the onLoad prop to store the mapbox ref in the context.

The main concern, I can imagine, is introducing greater API surface. I guess, are you OK introducing a public context provider to allow storing the ref and providing it to the hook?

What I have pushed so far will work, however, it'll only be accessible within the two map components. It will throw an error when used outside.

If you're happy with the additional public provider, I can go ahead and create it. I can then set the mapbox ref in StaticMap within the (private) `MapContext.Consumer` here:

https://github.com/visgl/react-map-gl/blob/15ae86247b934cd0a43f0eba1cb109c74c3eee6b/src/components/static-map.js#L264

My understanding is that this will bubble up for `InteractiveMap` as well.

Let me know.

### DIFF
--- a/test/src/components/map-provider.js
+++ b/test/src/components/map-provider.js
@@ -1,0 +1,35 @@
+import test from 'tape-catch';
+import {MapContext, MapProvider} from 'react-map-gl';
+import * as React from 'react';
+import ReactTestRenderer from 'react-test-renderer';
+import WebMercatorViewport from 'viewport-mercator-project';
+
+const testCase = (
+  <MapProvider>
+    <MapContext.Consumer>
+      {({viewport, map}) => (
+        <>
+          <div className="viewport">viewport: {JSON.stringify(viewport)}</div>
+          <div className="map">map: {JSON.stringify(map)}</div>
+        </>
+      )}
+    </MapContext.Consumer>
+  </MapProvider>
+);
+
+test.only('MapProvider - Provides a default viewport', t => {
+  const DEFUALT_VIEWPORT = {latitude: -37, longitude: 144, zoom: 15};
+  const result = ReactTestRenderer.create(
+    <MapProvider defaultViewport={new WebMercatorViewport(DEFUALT_VIEWPORT)}>
+      {testCase}
+    </MapProvider>
+  );
+
+  const resultRoot = result.root;
+
+  /** Assert the viewport is coming through the context consumer. */
+  t.ok(
+    resultRoot.findByProps('className', 'viewport').children[0],
+    `viewport: ${JSON.stringify(DEFUALT_VIEWPORT)}`
+  );
+});
--- a/src/components/map-provider.js
+++ b/src/components/map-provider.js
@@ -1,0 +1,51 @@
+// @flow
+import React, {useState, useContext, createContext} from 'react';
+import MapContext from './map-context';
+import WebMercatorViewport from 'viewport-mercator-project';
+import {Map} from 'mapbox-gl';
+
+type MapProviderProps = {
+  defaultViewport: ?WebMercatorViewport
+};
+
+const SetMapContext = createContext(undefined);
+
+const MapProvider: React$ComponentType<MapProviderProps> = ({defaultViewport, children}) => {
+  /** Store viewport in state */
+  const [viewport, setViewport] = useState<typeof undefined | WebMercatorViewport>(undefined);
+
+  /** Store viewport in state */
+  const [mapRef, setMapRef] = useState<typeof undefined | Map>(undefined);
+
+  return (
+    <MapContext.Provider
+      value={{
+        viewport,
+        map: mapRef,
+        eventManager: null,
+        isDragging: false,
+        mapContainer: null,
+        onViewStateChange: setViewport,
+        onViewportChange: setViewport
+      }}
+    >
+      <SetMapContext.Provider value={{setViewport, setMapRef}}>{children}</SetMapContext.Provider>
+    </MapContext.Provider>
+  );
+};
+
+export default MapProvider;
+
+/** Hook to expose the MapContext */
+export const useMap = () => {
+  const context = useContext(MapContext);
+  if (!context) throw Error('Not inside <MapProvider />');
+  return context;
+};
+
+/** Split out state updaters into it's own hook to prevent unnecessary re-renders (when viewport changes, often). */
+export const useSetMap = () => {
+  const context = useContext(SetMapContext);
+  if (!context) throw Error('Not inside <MapProvider />');
+  return context;
+};
--- a/src/components/use-map-ref.js
+++ b/src/components/use-map-ref.js
@@ -1,11 +1,14 @@
-import MapContext from './map-context';
+// @flow
+
 import {useContext} from 'react';
+import Mapbox from '../mapbox/mapbox';
+import MapContext from './map-context';
 
 /**
  * Hook to access underlying Mapbox instance. Use only available
  * inside a child component of InteractiveMap or StaticMap.
  */
-export const useMapRef = () => {
+export const useMapRef = (): Mapbox | null => {
   const context = useContext(MapContext);
   if (!context) throw Error('Trying to useMapRef outside of MapContext.');
   return context.map;
--- a/src/components/use-map-ref.js
+++ b/src/components/use-map-ref.js
@@ -1,0 +1,12 @@
+import MapContext from './map-context';
+import {useContext} from 'react';
+
+/**
+ * Hook to access underlying Mapbox instance. Use only available
+ * inside a child component of InteractiveMap or StaticMap.
+ */
+export const useMapRef = () => {
+  const context = useContext(MapContext);
+  if (!context) throw Error('Trying to useMapRef outside of MapContext.');
+  return context.map;
+};
--- a/yarn.lock
+++ b/yarn.lock
@@ -3612,7 +3612,7 @@ debug@3.1.0:
   dependencies:
     ms "2.0.0"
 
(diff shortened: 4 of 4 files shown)


Pessimistress:
CI failure is irrelevant - I will fix in another PR.

I'm not super familiar with custom hooks. What is the benefit of a new context provider over the current `MapContext`? Can we modify/repurpose the existing API (still marked experimental) to achieve the desired effect?

We need documentation for the new hook.

mayteio:
We can most certainly re-use MapContext. We still need a Provider to hoist the context to the whole app - which was what the original issue was about. Currently there is no Provider other than using the MapContext.Provider directly inside `InteractiveMap` and `StaticMap`. Therefore it's limited to any children of those components.

The hook simply consumes any ancestral MapContext. So, based off the logic above, you can use the useMapRef or useViewport hooks as children of the components, not as siblings or aunts/uncles (😂).

I will work on a solution today and push another commit.

mayteio:
Hey @Pessimistress, please see the `MapProvider` in `src/components/map-provider.js` and corresponding test in `test/src/components/map-provider.js`.

The way it works is:
1. User wraps their app in `<MapProvider />`.
2. `MapProvider` stores `viewport` and `mapRef` in state (useState hook)
3. `MapProvider` passes state via MapContext.Provider and updaters via (new) `SetMapContext.Provider`.
4. 2x hooks are exposed to consume these contexts, `useMap` and `useSetMap`

Now, if a user wants access to the map or viewport in their component, they can access it like so:
```js
const Example = () => {
  const {viewport, map} = useMap();
  useEffect(() => {
    map.addSource(...)
  // dependency array means this 'effect' will only run when the value of map changes (i.e. when we set it).
  }, [map])
}
```

If the user wants to update the viewport from anywhere in their app, they can do so;
```js
const Example = () => {
  const {setViewport} = useSetMap();
  const handleClick = () => setViewport({latitude: 40, longitude: 74, zoom: 10, transitionDuration: ...});
  return <button onClick={handleClick}>Go to New York</button>
```

Does that all make sense? 

Now, next we need to make static-map and interactive-map aware of the parent context **provided by `MapProvider`**.

test/src/components/map-provider.js
```diff
@@ -0,0 +1,35 @@
...
+        <>
+          <div className="viewport">viewport: {JSON.stringify(viewport)}</div>
+          <div className="map">map: {JSON.stringify(map)}</div>
+        </>
+      )}
+    </MapContext.Consumer>
+  </MapProvider>
+);
+
+test.only('MapProvider - Provides a default viewport', t => {
```
Pessimistress, line 20: .only

mayteio:
Sure, I can remove viewport management. I'll open another issue for that, I'm interested to hear how you'd approach it. 

Yes, totally regarding the mapRef. I guess it's a case of 'how much free functionality do we want to give the user?'. To access the mapRef from anywhere in the app as-is, you have to create your own context that provides the [map, mapRef] and a corresponding hook that consumes it. 

```
// MapRefContext.js
const MapRefContext = createContext();
export const MapRefProvider = props => {
  const mapState = useState();
  return <MapRefContext.Provider value={mapState} {...props} />
}
export const useMapRef = () => useContext(MapRefContext);

// App.js
<MapRefProvider>{children}</MapRefProvider>
```

```
// anywhere in your app
const [, setMapRef] = useMapRef();
<MapGL mapRef={ref => ref && setMap(ref.getMap())} ... />
```

That is the solution to the original issue #904 in its entirety (for which I already posted on that issue). We found ourselves re-writing this often. so we moved it, along with a slew of other things like viewport management/deckGL layer management/etc, into [@citydna/maps](https://www.npmjs.com/package/@citydna/maps).

I'd argue that there's more value in working on a set of hooks that manages viewport than the mapRef, because there's a lot more to consider. We built functionality for bounds, setting viewports to feature, etc.

Pessimistress:
That's fair. I'm curious to see how you connect this with the existing components. I am a little nervous about this mixture of hooks and class components in the source code (not 100% sure how they will interact).

mayteio:
Simply using `<MapContext.Consumer>` in the class components. It would consume a new `<MapProvider>` that handles `viewport`/`mapboxRef` state anywhere in the app (and uses `<MapContext.Provider>` to provide these things). 

So, do we want to close this one in favour of the solution described above and in #904 or are you still keen on getting a `useMapboxRef` hook going?

Pessimistress:
> Simply using `<MapContext.Consumer>` in the class components. 

We have discussed multiple approaches in this thread and I think I'm misunderstanding how the proposed feature works. Before we figure out what PR should the code go into, can we step back a little and talk about the design?

The way I see it, right now we use `MapContext` throughout the `ReactMapGL` stack, and you wish to have access to the context outside of the map component. I think that is a valid use case and it is not exactly easy with the current API.

Current

```js
import ReactMap from 'react-map-gl';

function Root() {
  const [map, setMap] = useState(null);
  const [viewport, setViewport] = useState({longitude, latitude, zoom});

  return (
    <>
       <Map setMap={setMap} viewport={viewport} setViewport={setViewport} />
       <OtherComponent map={map} viewport={viewport} />
    </>
  );
}

function Map({setMap, viewport, setViewport}) {
  return <ReactMap ref={ref => setMap(ref && ref.getMap())}  style="mapbox://light" {...viewport} onViewportChange={setViewport} />
}

function OtherComponent({map, viewport}) {
  // consume map and viewport
}
```

You proposal (?)

```js
import {MapRefContext, useMapRef} from 'react-map-gl';

function Root() {
  return (
    <MapRefContext.Provider>
       <>
         <Map />
         <OtherComponent />
       </>
    </MapRefContext.Provider>
  );
}

function Map() {
  return <ReactMap style="mapbox://light" />
}

function OtherComponent() {
  const {map, viewport} = useMapRef();
  // consume map and viewport
}
```

If that is the case, `InteractiveMap` should check if `MapContext` is provided by an ancestor, and use the viewport/callbacks from the context if it exists. If we are adding a new API anyways, might as well make the user do as little work as possible?

mayteio:
> If that is the case, InteractiveMap should check if MapContext is provided by an ancestor, and use the viewport/callbacks from the context if it exists.

OK great, we are on the same page. Let's focus on `mapboxRef` here (though the same principle applies for viewport). 

All I am proposing is creating a new component in called `MapProvider` that does exactly what you've described above out of the box. It would;

1. track state of `mapRef`
2. Pass value/setters down via `MapContext.Provider` to `InteractiveMap`

The user can wrap their app in this `MapProvider`.

Then, we update `InteractiveMap` to use `MapContext.Consumer` to check whether there's context above. If so, it uses `setMapRef(map.getMap())` from the parent context to set the mapRef there. This will lift the `map.getMap()` result out of the local component and to app-wide context.

Similarly, a new `useMapRef` hook consumes the context above. If it's there, it provides the mapRef. If not, it's undefined and can warn that the hook is not being used inside `MapProvider`.

*This is exactly the functionality I have described in #904*, we are just hiding the implementation from the user, i.e. 

>  If we are adding a new API anyways, might as well make the user do as little work as possible?

This change requires three things:
1. New `MapProvider`
2. New `useMapRef` hook
3. Refactoring `InteractiveMap` to check for parent `MapContext.Provider`

If you are concerned about mixing functional components with class components then this can be achieved with the latter. The core concept I am introducing here is a *provider that optionally lifts the value of `map.getMap()` from `InteractiveMap` to app-wide context*.

mayteio:
Here's the simplest code example I can put together to describe this:

```ts
import React from 'react';
import { MapContext } from 'react-map-gl';

/** New MapProvider component that uses MapContext */
export const MapProvider = ({children}) => {
  const [map, setMap] = React.useState();
  return <MapContext.Provider value={ { map, setMap, ... } }>
    {children}
  </MapContext.Provider>
}

/** New useMapRef hook */
export const useMapRef = () => {
  const context = React.useContext(MapContext);
  if (!context) throw Error("Not inside <MapProvider />")
  return context;
}

/** Inside existing InteractiveMap */
export default class InteractiveMap extends PureComponent<InteractiveMapProps, State> {
  ...
  render() {
    ...
+    return <MapContext.Consumer>
+      { context => (
        <MapContext.Provider value={ this._interactiveContext } >
            <StaticMap
+            ref={ref => context && context.setMap(ref.getMap())}
            >
              {this.props.children}
            </StaticMap>
          </MapContext.Provider>
+      )}
+    </MapContext.Consumer>
  }
}
```

mayteio:
I think maybe the name `MapProvider` is confusing us here... It's a **new** component exported from the library, that *uses* `MapContext.Provider` or a new `MapRefContext.Provider` in it's render. Otherwise I think we're both on the money 👍

Pessimistress:
See #1233

snickell:
For future searchers, this one is fixed by #1233, as @Pessimistress references above

PR #1116, opened 2020-05-30
[Geolocate] Auto trigger when component loads

anuragkyal:
This PR brings a functionality parallel to https://github.com/mapbox/mapbox-gl-js/issues/5464 which was addressed by exposing a `trigger()` method in https://github.com/mapbox/mapbox-gl-js/pull/6205.

To achieve the same in the react way, added a new property `auto` to the Geolocate component which when set to true, automatically geolocates the user on the map (when permission granted), similar to what happens when the geolocate button is clicked. I feel this is an important feature to bring this library in parity with the other available map libraries.

If I missed an edge case or if there is a better way to get this done, would be happy to make the changes.

### DIFF
--- a/docs/api-reference/geolocate-control.md
+++ b/docs/api-reference/geolocate-control.md
@@ -45,13 +45,13 @@ Callback when Geolocation API position updates. It is called with a Geolocation 
 
 ##### `positionOptions` (Object)
 
-- default: `{enableHighAccuracy: false, timeout: 6000}` 
+- default: `{enableHighAccuracy: false, timeout: 6000}`
 
 A Geolocation API [PositionOptions](https://developer.mozilla.org/en-US/docs/Web/API/PositionOptions) object.
 
 ##### `fitBoundsOptions` (Object)
 
-- default: `{maxZoom: 15}` 
+- default: `{maxZoom: 15}`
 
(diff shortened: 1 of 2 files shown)


anuragkyal:
Looks like the build errors are not related to this change

anuragkyal:
@Pessimistress what's the process of merging this and going for release? is there something that I could help with?
Could really use this feature for our use case :)

Pessimistress:
Sorry I was busy the last few days. I will try publish a new version today.

anuragkyal:
Thanks a lot!

Pessimistress:
@anuragkyal Sorry to go back and forth on this, could it be more useful if we just expose a `trigger` method on the component, so that the user can control when it is fired:

```js
ref.current.trigger();
```

Alternatively, should this component monitor changes in the `auto` prop after it's mounted?

anuragkyal:
@Pessimistress it's a fair use case to let the user control the firing. 

going by the react way, like the idea of monitoring changes to the prop. we can allow one change of the prop from `false` -> `true` while `true` -> `false` remains a no-op always. Think that should suffice?

Pessimistress:
Published 5.2.6: http://visgl.github.io/react-map-gl/docs/api-reference/geolocate-control#auto

anuragkyal:
Thanks a lot @Pessimistress

PR #1123, opened 2020-06-04
Check if map style is loaded before removing source or layer

### LINKED ISSUE #1122: "Error: Style is not done loading" when unmounting MapGL component
When unmounting the MapGL component while the Mapbox style is still loading (e.g. user navigates away shortly after the component is mounted), I am seeing the following issue with react-map-gl 5.2.5 and mapbox-gl 1.10.1 across various browsers / users:

Screenshot 2020-06-04 at 12 10 21

This is the relevant bit in react-map-gl: https://github.com/visgl/react-map-gl/blob/1ed8b3ce8a2918a348d457cb5ea4450504853b80/src/components/source.js#L62

What Mapbox are saying about similar issues (like https://github.com/mapbox/mapbox-gl-js/issues/6947), is essentially: don't modify the map before the map "load" event has fired.

I think this is missing a check whether the style is loaded. Will attempt a fix/PR and do more testing.

jonaswitt:
Avoids "Style is not done loading" error

Resolves #1122 

Although I was unable to reproduce #1122 in our application on every attempt (it depends on the timing of styles being loaded), with this change I could not reproduce it.

### DIFF
--- a/src/components/layer.js
+++ b/src/components/layer.js
@@ -104,7 +104,7 @@ export default class Layer<Props: LayerProps> extends PureComponent<Props> {
     const map = this._map;
     if (map) {
       map.off('styledata', this._updateLayer);
-      if (map.style) {
+      if (map.style && map.style._loaded) {
         map.removeLayer(this.id);
       }
     }
--- a/src/components/source.js
+++ b/src/components/source.js
@@ -58,9 +58,11 @@ export default class Source<Props: SourceProps> extends PureComponent<Props> {
     const map = this._map;
     if (map) {
       map.off('styledata', this._updateSource);
-      if (map.style) {
-        requestAnimationFrame(() => map.removeSource(this.id));
-      }
+      requestAnimationFrame(() => {
+        if (map.style && map.style._loaded) {
+          map.removeSource(this.id);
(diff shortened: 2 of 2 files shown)


Pessimistress:
Thank you! This is published in 5.2.6

RobJJ:
Hi @jonaswitt , Im still running into this error using react-map-gl. Is there any guidance on best practice to follow to avoid this? A difficult to reproduce error but happens often enough for application to be unusable.

For context, I ran into this issue too with mapbox-gl too (same repo, different component), and was able to solve using:
`mapRef.current.on('style.load', () => {})` **instead** of `mapRef.current.on('load', () => {}) `event handler. 
Is there a mapping of this to react-map-gl? Thx!

jonaswitt:
@RobJJ unfortunately I'm not familiar with recent changes to the inner workings of mapbox-gl or react-map-gl, it's quite possible that mapbox-gl add a `style.load` event that is different from the `load` event. Check Mapbox docs to understand which event needs to be received before any styles can be removed. This PR was for a 5 year old version of both libraries and I'm not a maintainer for either of them, sorry.

PR #1180, opened 2020-09-14
Update Source docs

### LINKED ISSUE #1178: Source and Layer components without Mapbox token
ReactMapGL does not render with Source and Layer components if Mapbox token is missing.

This is an OpenStreetMaps example:
https://codesandbox.io/s/react-map-gl-without-token-yqo4v?file=/src/App.js

This is the [example from the docs](https://visgl.github.io/react-map-gl/docs/api-reference/source#source):
https://codesandbox.io/s/react-map-gl-without-token-docs-example-17kqb?file=/src/App.js

Neither renders the map, only `NO_TOKEN_WARNING` error message is displayed.

bkiac:
[Source](https://visgl.github.io/react-map-gl/docs/api-reference/source) documentation is updated with an example of how to use `ReactMapGL` with `Source` and `Layer` components without a Mapbox token.

Closes  #1178

### DIFF
--- a/docs/api-reference/source.md
+++ b/docs/api-reference/source.md
@@ -8,6 +8,12 @@ This component allows apps to create a [map source](https://docs.mapbox.com/mapb
 import * as React from 'react';
 import ReactMapGL, {Source, Layer} from 'react-map-gl';
 
+const EMPTY_STYLE = {
+  version: 8,
+  sources: {},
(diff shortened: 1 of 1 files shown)


Pessimistress (approved):

Pessimistress:
Thank you!

PR #1194, opened 2020-10-07
Fix for layout and paint diffing logic

### LINKED ISSUE #1193: Layer Styles Diff Layout and Paint Issue
There's an issue with diffing layer styles and applying only the new style keys. I have a fix for this and will file a PR.

https://github.com/visgl/react-map-gl/blob/ba454a475ef6513de3689265f6bd4912e0a742f0/src/components/layer.js#L56-L69

For example, say we have a layer with this paint
```json
{
  "paint": {
    "text-halo-width": 8,
    "text-halo-color": "#c52020",
    "text-color": "#000000"
  }
}
```
and we change that to 
```json
{
  "paint": {
    "text-color": "#000000"
  }
}
```
The expected behavior is the text to have no halo with the second config, but the red halo stays.
This _should_ set the key to undefined with 
```javascript
map.setPaintProperty(id, "text-halo-color", undefined);
map.setPaintProperty(id, "text-halo-width", undefined);
```

ctaylor4874:
Fixes #1193

Say we have a layer with this paint
```json
{
  "paint": {
    "text-halo-width": 8,
    "text-halo-color": "#c52020",
    "text-color": "#000000"
  }
}
```
and we change that to 
```json
{
  "paint": {
    "text-color": "#000000"
  }
}
```
The expected behavior is the text to have no halo with the second config, but the red halo stays.
This _should_ set the key to undefined with 
```javascript
map.setPaintProperty(id, "text-halo-color", undefined);
map.setPaintProperty(id, "text-halo-width", undefined);
```

### DIFF
--- a/src/components/layer.js
+++ b/src/components/layer.js
@@ -54,20 +54,28 @@ function diffLayerStyles(map: any, id: string, props: LayerProps, prevProps: Lay
     map.moveLayer(id, beforeId);
   }
   if (layout !== prevProps.layout) {
-    const keys = new Set([...Object.keys(layout), ...Object.keys(prevProps.layout)]);
-    keys.forEach(key => {
+    for (const key in layout) {
       if (!deepEqual(layout[key], prevProps.layout[key])) {
         map.setLayoutProperty(id, key, layout[key]);
       }
-    });
+    }
+    for (const key in prevProps.layout) {
+      if (!layout.hasOwnProperty(key)) {
+        map.setLayoutProperty(id, key, undefined);
+      }
+    }
   }
   if (paint !== prevProps.paint) {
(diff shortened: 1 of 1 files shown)


Pessimistress:
Thank you!

I wonder what's the perf impact of merging two lists + forEach (creates new function)? I don't have any benchmarks but this should be faster on principle:

```js
for (const key in layout) {
  if (!deepEqual(layout[key], prevProps.layout[key])) {
    // set
  }
}

for (const key in prevProps.layout) {
  if (!layout.hasOwnProperty(key)) {
    // set
  }
}
```

ctaylor4874:
> ```js
> !layout.hasOwnProperty(key)
> ```

I'll make the change and push it up. Thank you!

Pessimistress (approved):

PR #1235, opened 2020-11-25
Migrate CI to Github Actions

Pessimistress:


.github/workflows/test.yml
```diff
@@ -0,0 +1,28 @@
+name: test
+
+on: [push]
```
kylebarron, line 3: Note that this won't trigger for pull releases that initiate from external repositories. You need `pull_request_target` for that. See https://github.com/visgl/loaders.gl/pull/1016 and https://github.com/visgl/loaders.gl/pull/1018

kylebarron (approved):

PR #1249, opened 2020-12-09
Upgrade gatsby theme for website

Pessimistress:


ibgreen (approved):
Breaking changes in the config in minor releases are a bit hard on users.

kylebarron (approved):

Issue #1267, opened 2020-12-18
Layer component gives Invalid prop warning when `type` is set to `sky`


elliotwutingfeng:
When declaring a Layer component with the new "sky" type introduced in Mapboxgl v2
```
<Layer id="sky" type="sky" paint={mapStyle.skybox} />
```
the following warning is given.
```
Warning: Failed prop type: Invalid prop `type` of value `sky` supplied to `Layer`,
 expected one of ["fill","line","symbol","circle","fill-extrusion","raster","background","heatmap","hillshade"]
```
However, the sky still renders correctly.

Pessimistress:
Thanks for reporting. Would you open a PR to fix this?

Pessimistress:
Published 6.0.2

PR #1270, opened 2020-12-19
Add support for 'sky' layer type to Layer component

elliotwutingfeng:
Addresses #1267. Removes incorrect `Invalid prop type` warning when Layer component prop type is set to `sky`.


Pessimistress (approved):
Thank you!

kylebarron (approved):

PR #1271, opened 2020-12-21
Refactor StaticMap and InteractiveMap into functional components

Pessimistress:


src/components/interactive-map.js
```diff
@@ -125,153 +124,178 @@ const defaultProps = Object.assign(
...
+    event = normalizeEvent.call(this, event);
+    // backward compatibility: v3 `onClick` interface
+    event.features = getFeatures.call(this, event.point);
+    callbacks.forEach(cb => cb(event));
+  }
+}
+/* End of event handers */
+
+const InteractiveMap = forwardRef((props, ref) => {
+  const [controller] = useState(() => props.controller || new MapController());
```
xintongxia, line 270: Maybe use `useRef` here?

src/components/interactive-map.js
```diff
@@ -125,153 +124,178 @@ const defaultProps = Object.assign(
...
+  // Event handlers are registered once but need access to the latest props
+  // This is an anti-pattern, though it maintains a persistent reference to the latest props/state of this component
+  const _thisRef = useRef({
+    width: 0,
+    height: 0,
+    state: {
+      isHovering: false,
+      isDragging: false
+    }
+  });
```
macrigiuseppe, line 288: thank you for sharing this, i didn't know about this pattern. Based on this [reading](https://www.emgoto.com/storing-values-with-useref/) it looks like storing values in useRef doesn't trigger a re-render

src/components/interactive-map.js
```diff
@@ -125,153 +124,178 @@ const defaultProps = Object.assign(
...
+  thisRef.props = props;
+  thisRef.map = staticMapRef.current && staticMapRef.current.getMap();
+  thisRef.setState = newState => {
+    const state = Object.assign(thisRef.state, newState);
+    eventCanvasRef.current.style.cursor = props.getCursor(state);
   };
 
-  _onViewportChange = (viewState, interactionState, oldViewState) => {
-    const {onViewStateChange, onViewportChange} = this.props;
+  const handleViewportChange = (viewState, interactionState, oldViewState) => {
```
macrigiuseppe, line 297: useCallback

src/components/interactive-map.js
```diff
@@ -125,153 +124,178 @@ const defaultProps = Object.assign(
...
+      isDragging: false
+    }
+  });
+  const thisRef = _thisRef.current;
+  thisRef.props = props;
+  thisRef.map = staticMapRef.current && staticMapRef.current.getMap();
+  thisRef.setState = newState => {
+    const state = Object.assign(thisRef.state, newState);
+    eventCanvasRef.current.style.cursor = props.getCursor(state);
   };
```
macrigiuseppe, line 295: can you init these values when you create the ref?

src/components/interactive-map.js
```diff
@@ -280,175 +304,92 @@ export default class InteractiveMap extends PureComponent {
...
+    return () => {
+      eventManager.destroy();
+    };
+  }, []);
+
+  updateControllerOpts();
+
+  const {width, height, style, getCursor} = props;
+
+  const eventCanvasStyle = Object.assign({position: 'relative'}, style, {
```
macrigiuseppe, line 369: useMemo to memoize values while across re-renderings

macrigiuseppe:
@Pessimistress you can take advantage of useCallback and useMemo to speed up performances

src/components/interactive-map.js
Pessimistress: What's the difference? We just need a persistent pointer that never updates

src/components/interactive-map.js
Pessimistress: That is correct. The problem I'm trying to solve here is that event listeners are registered once but need access to the latest state in the handlers. We want to avoid rerender when this object is mutated. Its only purpose is to provide pointers to the current props/state when events are fired.

src/components/interactive-map.js
Pessimistress: `map` only exists after mount. `setState` depends on `props` so it has to be updated on each render anyways.

src/components/interactive-map.js
Pessimistress: `handleViewportChange` is not passed to any React component. There is no optimization based on the equality of this callback.

src/components/interactive-map.js
macrigiuseppe: this looks like a simple callback, why don't you use: useCallback?

src/components/interactive-map.js
```diff
@@ -125,153 +124,178 @@ const defaultProps = Object.assign(
...
+    // backward compatibility: v3 `onClick` interface
+    event.features = getFeatures.call(this, event.point);
+    callbacks.forEach(cb => cb(event));
+  }
+}
+/* End of event handers */
+
+const InteractiveMap = forwardRef((props, ref) => {
+  const [controller] = useState(() => props.controller || new MapController());
+  const [eventManager] = useState(() => new EventManager(null, {touchAction: props.touchAction}));
```
macrigiuseppe, line 271: same for this one

src/components/interactive-map.js
Pessimistress: The function passed to `useState` is the initializer.

macrigiuseppe (approved):

Issue #1129, opened 2020-06-11
Custom attribution control


daniel-j-h:
Right now we have two ways to change the attribution
1. passing a [boolean `attributionControl`](https://github.com/visgl/react-map-gl/blob/28f0b6e01afb52e69be5aa999156b9703b0620aa/src/mapbox/mapbox.js#L41)
2. passing [`customAttributions` in `mapOptions`](https://github.com/visgl/react-map-gl/blob/28f0b6e01afb52e69be5aa999156b9703b0620aa/src/mapbox/mapbox.js#L45)

If `attributionControl` is false, the map attribution will not be shown at all. If it's true and there is a `customAttributions` it will get **added** to the default map attributions.

This is effectively mapping to https://docs.mapbox.com/mapbox-gl-js/api/map/#map-parameters

What's missing is a way to completely control the map attributions, e.g.
1. not just add to the default attributions but customize it completely
2. make the attributions compact by default

Which maps to https://docs.mapbox.com/mapbox-gl-js/api/#attributioncontrol

The use cases this would allow is e.g. localizing the "Improve this map" attribution, and having full control over the attribution content and styling while at the same time adhering to the Mapbox attribution guidelines.

Related
- https://github.com/visgl/react-map-gl/issues/862
- https://docs.mapbox.com/help/how-mapbox-works/attribution/

makeupsomething:
I can work on this issue over the weekend as we also need this function

PR #1276, opened 2020-12-28
Feature/allow user to set compact attribution

makeupsomething:
Address part of #1129 to allow the user to set the map attribution to the compact size.

I did not update to include the request to `not just add to the default attributions but customize it completely` as this does not match the API behaviour from mapbox js.

### DIFF
--- a/src/mapbox/mapbox.js
+++ b/src/mapbox/mapbox.js
@@ -273,7 +273,15 @@ export default class Mapbox {
       if (props.transformRequest) {
         mapOptions.transformRequest = props.transformRequest;
       }
-      this._map = new this.mapboxgl.Map(Object.assign({}, mapOptions, props.mapOptions));
+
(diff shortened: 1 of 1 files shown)


Pessimistress:
`mapOptions` directly maps to the options passed to the [Map constructor](https://docs.mapbox.com/mapbox-gl-js/api/map/). We should not add our own API to this object.

I believe the proper solution to this is to add an `AttributionControl` react component.

makeupsomething:
Thank you for the feedback!

This makes sense, I'l work on adding an `AttributionControl` component.

Pessimistress:
@makeupsomething This has been inactive for a while. We are aiming to make the 5.3/6.1 release in the next couple of weeks. Taking over with https://github.com/visgl/react-map-gl/pull/1295

makeupsomething:
@Pessimistress Sorry I got busy with other work. Thank you for taking it over 🙏

Issue #1171, opened 2020-08-31
Got error about updating during existing state transition


Jamie5:
In InteractiveMap, the onLoad event seems to call `setState`, which is usually fine, except when the load event is fired during a render. So far this reproduces reliable within a synchronous browser-based unit test environment, but it is probably nondeterministic enough to not be easy to reproduce in actual usage. Nevertheless, looking at the error and stack trace (which are included below), there doesn't seem to be anything which prevents this from happening in the wild.

Throwing a `setTimeout` around the onLoad handler works to get rid of the error, but is probably not the best solution.

This seems similar to #872 in terms of what's going on (a render leads to an event which then leads to a setState call).

In this case, it seems during `render` of `StaticMap` we call `_renderOverlays` which leads to setting the size, and hence calling mapbox-gl redraw, which notices that the map became loaded but we haven't fired the load event, and so fires it. This calls InteractiveMap onLoad which calls setState, hence leading to the problem.

Error message

```
Warning: Cannot update during an existing state transition (such as within `render`). Render methods should be a pure function of props and state.
    in AutoSizer (created by StaticMap)
    in div (created by StaticMap)
    in StaticMap (created by InteractiveMap)
    in div (created by InteractiveMap)
    in InteractiveMap (at <censored>)
    <censored>
```

Stack trace
```
method context.js:265
    spiedMethod jasmineSetup.js:86
    React 6
        printWarning
        error
        warnAboutRenderPhaseUpdatesInDEV
        scheduleUpdateOnFiber
        enqueueSetState
        setState
    InteractiveMap interactive-map.js:200
    fire mapbox-gl.js:1499
    _render mapbox-gl.js:42546
    redraw mapbox.js:156
    _update mapbox.js:318
    setProps mapbox.js:133
    _updateMapProps static-map.js:187
    _updateMapSize static-map.js:165
    _renderOverlays static-map.js:227
    render index.esm.js:387
    React 15
        finishClassComponent
        updateClassComponent
        beginWork
        beginWork$1
        performUnitOfWork
        workLoopSync
        performSyncWorkOnRoot
        flushSyncCallbackQueueImpl
        unstable_runWithPriority
        runWithPriority$1
        flushSyncCallbackQueueImpl
        flushSyncCallbackQueue
        unbatchedUpdates
        legacyRenderSubtreeIntoContainer
        render
    renderComponent setup.js:77
    <anonymous> Foo-test.tsx:218
    attempt jasmine.js:6365
[...]

PR #1277, opened 2020-12-29
Fix React error when transition is triggered

Pessimistress:
This fixes the React error when a viewport transition is triggered. #1171

The root cause is that we call `controller.setOptions` during render, in which `onViewportChange` may be called. React then logs an error because `onViewportChange` usually performs a state update.

This PR defers any viewport change callback till after render.

### DIFF
--- a/src/components/interactive-map.js
+++ b/src/components/interactive-map.js
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import {useState, useRef, useMemo, useEffect, forwardRef} from 'react';
+import {useState, useRef, useMemo, useEffect, useLayoutEffect, forwardRef} from 'react';
 import * as PropTypes from 'prop-types';
 
 import StaticMap from './static-map';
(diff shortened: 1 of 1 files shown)


xintongxia (approved):

src/components/interactive-map.js
```diff
@@ -362,6 +372,13 @@ const InteractiveMap = forwardRef((props, ref) => {
     };
   }, []);
 
+  useLayoutEffect(() => {
```
macrigiuseppe, line 375: based on [this](https://reactjs.org/docs/hooks-reference.html#uselayouteffect), `useLayoutEffect` offers a sync side effect and it may block the painting process. Is this the expected behavior?

Pessimistress: Yes. The reason the controller needs to call `onViewportChange` is to intercept any viewport change before it's rendered. An example:

- Current viewport - `longitude: 0, latitude: 0, zoom: 10`
- User set props - `longitude: 0, latitude: 0, zoom: 6, transitionInterpolator: new LinearInterpolator(), transitionDuration: 1000`
- Controller triggers transition, calls `onViewportChange` - `longitude: 0, latitude: 0, zoom: 9.9`
- Updates map - `longitude: 0, latitude: 0, zoom: 9.9`

Note here we don't want `zoom: 6` to render before `zoom: 9.9`.

PR #1278, opened 2020-12-29
Clean up and document MapContext and useMapControl APIs

Pessimistress:
### Change List:
- Promote `MapContext` to official API (`_MapContext` export is kept for backward compatibility), add documentation
- Update "custom components" doc to use `MapContext` and `_useMapControl`
- Update component docs to cover missing interactive props
- Simplify `useMapControl` API to take only 1 argument
- Tested sample code in new docs

### Breaking Change
- Removed `isDragging` from `MapContext`. `InteractiveMap` no longer updates with this state to improve perf. This state is only used by a custom overlay example, and the same effect can be achieved by watching the interactive state in the app.


snickell:
Appreciate the clear docs of MapContext, this has already saved me time!

docs/advanced/custom-components.md
```diff
@@ -1,84 +1,92 @@
...
-      <div ref={this._containerRef}
-        style={markerStyle} >
-        ({longitude}, {latitude})
-      </div>
-    );
-  }
+import MapGL, {MapContext} from 'react-map-gl';
+
+function CustomMarker(props) {
+  const context = React.useContext(MapContext);
```
macrigiuseppe, line 14: we could do a further improvement, maybe we can export something like the following:
```
import {useContext} from 'react';
import {MapContext} from '../context';

const useMapContext = () => useContext(FeatureFlagsContext);

export default useMapContext;
```

by doing the following, developers don't need to use `React.useContext(MapContext)` but they can directly use `useMapContext`

docs/advanced/custom-components.md
```diff
@@ -1,84 +1,92 @@
...
 
-- `viewport` {WebMercatorViewport} - the current viewport
-- `map` {mapboxgl.Map} - the Mapbox map instance
-- `eventManager` {EventManager} - the event manager. Only available if using `InteractiveMap`.
-- `isDragging` {Bool} - whether the map is being dragged. Only available if using `InteractiveMap`.
+`useMapControl` is an experimental API that makes it easier to create controls with event handling.
 
+```js
+import * as React from 'react';
+import MapGL, {_useMapControl as useMapControl} from 'react-map-gl';
```
macrigiuseppe, line 49: i noticed in here: https://github.com/visgl/react-map-gl/pull/1278/files#diff-b8e38018e9d59b4598ff7ff7f767e98641d1148372578b5323fbc916e02f67c9R22
we import directly `useMapControl`. Why do we import `_ useMapControl` in this statement?

macrigiuseppe (approved):
@Pessimistress it looks good, just few comments but nothing bloking

docs/advanced/custom-components.md
Pessimistress: Which file are you referring to?

docs/advanced/custom-components.md
Pessimistress: Is there a significant difference? I personally like `useContext` because the user will be able to refer to React's documentation for the behaviors of context. If we make a custom hook then that's all hidden under the hood.

PR #1295, opened 2021-01-13
Add AttributionControl

Pessimistress:
#1129

### DIFF
--- a/src/components/attribution-control.js
+++ b/src/components/attribution-control.js
@@ -1,0 +1,94 @@
+import * as React from 'react';
+import {useEffect, useCallback, useState, useRef} from 'react';
+import * as PropTypes from 'prop-types';
+import mapboxgl from '../utils/mapboxgl';
+import useMapControl, {mapControlDefaultProps, mapControlPropTypes} from './use-map-control';
+
+const propTypes = Object.assign({}, mapControlPropTypes, {
+  toggleLabel: PropTypes.string,
+  className: PropTypes.string,
+  style: PropTypes.object,
+  compact: PropTypes.bool,
+  customAttribution: PropTypes.oneOfType([PropTypes.string, PropTypes.arrayOf(PropTypes.string)])
+});
+
+const defaultProps = Object.assign({}, mapControlDefaultProps, {
+  className: '',
+  toggleLabel: 'Toggle Attribution'
+});
+
+function setupAttributioncontrol(opts, map, container, attributionContainer) {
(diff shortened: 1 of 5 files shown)


src/components/attribution-control.js
```diff
@@ -0,0 +1,94 @@
...
+
+const propTypes = Object.assign({}, mapControlPropTypes, {
+  toggleLabel: PropTypes.string,
+  className: PropTypes.string,
+  style: PropTypes.object,
+  compact: PropTypes.bool,
+  customAttribution: PropTypes.oneOfType([PropTypes.string, PropTypes.arrayOf(PropTypes.string)])
+});
+
+const defaultProps = Object.assign({}, mapControlDefaultProps, {
```
macrigiuseppe, line 15: how do you feel about destructuring here:
```
const defaultProps ={...mapControlDefaultProps, {
  className: '',
  toggleLabel: 'Toggle Attribution'
}};
```

src/components/attribution-control.js
```diff
@@ -0,0 +1,94 @@
...
+  useEffect(() => {
+    if (!compact && showCompact) {
+      setShowCompact(false);
+    }
+  }, [compact]);
+
+  const toggleAttribution = useCallback(() => setShowCompact(value => !value), []);
+
+  return (
+    <div style={{position: 'absolute', ...props.style}} className={props.className}>
```
macrigiuseppe, line 72: take advantage of useMemo the same way you did in other PRs
```
useMemo(() => ({position: 'absolute', ...(props.style || {})}), [props.style])
```
I would make sure prop.style is not null

src/components/attribution-control.js
```diff
@@ -0,0 +1,94 @@
...
+      <div
+        ref={containerRef}
+        aria-pressed={showCompact}
+        className={`mapboxgl-ctrl mapboxgl-ctrl-attrib ${compact ? 'mapboxgl-compact' : ''} ${
```
macrigiuseppe, line 76: if we are planning on having several instances of class name manipulation we should think about using something like `classnames` to better handle these cases

macrigiuseppe (approved):

src/components/attribution-control.js
Pessimistress: I believe the spread operator supports null/undefined (similar to `Object.assign`)

Pessimistress: Will address in #1297

Issue #835, opened 2019-07-01
All components support inline styles


xintongxia:

Pessimistress:
Addressed in v5.3/v6.1/

PR #1297, opened 2021-01-13
Align control component APIs

Pessimistress:
#835

- Add `className` props to all controls
- Add `style` prop to all controls except Marker and Popup
- Make all control components pure
- Make sure all props are documented

### DIFF
--- a/examples/controls/src/app.js
+++ b/examples/controls/src/app.js
@@ -18,28 +18,24 @@ import CITIES from '../../.data/cities.json';
 const TOKEN = ''; // Set your mapbox token here
 
 const geolocateStyle = {
-  position: 'absolute',
   top: 0,
   left: 0,
   padding: '10px'
 };
 
 const fullscreenControlStyle = {
-  position: 'absolute',
   top: 36,
   left: 0,
   padding: '10px'
 };
 
 const navStyle = {
-  position: 'absolute',
   top: 72,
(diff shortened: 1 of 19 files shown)


examples/controls/src/app.js
```diff
@@ -18,28 +18,24 @@ import CITIES from '../../.data/cities.json';
 const TOKEN = ''; // Set your mapbox token here
 
 const geolocateStyle = {
-  position: 'absolute',
```
macrigiuseppe, line 21: was position absolute not useful in these cases?

macrigiuseppe (approved):

src/components/attribution-control.js
```diff
@@ -67,9 +67,10 @@ function AttributionControl(props) {
   }, [compact]);
 
   const toggleAttribution = useCallback(() => setShowCompact(value => !value), []);
+  const style = useMemo(() => ({position: 'absolute', ...props.style}), [props.style]);
```
macrigiuseppe, line 70: make sure props.style is not null
```
const style = useMemo(() => ({position: 'absolute', ...(props.style || {})}), [props.style]);
```

examples/controls/src/app.js
Pessimistress: We already merge user-provided style with `{position: 'absolute'}`

Issue #1306, opened 2021-01-27
Typescript - Source props is not fully defined


kentongray:
I noticed after upgrading to React Map GL I could no longer set the source props on the layer for cluster (like in the example). 

Also in general the types are not exported making it difficult to work with (with the definitely typed I'd reference the source props and things like PointerEvent types)

Here is the source from the definitely typed repo which is what I was using with version 5 and it worked well. I'd recommend basing the types off of that.

Hopefully I'm not doing something dumb, thanks!

# DefinitelyTyped Definition

```
export interface SourceProps {
    id?: string;
    type: string;
    url?: string;
    tiles?: string[];
    tileSize?: number;
    bounds?: number[];
    scheme?: 'xyz' | 'tms';
    minzoom?: number;
    maxzoom?: number;
    attribution?: string;
    encoding?: 'terrarium' | 'mapbox';
    data?: GeoJSON.Feature<GeoJSON.Geometry> | GeoJSON.FeatureCollection<GeoJSON.Geometry> | string;
    buffer?: number;
    tolerance?: number;
    cluster?: boolean;
    clusterRadius?: number;
    clusterProperties?: object;
    clusterMaxZoom?: number;
    lineMetrics?: boolean;
    generateId?: boolean;
    coordinates?: number[][];
    urls?: string[];
    children?: React.ReactNode;
}
```

# Source.d.ts in Repo

```
type SourceProps = {
  id?: string,
  type: string,
  children?: any,
  data?: any,
  coordinates?: any,
  url?: any,
  tiles?: any
};
```

Pessimistress:
@kentongray Thanks for the report. We don't use TypeScript in all the examples so it's hard to catch such issues. Would you open a PR to fix the `SourceProps` typing and maybe suggest which ones we should export?

ediblecode:
I've come across the same problem today with clusters specifically:

![image](https://user-images.githubusercontent.com/893676/106195515-26e6ec00-61a8-11eb-9f38-b3524d102c61.png)

In terms of exporting props, I think probably the props for each component makes sense. Using the [clusters example](https://github.com/visgl/react-map-gl/tree/6.1-release/examples/clusters) I get the following type error:

> Types of property 'type' are incompatible. Type 'string' is not assignable to type '"symbol" | "circle" | "line" | "raster"...

For:

```jsx
const clusterLayer = {
	id: "clusters",
	type: "circle",
        ...
}
...
<Layer {...clusterLayer} />
```

[...]

PR #1308, opened 2021-01-28
Impremove Type Definitions

kentongray:
Fix for #1306 adding typescript definitions from DefinitelyTyped repo and adding some for exporting so they can be used by apps.

### DIFF
--- a/src/components/layer.d.ts
+++ b/src/components/layer.d.ts
@@ -3,7 +3,7 @@ import {PureComponent, Ref} from "react";
 
 export interface LayerProps {
   id?: string;
-  type:  'fill' | 'line' | 'symbol' | 'circle' | 'fill-extrusion' | 'raster' | 'background' | 'heatmap' | 'hillshade';
+  type:  'fill' | 'line' | 'symbol' | 'circle' | 'fill-extrusion' | 'raster' | 'background' | 'heatmap' | 'hillshade' | 'sky';
   source?: string;
   beforeId?: string;
   layout?: MapboxGL.AnyLayout;
--- a/src/components/layer.d.ts
+++ b/src/components/layer.d.ts
@@ -1,15 +1,27 @@
+import * as MapboxGL from "mapbox-gl";
+import {PureComponent} from "react";
 
-type LayerProps = {
-  id?: string,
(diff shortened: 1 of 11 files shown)


kentongray:
FYI I couldn't figure out a way to publish this as a contributor to where I could mess with it... When I tried to change the package name so I could publish locally it broke the unit tests.. So I "hope" this works but can't really figure out a way to test. If you have any ideas of how I can test locally I'm all ears

src/components/layer.d.ts
```diff
@@ -1,15 +1,27 @@
...
+    | MapboxGL.RasterPaint
+    | MapboxGL.CirclePaint
+    | MapboxGL.HeatmapPaint
+    | MapboxGL.HillshadePaint;
+  filter?: any[];
+  minzoom?: number;
+  maxzoom?: number;
+}
+
+export default class Layer extends PureComponent<LayerProps> {}
```
Pessimistress, line 25: Does it matter that this does not match the implementation (functional component)?

Pessimistress (approved):

src/components/layer.d.ts
kentongray: I did this so it would have the ref prop, I don't think it truly matters but alternatively we could define the ref prop to accomplish the same thing

kentongray: I kinda like that approach more so I switched it

src/components/layer.d.ts
```diff
@@ -1,15 +1,28 @@
...
-  beforeId?: string,
-  layout?: any,
-  paint?: any,
-  filter?: Array<any>,
-  minzoom?: number,
-  maxzoom?: number
-};
+export interface LayerProps {
+  id?: string;
+  type:  'fill' | 'line' | 'symbol' | 'circle' | 'fill-extrusion' | 'raster' | 'background' | 'heatmap' | 'hillshade';
```
Pessimistress, line 6: Looks like we are missing `'sky'` here

kentongray:
took care of your comments, I think I'm probably going to run into more so I'll just continue to submit PRs as I run into other areas but I'll stop pushing to this one

PR #1365, opened 2021-02-23
Import mapbox-gl such that it is transpilable

arindam1993:
Following up from the transpilation issues described in https://github.com/mapbox/mapbox-gl-js/issues/10173

With this PR I'm attempting to switch the import for mapbox-gl-js, and use the [`mapboxgl.workerClass`](https://docs.mapbox.com/mapbox-gl-js/api/#forcing-transpilation-with-babel) hook so that the build system can recognize the separate bundles for main-thread and worker code.
This should allow arbitrary transpilation with Babel because it makes it aware of the multi-entrypoint nature of the library.

### DIFF
--- a/package.json
+++ b/package.json
@@ -49,7 +49,7 @@
     "@babel/runtime": "^7.0.0",
     "@types/geojson": "^7946.0.7",
     "@types/mapbox-gl": "^2.0.3",
-    "mapbox-gl": "^2.0.1",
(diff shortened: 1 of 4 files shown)


Pessimistress:
react-map-gl does not ship its own pre-bundled version. All files are simply transpiled by babel. Correct me if I'm wrong, but this change means that every user will have to add a worker-loader config to their bundler to be able to use this library.

PR #1369, opened 2021-02-24
Improve typescript bindings

Jamie5:
- Make MapEvent.point/lngLat 2-arrays
- From MjolnirEvent remove deltaX/deltaY/scale/rotation/metaKey which seem like they might not exist, or at least are not in the public API documentation
- From MjolnirEvent make delta/pointerType/rightButton optional, as the API documentation states they are only present for some events
- From MjolnirEvent add stopImmediatePropagation/leftButton/middleButton/key which are supposed to exist per docs
- From MjolnirEvent make typings for stopPropagation/preventDefault/srcEvent more specific

### DIFF
--- a/src/utils/map-controller.d.ts
+++ b/src/utils/map-controller.d.ts
@@ -5,18 +5,22 @@ export type MjolnirEvent = {
   type: string,
   center: {x: number, y: number},
   offsetCenter: {x: number, y: number},
-  deltaX: number,
-  deltaY: number,
-  delta: number,
-  scale: number,
-  rotation: number,
-  pointerType: string,
-  metaKey: boolean,
-  rightButton: boolean,
-  stopPropagation: Function,
-  preventDefault: Function,
+  deltaX?: number,
+  deltaY?: number,
+  delta?: number,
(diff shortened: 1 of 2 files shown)


src/utils/map-controller.d.ts
```diff
@@ -5,18 +5,17 @@ export type MjolnirEvent = {
...
-  deltaY: number,
-  delta: number,
-  scale: number,
-  rotation: number,
-  pointerType: string,
-  metaKey: boolean,
-  rightButton: boolean,
-  stopPropagation: Function,
-  preventDefault: Function,
+  delta?: number,
```
Pessimistress, line 8: Why are some of the fields deleted? They are accessed in map-controller.js:
- deltaX
- deltaY
- scale
- rotation

Jamie5: Hmm they aren't documented as existing so I wasn't sure if they existed, but given we use them then they certainly do exist. So restored them (but made them optional, because they only exist for some events)

Pessimistress (approved):
Thank you!

Issue #1393, opened 2021-03-08
"There is no source with this ID" thrown when Source is removed quickly


fenech:
#### Description

Rendering a `<Source>` and removing it quickly causes an exception to be thrown. This is because `map.removeSource(id)` is called before the source has been created.

A possible easy fix would be to just add a check `map.getSource(id)` to [the existing check](https://github.com/visgl/react-map-gl/blob/bdddc49805f2bc51916864bda21bdbf7758ba208/src/components/source.js#L117).

Happy to open a PR with that if you think it's a good idea!

#### Repro Steps

```js
import { useEffect, useState } from "react";
import InteractiveMap, { Source } from "react-map-gl";

export default function App() {
  const [showSource, setShowSource] = useState(true);
  useEffect(() => {
    requestAnimationFrame(() => setShowSource(false));
  }, []);

  return (
    <InteractiveMap
      height={100}
      width={100}
      mapboxApiAccessToken="API TOKEN"
    >
      {showSource && <Source type="geojson" data={{}}/>}
    </InteractiveMap>
  );
}
```

#### Environment (please complete the following information):

- Library Version: react-map-gl 6.1.10
- Mapbox Version: mapbox-gl 2.1.1
- React Version: react 17.0.1
- Browser Version: Firefox 86.0
- OS: GNU/Linux Fedora 33

### There is no source with this ID

removeSource
https://tg2bo.csb.app/node_modules/mapbox-gl/dist/mapbox-gl.js:17999:34
removeSource
https://tg2bo.csb.app/node_modules/mapbox-gl/dist/mapbox-gl.js:23614:30
Source/</</<
https://tg2bo.csb.app/node_modules/react-map-gl/dist/esm/components/source.js:125:17

VictorVelarde:
Hi @fenech,

Thx for your report. I think adding the check is a nice fix, would you like to open a PR?

VictorVelarde:
This should be fixed with latest releases: 6.1.11 / 5.3.11. @fenech please confirm everything is fine on your side, and we'll close this issue

fenech:
I can confirm that it's fixed in 6.1.11, thanks!

PR #1396, opened 2021-03-10
Ensure source exists before trying to remove it

fenech:
Addresses #1393

### DIFF
--- a/src/components/source.js
+++ b/src/components/source.js
@@ -114,14 +114,14 @@ function Source(props) {
         // Removing source before dependent layers will throw error
         // TODO - find a more robust solution
         requestAnimationFrame(() => {
-          if (map.style && map.style._loaded) {
+          if (map.style && map.style._loaded && map.getSource(id)) {
             map.removeSource(id);
           }
         });
       };
     }
     return undefined;
-  }, [map]);
+  }, [map, id]);
 
   let source = map && map.style && map.getSource(id);
   if (source) {


src/components/source.js
```diff
@@ -114,14 +114,14 @@ function Source(props) {
...
-          if (map.style && map.style._loaded) {
+          if (map.style && map.style._loaded && map.getSource(id)) {
             map.removeSource(id);
           }
         });
       };
     }
     return undefined;
-  }, [map]);
+  }, [map, id]);
```
Pessimistress, line 124: `id` is not supposed to change.

fenech: Yeah, it makes sense that it shouldn't change, but it is still a dependency of the effect, so I included it to satisfy the ["exhaustive deps" recommendation](https://github.com/facebook/create-react-app/issues/6880#issuecomment-485912528)

If the component is used correctly, then adding this change will have no effect, and if the `id` does change, then this means that it is more likely that both sources are cleaned up properly.

Of course, if you prefer, I can update the PR to remove it.

Pessimistress: `id` is a constant ensured by the `useMemo(..., [])` call above, so it doesn't depend on whether the user is using this component correctly.

Pessimistress (approved):

src/components/source.js
fenech: Ah, my mistake! I should've zoomed out a bit :smile: I can update the PR no problem, up to you.

fenech: (just an aside, `useMemo` [doesn't guarantee](https://reactjs.org/docs/hooks-reference.html#usememo) that it will be a constant)
> You may rely on useMemo as a performance optimization, not as a semantic guarantee.

VictorVelarde:
Moving your contribution to https://github.com/visgl/react-map-gl/pull/1417, so we can merge it, thx!

PR #1398, opened 2021-03-12
Add promoteId as a source prop

amber-chen:
https://docs.mapbox.com/mapbox-gl-js/style-spec/sources/#geojson-promoteId

This field is already supported in mapbox for a geojson source.


kylebarron:
I don't think changing only the type declaration file will change the code as well. You'd probably need to edit the `.js` file as well.

Pessimistress (approved):

amber-chen:
🚀

PR #1441, opened 2021-04-21
Add click propagation

makeupsomething:
This PR fixes https://github.com/visgl/react-map-gl/issues/1433

It looks like `stopPropagation` was removed in this commit https://github.com/visgl/react-map-gl/commit/aae80369ed106f5c66aa4e657aa3fcac119336c9#diff-1ce43bcf84ec81f15ea313756c9204398efe0003d8a6dbecb69b7ebe7157b09c

I added it back in and it seems to no longer cause the problem outlined in the issue.

### DIFF
--- a/src/components/popup.js
+++ b/src/components/popup.js
@@ -123,6 +123,10 @@ function getContainerStyle(props, viewport, el, [x, y, z], positionType) {
 }
 
 function onClick(evt, {props, context}) {
+  if (props.captureClick) {
+    context.eventManager.once('click', e => e.stopPropagation(), evt.target);
+  }
+
   if (props.closeOnClick || evt.target.className === 'mapboxgl-popup-close-button') {
     props.onClose();
 


Pessimistress:
The handling of propagation is moved to the shared hook: https://github.com/visgl/react-map-gl/blob/850ae97d54671ea43bad11741cb05bde24caa362/src/components/use-map-control.js#L63 so something else is not working.

It doesn't make sense to register and block ANOTHER click event after a click event is already fired.

makeupsomething:
@Pessimistress 
Thank you for checking this!
I understand, yeah thats not the problem here.

After looking at it a bit more the issue seems to be that when the conditional element inside the popup is being clicked the propagation is never triggered. It looks like it is because the containerRef is determined on mount and never updated to include any conditional elements. I'l look into it a bit more later today.

PR #1477, opened 2021-05-28
Prevent attribution button from submitting forms

rajadain:
If the Map is embedded within a form, clicking the attribution button submits it by default.
By explicitly declaring it as a "button" and not "submit", we ensure that doesn't happen.

For details see https://stackoverflow.com/a/10836076/6995854


rajadain:
Thank you!

PR #1505, opened 2021-06-21
Fix stale `onLoad` and `onError` event handlers

khmm12:
The example:

```jsx
const [viewport, setViewport] = useState(DefaultViewport)

const [points, setPoints] = useState([])

useEffect(() => {
  console.log('effect')  
  setPoints(Points) /* the constant defined somewhere */
}, [])

const handleLoad = () => {
  console.log('[load] points.length = %d', points.length)  
  setViewport(panTo(points))
}

const handleError = (e) => {}

return (
  <MapGL
    {...viewport}
    onViewportChange={setViewport}
    onLoad={handleLoad}
    onError={handleError}
  />
)
```

Console (before the fix):
```
effect
[load] points = 0
```

Console (after the fix, expected):

```
effect
[load] points = 1
```

Also it fixes zombie `onError` handler.

### DIFF
--- a/src/mapbox/mapbox.js
+++ b/src/mapbox/mapbox.js
@@ -204,6 +204,11 @@ export default class Mapbox {
     });
   };
 
+  _handleError = event => {
+    // @ts-ignore
+    this.props.onError(event);
+  };
+
   _reuse(props) {
     this._map = Mapbox.savedMap;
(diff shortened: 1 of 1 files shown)


Pessimistress (approved):
Thank you!

PR #1555, opened 2021-08-24
Fix to not save to savedMap when reuseMaps is false

DuGlaser:
Hi! Thanks for the great library.
I have been using this library and noticed that when the Map is unmounted, the worker remains even if reuseMaps is false.
In this PR, I changed it so that when the Map is unmounted, if reuseMaps is false, it will not be saved in savedMap.

### DIFF
--- a/src/mapbox/mapbox.js
+++ b/src/mapbox/mapbox.js
@@ -161,8 +161,8 @@ export default class Mapbox {
   width = 0;
   height = 0;
 
-  finalize(props) {
-    this._destroy(props);
+  finalize() {
+    this._destroy();
     return this;
   }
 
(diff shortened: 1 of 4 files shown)


src/mapbox/mapbox.js
```diff
@@ -279,12 +279,12 @@ export default class Mapbox {
...
   }
 
-  _destroy() {
+  _destroy(props) {
     if (!this._map) {
       return;
     }
 
-    if (!Mapbox.savedMap) {
+    if (props.reuseMaps && !Mapbox.savedMap) {
```
Pessimistress, line 287: This is going to break an external use case where `finalize()` is called without arguments. Can you save `reuseMaps` in `setProps` instead?

DuGlaser: Thanks for the review!
I changed it as shown in 936cbb1.

Pessimistress (approved):
Thank you!

PR #1556, opened 2021-08-25
Fix link in docs

dangbert:
tiny change to fix link to https://github.com/visgl/react-map-gl/blob/master/docs/api-reference/interactive-map.md


Pessimistress (approved):
Thank you!

PR #1572, opened 2021-09-08
Add automated release workflow

Pessimistress:


kylebarron (approved):
Is this `github-release.js` identical to deck's? Maybe we could put in ocular dev tools in the future to share among all visgl repos?

PR #1580, opened 2021-09-21
added showUserHeading to be able to show heading.

Cmoen11:


Pessimistress:
This also need to be added to `propTypes` and `defaultProps` in [geolocate-control.js](https://github.com/visgl/react-map-gl/blob/3ad2bf37f3fc25d5fc86b792c7288cfb1391648f/src/components/geolocate-control.js), and the [documentation](https://github.com/visgl/react-map-gl/blob/3ad2bf37f3fc25d5fc86b792c7288cfb1391648f/docs/api-reference/geolocate-control.md).

Cmoen11:
@Pessimistress I've updated the commit.. I see that my prettier formatted some code, unsure if .prettierrc has changed since last edit of these files? If not, I'll revert the style changes.

Pessimistress (approved):

PR #1603, opened 2021-11-03
Omit container prop on mapOptions in types

dellisd:
react-map-gl does not use the `container` property in `MapboxOptions`, but `MapboxOptions` *requires* this property making it impossible to pass any map options to `ReactMapGL` without breaking it.

### DIFF
--- a/src/mapbox/mapbox.d.ts
+++ b/src/mapbox/mapbox.d.ts
@@ -50,7 +50,7 @@ export type MapboxProps = Partial<{
   bearing: number,
   pitch: number,
   altitude: number,
-  mapOptions: MapboxGL.MapboxOptions
+  mapOptions: Omit<MapboxGL.MapboxOptions, 'container'>
 }>;
 
 export default class Mapbox {


Pessimistress (approved):

easbar:
Also setting `InteractiveMapProps#transitionEasing` seems required since v5.3.17, even when we just like to use the default. And by the way should react-map-gl v5 use the `v1` folder of the mapbox types somehow? See here: https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/mapbox-gl

Pessimistress:
Master is currently using Mapbox v2. v5 is published from the 5.x-release branch and we'll need to make a PR directly to it. Looks like v5 should add `@types/mapbox-gl` as a dependency and import the type accordingly?

easbar:
v5 already has `  "@types/mapbox-gl": "^2.0.3",` as a dependency. Assuming the types might change while v5 stays at mapbox-gl v1.x this is probably not ideal. So yes either keep this dependency, but make sure the `v1` folder is used somehow (not sure how) or take the typings for Mapbox 1.x from else where. Copying over `MapboxOptions` wouldn't be sufficient at the moment, because there is also this:

```javascript
export interface MapLoadEvent {
  type: string;
  target: MapboxGL.Map;
}
```
in v5's `mapbox.d.ts` already. We would have to copy `MapboxGL.Map` as well.

Pessimistress:
Thanks for checking. While it's correct to omit `container` (the map should always render into the React one), shouldn't we use `Partial<MapboxOptions>` to begin with?

dellisd:
All of the other properties in `MapboxOptions` are optional, so I feel using `Partial` wouldn't make a practical difference unless we were expecting some other required properties to be added in the future.

In either case, if we want to use `Partial` I think it would be best to go with `Partial<Omit<MapboxGL.MapboxOptions, 'container'>>` since including the `container` property actually _breaks_ this library.

PR #1698, opened 2022-01-20
✨ add support for style prop in Marker

A-Tokyo:
- Add new prop `style` to Marker component.

This allows users to pass inline styles to the Marker component, eg: zIndex

Note: This PR is targeted towards the 5.3 release since create react app users can't yet upgrade to v6 due to transpilation issues.

closes https://github.com/visgl/react-map-gl/issues/1697

### DIFF
--- a/src/components/marker.d.ts
+++ b/src/components/marker.d.ts
@@ -4,7 +4,8 @@ import type {DraggableControlProps} from './draggable-control';
 export type MarkerProps = DraggableControlProps & {
   className?: string,
   longitude: number,
-  latitude: number
+  latitude: number,
+  style?: { [key: string]: string }
 };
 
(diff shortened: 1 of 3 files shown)


Pessimistress (approved):

src/components/marker.d.ts
```diff
@@ -4,7 +4,8 @@ import type {DraggableControlProps} from './draggable-control';
 export type MarkerProps = DraggableControlProps & {
   className?: string,
   longitude: number,
-  latitude: number
+  latitude: number,
+  style?: { [key: string]: string }
```
A-Tokyo, line 8:
```suggestion
  style?: { [key: string]: string | number }
```

PR #1785, opened 2022-03-02
Remove optional chained operator

remotezygote:
Currently this library breaks when compiling under `fuse-box` due to an issue with optional chaining (see https://github.com/fuse-box/fuse-box/issues/1981). This removes the one instance that is not handled correctly. This change allows the library to be used with `fuse-box` again (it was unusable since v7), and does not change the logic.

### DIFF
--- a/src/components/use-control.ts
+++ b/src/components/use-control.ts
@@ -18,7 +18,8 @@ export default function useControl<T extends IControl>(
   useEffect(() => {
     const {map} = context;
     if (!map.hasControl(ctrl)) {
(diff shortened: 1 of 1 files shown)


Pessimistress:
There are plenty more optional chaining in the src (search for `?.`). I do not think we should change how the code is written based on a "use at your own risk" version of a bundler.

FWIW, the published package contains a fully transpiled version (package.json `"main": "dist/es5/index.js"`) that does not include any new JS features. You can import from `react-map-gl/dist/es5` in your project while the bundler resolves its issues.

PR #1789, opened 2022-03-03
[Bug] MarkerProps "onClick" property should be optional

slaymance:
Fix for [issue 1788](https://github.com/visgl/react-map-gl/issues/1788).

Changing the property to optional is backwards compatible for anyone using `MarkerProps` so I think this fix is pretty straightforward. Documentation doesn't need updating as `onClick` isn't described as required: https://visgl.github.io/react-map-gl/docs/api-reference/marker#onclick


Pessimistress (approved):
Thank you!

PR #1803, opened 2022-03-10
removed weird type

donnikitos:
Return type of `useMap()` was breaking builds:

```
node_modules/react-map-gl/dist/esm/components/use-map.d.ts:13:5 - error TS2411: Property 'current' of type 'MapRef | undefined' is not assignable to 'string' index type 'MapRef'.

13     current?: MapRef;
       ~~~~~~~
```


Pessimistress:
#1799

PR #1831, opened 2022-04-15
Fix rendering issue with globe view at high zoom

SnailBones:
Fixes https://github.com/mapbox/mapbox-gl-js/issues/11750

In the upcoming globe projection in mapbox-gl-js, the projection automatically transitions to Mercator at high zoom levels. When used with react-map-gl, this created an inconsistent state where between `Mapbox._renderTransform.projection` would be correctly set to Mercator but `Mapbox._map.transform.projection` would be reset to globe at every update, creating strange rendering artifacts.

This PR provides a fix by preserving `projection` from the previous state of the transform.

### Before:

https://user-images.githubusercontent.com/14878684/163887380-fa00ef5b-9b55-4394-938f-4b51fdcd9c31.mov

### After:

https://user-images.githubusercontent.com/14878684/163887390-7bc65cda-9c47-44c9-8c0d-0a687b009cdb.mov

### DIFF
--- a/src/mapbox/mapbox.ts
+++ b/src/mapbox/mapbox.ts
@@ -852,7 +852,9 @@ export default class Mapbox {
         // Replace map.transform with ours during the callbacks
         map.transform = this._renderTransform;
         baseFire.call(map, event, properties);
+        const savedProjection = map.transform.projection;
         map.transform = tr;
+        map.transform.projection = savedProjection;
 
         return map;
       }
--- a/src/types/index.ts
+++ b/src/types/index.ts
@@ -17,6 +17,8 @@ export type {
   ScaleControl as MapboxScaleControl
 } from 'mapbox-gl';
 
+import type {ProjectionSpecification} from '../types';
+
 /**
  * Stub for mapbox's Transform class
  * https://github.com/mapbox/mapbox-gl-js/blob/main/src/geo/transform.js
@@ -29,6 +31,7 @@ export type Transform = {
   bearing: number;
   pitch: number;
   padding: PaddingOptions;
+  projection: ProjectionSpecification;
 
   clone: () => Transform;
   resize: (width: number, height: number) => void;


src/mapbox/mapbox.ts
```diff
@@ -852,7 +852,9 @@ export default class Mapbox {
         // Replace map.transform with ours during the callbacks
         map.transform = this._renderTransform;
         baseFire.call(map, event, properties);
+        const savedProjection = map.transform.projection;
         map.transform = tr;
+        map.transform.projection = savedProjection;
```
Pessimistress, line 857: Looks like the auto switching of projection happens in `map._render`:

https://github.com/mapbox/mapbox-gl-js/blob/9fff53107b78bdaae875693b7b34c7f9acb5a1ae/src/ui/map.js#L2950

Would it make more sense for us to match the two transforms in

https://github.com/visgl/react-map-gl/blob/bf5514908f653c60741136d02befefbadd0dff82/src/mapbox/mapbox.ts#L884

`tr.projection = this._map.transform.projection`?

SnailBones: I tried that initially, and while it fixes the rendering issue, it introduces a bug where the map position jumps on zooming out past the Mercator/Globe threshold.

https://user-images.githubusercontent.com/14878684/165402326-332c9939-4c2f-4db1-9860-ff0aa809e0e9.mov

SnailBones:
@Pessimistress Just wanted to check in on the status of this PR! Do you have any other concerns that I can address?

It would be great to have this fixed before the mapbox-gl-js globe release coming later this month.

Pessimistress:
Can you provide steps I can follow to test this feature?

src/mapbox/mapbox.ts
Pessimistress: I have seen the same issue in matching elevation between the two transforms. In addition to assigning

`tr.projection = this._renderTransform.projection`

You may also need to call `_calcMatrices()` on the transform.

SnailBones:
@Pessimistress Thank you for the quick response! You can install the globe alpha with `npm i mapbox/mapbox-gl-js#globe-alpha -s`

I've also been able to test the development version of `mapbox-gl-js` as follows:

- Install and build the latest version of mapbox-gl-js. (Docs [here](https://github.com/mapbox/mapbox-gl-js/blob/main/CONTRIBUTING.md))
- Use `npm link` to link the test application's `node_modules/mapbox-gl` to your local version of the `mapbox-gl-js` repository.
- Then in my test application:

```js
import mapboxgl from '!mapbox-gl';
...
    <Map
        mapLib={mapboxgl}
        projection="globe"
        ...
    >
    </Map>
```

Hope that helps!

src/mapbox/mapbox.ts
SnailBones: Thank you, I'll give that a try!

SnailBones:
Testing this again, it seems this has since been fixed by https://github.com/visgl/react-map-gl/pull/1841! Many thanks! 🙏

PR #1843, opened 2022-04-23
 Fix MapProvider prop types

shermify:
As of React 18 (@types/react), `children` is no longer an implicit prop on the React.FunctionComponent type.  This PR adds the type explicitly.  More info here

https://github.com/DefinitelyTyped/DefinitelyTyped/issues/46691
https://solverfox.dev/writing/no-implicit-children/


Pessimistress (approved):
Thank you!

PR #1851, opened 2022-05-02
feat: Support for custom mapbox layer

ptbrowne:
Instead of using DeckGl as the container for Mapbox, we use Mapbox
as the container for DeckGl layer. This lets us insert the deckgl layer
between the mapbox layers, which is very handy when dealing with labels.

We use DeckGL's custom Mapbox layer here. Since DeckGL's custom layer
need method on the prototype, we can copy the layer props with spreading,
otherwise we lose the methods. This is why the `layer` prop is added
on the Layer so that we can pass the custom layer without any copying.

I'd like to have your opinion on this way of doing before continuing the work on this PR (maybe improving the documentation or the example).


Pessimistress:
Thank you for the proposal. To simplify things, I suggest we don't change the prop types for the existing use cases, and just add

```typescript
type CustomLayer = {
  type: 'custom',
  layer: CustomLayerInterface
}
```

What do you think?

sven-meyer-wetter-com:
Any news on that?

hammaIX:
+1. This would be a really useful feature.

Reichenberg:
I'm very much interested in seeing this merged. is there a way to help this along?

Reichenberg:
@ptbrowne

stampycode:
Was trying to recreate in react-map-gl the example [here](https://docs.mapbox.com/mapbox-gl-js/example/add-3d-model/) to display a 3d model, but it looks like this isn't currently possible.

Is there a known workaround for this?

Many thanks!

src/components/layer.ts
```diff
@@ -59,27 +63,40 @@ function updateLayer(map: MapboxMap, id: string, props: LayerProps, prevProps: L
...
+function createLayer(map: MapboxMap, id: string, layerProps: LayerProps, beforeId: string) {
+  // @ts-ignore
+  if (isMapStyleLoaded(map) && (!('source' in layerProps) || map.getSource(layerProps.source))) {
     // @ts-ignore
-    map.addLayer(options, props.beforeId);
+    map.addLayer(layerProps, beforeId);
   }
 }
 
 /* eslint-enable complexity, max-statements */
```
ZiaurRehmanmalhi, line 80: remove extra lines

mourner:
@Pessimistress hi! Just wanted to chime in on the custom layer feature too — it's very useful. I got around that somewhat by setting it up as a factory function making an object (instead of a class instance), and silencing TypeScript errors when using those props on `<Layer .../>`, but it would be really nice to have an easier time setting up custom layers with this project.

PR #1872, opened 2022-05-23
Fix create react app builds with es2015 target

### LINKED ISSUE #1871: [Bug] es2020 target breaks CRA 4 builds
### Description

When I use Create React App version 4 (react-scripts, ^4.0.3) with `react-map-gl`, I get the following error:

```bash
Failed to compile.

./node_modules/react-map-gl/dist/esm/components/use-control.js 12:46
Module parse failed: Unexpected token (12:46)
File was processed with these loaders:
 * ./node_modules/babel-loader/lib/index.js
You may need an additional loader to handle the result of these loaders.
| 
|     if (!map.hasControl(ctrl)) {
>       map.addControl(ctrl, (opts || onRemove)?.position);
|     }
|
```

This appears to be due to an [aggressive target of ES2020](https://github.com/visgl/react-map-gl/blob/master/tsconfig.json#L3) which does not transform the optional chaining `?` code. If the `target` was reduced to `es2015` the optional chaining code would be transformed and Webpack would no longer throw on this code. I tested this out and the transformed code in `/dist/ems/components/use-controls.js` no longer has the optional chaining token:

```js
        if (!map.hasControl(ctrl)) {
            map.addControl(ctrl, (_a = (opts || onRemove)) === null || _a === void 0 ? void 0 : _a.position);
        }
```

This should not be a breaking change because previous versions of JavaScript are compatible with newer versions.

If you are open to this change, I'm willing to open a PR if that would be helpful.

### Expected Behavior

Create React App 4 shouldn't throw an error when building `react-map-gl`.

### Steps to Reproduce

I created a sample repo that will show the error:
https://github.com/reesscot/react-mapbox-gl-test

Run:
```
git clone git@github.com:reesscot/react-mapbox-gl-test.git
cd react-mapbox-gl-test 
npm install
npm start
```
Observe error above.

### Environment

- Framework version: `"react-map-gl": "^7.0.14",`
- Map library: `"mapbox-gl": "^2.8.2",`
- Browser: N/A
- OS: Mac OS X

### Logs

N/A it's a build error

reesscot:
This fixes #1871 by setting the `target` to `es2015`

Tested using the following steps:
1. Edited target to `es2015`
1. Ran `yarn build`
1. Verified that optional chaining code in use-control.js was tranformed:

```js
        if (!map.hasControl(ctrl)) {
            map.addControl(ctrl, (_a = (opts || onRemove)) === null || _a === void 0 ? void 0 : _a.position);
        }
```
4. Copied dist folder to https://github.com/reesscot/react-mapbox-gl-test project's `node_modules` folder.
5. Noted that error message around optional chaining in `use-control.js` went away and got successful build:

image

### DIFF
--- a/tsconfig.esm.json
+++ b/tsconfig.esm.json
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es2020",
+    "target": "es2015",
     "jsx": "react",
     "moduleResolution": "node",
     "module": "ES2020",
--- a/tsconfig.json
+++ b/tsconfig.json
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "target": "es2020",
+    "target": "es2015",
     "jsx": "react",
     "allowSyntheticDefaultImports": true,
     "resolveJsonModule": true,


Pessimistress:
See relevant discussion here: https://github.com/visgl/react-map-gl/discussions/1791#discussioncomment-2289886

The rationale behind the current build target is to allow users to retain the max control of their transpilation. The transform of optional chaining is non reversible. We already have a dist/es5 entry point that is fully transpiled if that is what you wish to target.

Even if we do want to downgrade the build target, is ES2018 not enough?

Issue #1950, opened 2022-08-11
[Bug] `onMouseMove` ignores `interactiveLayerIds`


tordans:
### Description

The way I read the docs for `onMouseMove` (https://visgl.github.io/react-map-gl/docs/api-reference/map#onmousemove) when I specify a layer in `interactiveLayerIds`, the events should only fire while I move the mouse over one of the elements of those specified layers. 

That is how it works for `onMouseEnter` (console log `a`) and `onMouseLeave` (console log `b`) on the blue line.

However, for `onMouseMove` the events fire all the time.

<img width="886" alt="image" src="https://user-images.githubusercontent.com/111561/184223892-933f59cd-26de-49ee-b8fe-cfc6c687a0a5.png">

### Expected Behavior

I expected `onMouseMove` to behave the same way as `onMouseEnter/Leave`, to only fire when hovering the blue lines (which are the layer I specified with `interactiveLayerIds`).

### Steps to Reproduce

(On request)

### Environment

- Framework version: `"react-map-gl": "^7.0.17",`
- Map library: `"maplibre-gl": "^2.1.9",`
- Browser: Chrome
- OS: macOS

Pessimistress:
Please create a CodeSandbox that reproduces your issue.

tordans:
> Please create a CodeSandbox that reproduces your issue.

I created https://github.com/visgl/react-map-gl/pull/1953 which shows the issue.

<img width="592" alt="image" src="https://user-images.githubusercontent.com/111561/184693848-a5337ee8-71f0-4493-ad10-eabba1b79a2f.png">

The console.error is the `onEnter` logging, which only fires once when I enter the first state from the bottom of the screen. The 350+ console.log fire for `onMouseMove` which I take as an indicator that the `interactiveLayerIds` are not taken into account.

HTH

Pessimistress:
Ok this was changed because of https://github.com/visgl/react-map-gl/issues/1893. In mapbox-gl you can register multiple callbacks for the same event like

```js
map.on('click', () => {
  // clicked anywhere
});
map.on('click', 'some-layer', () => {
  // clicked on layer
});
```

Since the React API can only take one callback for each event, it becomes tricky to handle both the layer-specific and "anywhere" situation. While it is relatively easy to check for empty `features` (this issue), there is no quick workaround if the event is not fired (issue #1893), so I think that the current behavior covers more use cases.

The documentation should be updated to reflect the current behavior.

tordans:
[...]

PR #1953, opened 2022-08-15
This is a debug example for #1950

tordans:
https://github.com/visgl/react-map-gl/issues/1950


Pessimistress:
Thanks for putting this together.

PR #1954, opened 2022-08-16
Update event callback documentation

Pessimistress:
For https://github.com/visgl/react-map-gl/issues/1950

### DIFF
--- a/docs/api-reference/map.md
+++ b/docs/api-reference/map.md
@@ -280,13 +280,13 @@ Called when an error occurs.
 
 Called when a pointing device (usually a mouse) is pressed within the map.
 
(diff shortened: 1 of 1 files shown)


docs/api-reference/map.md
```diff
@@ -280,13 +280,13 @@ Called when an error occurs.
 
 Called when a pointing device (usually a mouse) is pressed within the map.
 
-If `interactiveLayerIds` is specified, the event will fire only when the the cursor is pressed while inside a visible portion of the specifed layer(s).
+If `interactiveLayerIds` is specified, the event will contain an additional `features` field that contains features under the cursor from the specifed layer.
```
kylebarron, line 283:
```suggestion
If `interactiveLayerIds` is specified, the event will contain an additional `features` field that contains features under the cursor from the specified layer.
```

ditto in each below

kylebarron (approved):

Issue #1844, opened 2022-04-24
[Bug] SymbolLayer crashes on style update


wafisher:
### Description

When using `Source/Layer` with a `SymbolLayer` and a (modified) Immutable base style, I get the following in the console.

```
mapbox-gl.js:35 Uncaught TypeError: Cannot read properties of undefined (reading 'get')
    at new It (mapbox-gl.js:35:1)
    at zt.continuePlacement (mapbox-gl.js:35:1)
    at Nt._updatePlacement (mapbox-gl.js:35:1)
    at Map._render (mapbox-gl.js:35:1)
    at Map.map._render (mapbox.ts:552:1)
    at mapbox-gl.js:35:1
```

There's some interaction with both of these that brings about the issue. If I remove the SymbolLayer, I don't get this issue. And if I don't modify the style, I also don't get the issue.

### Steps to Reproduce

I've created a minimal reproduction here: https://codesandbox.io/s/great-mccarthy-0pozvg. Click on the map to trigger the issue.

### Environment

- Framework version: 7.0.11
- Browser: latest Chrome and Safari

Pessimistress:
I can reproduce the crash with native mapbox-gl: https://codesandbox.io/s/proud-wave-gohost

You may wish to report this to the [Mapbox GL JS repo](https://github.com/mapbox/mapbox-gl-js/issues).

wafisher:
Done https://github.com/mapbox/mapbox-gl-js/issues/11820

wafisher:
I believe this is fixed with the new release?

Just confirming: #2047 is strictly a workaround because https://github.com/mapbox/mapbox-gl-js/issues/11820 has not been resolved, correct?

Pessimistress:
You are correct that the commit is only a workaround. We can revisit if Mapbox fixes the issue.

Closing as resolved as of v7.0.20.

PR #2047, opened 2022-11-14
Fix SymbolLayer crash on style update

sim51:
Symbol layer crashed when style are updated with the error `Cannot read properties of undefined (reading 'get')`  (see  #1844)

This PR handles the condition when the `Source` is updated while the style is loading, by using a `setTimeout(...,0)`. 

As describe [here](https://github.com/maplibre/maplibre-gl-js/issues/1835#issuecomment-1310741571), an other solution is to listen on `style.load` instead of `styledata`, but it's a private event...

So IMHO, the `setTimeout` is better.

### DIFF
--- a/src/components/source.ts
+++ b/src/components/source.ts
@@ -94,7 +94,7 @@ function Source(props: SourceProps) {
 
   useEffect(() => {
     if (map) {
-      const forceUpdate = () => setStyleLoaded(version => version + 1);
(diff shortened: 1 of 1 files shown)


Pessimistress (approved):
Thank you!

sim51:
Sorry my code was bad, I amended my commit. 
This time the build task should pass.

sim51:
Thanks a lot for the merge @Pessimistress 
Is there a planned release for the v7 ?

PR #2065, opened 2022-12-08
Add a null-check to map.style in createSource

NameFILIP:
Fixes https://github.com/visgl/react-map-gl/issues/2064

### DIFF
--- a/src/components/source.js
+++ b/src/components/source.js
@@ -33,7 +33,7 @@ const propTypes = {
 let sourceCounter = 0;
 
 function createSource(map, id, props) {
-  if (map.style && map.style._loaded) {
+  if (map && map.style && map.style._loaded) {
     const options = {...props};
     delete options.id;
     delete options.children;


Pessimistress (approved):

NameFILIP:
not sure if I'm supposed to do anything about the build failure:
```
error cosmiconfig@8.0.0: The engine "node" is incompatible with this module. Expected version ">=14". Got "12.22.12"
error Found incompatible module.
```

It looks unrelated to my change. Probably `--ignore-engines` should be added to the `yarn` command

NameFILIP:
@Pessimistress it's interesting that `yarn.lock` file does not reference `cosmiconfig@8.0.0` and `yarn install` works fine. The error is caused by `ocular-bootstrap` called in `yarn bootstrap`: https://github.com/visgl/react-map-gl/blob/5.3-release/package.json#L35

There are several options to solve it:
1. Increase the `node` version to `14` in https://github.com/visgl/react-map-gl/tree/5.3-release/.github/workflows files
2. Add `--ignore-engines` in the same files as above
3. Use older version of `ocular-dev-tools `

I think number 1 should be good or which one do you recommend choosing?

Pessimistress:
Yes we should use a newer version of node in CI

NameFILIP:
@Pessimistress this one is now ready with the passing CI.

NameFILIP:
@Pessimistress would it be possible to merge this fix and release it? 🙏

NameFILIP:
Thank you! 🥳

Issue #2071, opened 2022-12-12
[Bug] ReuseMaps does not destroy its children


evrazi:
### Description

Maps that have reuseMaps="true" do not destroy its children

### Expected Behavior

Maps that have reuseMaps="true" destroy their children like in v6 of react-map-gl

### Steps to Reproduce

https://codesandbox.io/s/flamboyant-browser-9j9clt?file=/src/Map.js

### Environment

- Framework version: "react-map-gl": "^7.0.20",
- Map library: "mapbox-gl": "^2.9.1",
- Browser: Chrome 108.0
- OS:

lh0x00:
I tried solving it at #2092.

Pessimistress:
Fix landed in v7.0.21

PR #2092, opened 2022-12-29
fix `reuseMaps` actions wrong expectations

lh0x00:
In this PR I have solved issue #2071 
I've tried and created a temporary release in the meantime, you can check out the results [here](https://codesandbox.io/s/modern-breeze-v30usd?file=/src/Map.js), this forked from issue report.

Please review and review it, many thanks. ^^

cc @Pessimistress @theRuslan

### DIFF
--- a/src/components/map.tsx
+++ b/src/components/map.tsx
@@ -70,6 +70,7 @@ const Map = forwardRef<MapRef, MapProps>((props, ref) => {
   const mountedMapsContext = useContext(MountedMapsContext);
   const [mapInstance, setMapInstance] = useState<Mapbox>(null);
   const containerRef = useRef();
+  const childrenRef = useRef<HTMLDivElement>();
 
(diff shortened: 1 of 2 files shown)


Pessimistress:
Thanks for the proposal. `cloneElement` could be quite expensive if there are a lot of child elements. Can you try wrapping all children in a single div and removing that div upon dismount?

lh0x00:
I tried using react's ref but it didn't give the desired effect because the mechanism will go after we save the value, you can see more here https://github.com/visgl/react-map-gl/pull/2092/commits/2e4df069db0056c0db20f9481bcf51cad228cb13.

cc @Pessimistress

lh0x00:
Please review and release it soon if possible, I'm really waiting for it. Thank you ^^.

Pessimistress (approved):

Pessimistress:
Can you run `yarn lint fix`?

lh0x00:
@Pessimistress it's ready to merge

PR #2102, opened 2023-01-15
add support for all DOM events to Marker

charkour:
Thank you for supporting this package!

When using the Marker component, I'd like to attach more event handlers in this capacity below. `onMouseEnter` is not currently supported by react-map-gl. The proposed code will add the proper event listeners when the props are provided. 

```tsx
      <Map
        mapStyle="mapbox://styles/mapbox/streets-v11"
        mapboxAccessToken={process.env.NEXT_PUBLIC_MAPBOX_TOKEN}
      >
        {Object.values(markers).map((marker) => (
          <Marker
            key={marker.id}
            onMouseEnter={() => {
              console.log("marker hovered", marker.id);
              setHighLightedId(marker.id);
            }}
            style={{
              backgroundColor:
                marker.id === highLightedId ? "red" : "blue",
            }}
            longitude={marker.lng}
            latitude={marker.lat}
          />
        ))}
      </Map>
```

If this is a helpful addition, please let me know! We can add tests and improve the documentation. 

Thank you

### DIFF
--- a/src/components/marker.ts
+++ b/src/components/marker.ts
@@ -17,7 +17,7 @@ import type {
 import {MapContext} from './map';
 import {arePointsEqual} from '../utils/deep-equal';
 
-export type MarkerProps = {
+export type MarkerProps = Omit<React.HTMLAttributes<HTMLElement>, 'color' | 'draggable'> & {
   /** Longitude of the anchor location */
   longitude: number;
   /** Latitude of the anchor location */
@@ -65,9 +65,6 @@ export type MarkerProps = {
   scale?: number;
   /** A Popup instance that is bound to the marker */
   popup?: MapboxPopup;
-  /** CSS style override, applied to the control's container */
-  style?: React.CSSProperties;
-  onClick?: (e: MapboxEvent<MouseEvent>) => void;
   onDragStart?: (e: MarkerDragEvent) => void;
   onDrag?: (e: MarkerDragEvent) => void;
   onDragEnd?: (e: MarkerDragEvent) => void;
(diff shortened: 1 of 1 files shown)


src/components/marker.ts
```diff
@@ -102,13 +99,20 @@ function Marker(props: MarkerProps) {
...
-        target: mk,
-        originalEvent: e
-      });
-    });
+    for (const propName of Object.keys(thisRef.current.props)) {
+      const isHandler = /^on[A-Z]/.test(propName);
+      if (isHandler) {
+        // slice(2) to remove 'on' prefix
+        const eventName = propName.slice(2).toLowerCase();
+        mk.getElement().addEventListener(eventName, (e: MapboxEvent<React.SyntheticEvent>) => {
```
charkour, line 107: We'll have to see what the proper event typing is.

src/components/marker.ts
```diff
@@ -17,7 +17,7 @@ import type {
 import {MapContext} from './map';
 import {arePointsEqual} from '../utils/deep-equal';
 
-export type MarkerProps = {
+export type MarkerProps = Omit<React.HTMLAttributes<HTMLElement>, 'color' | 'draggable'> & {
```
charkour, line 20: I was not sure if `color` and `draggable` have different types that the default `HTMLElement` type so I omitted them.

Pessimistress (commented):
Thanks for the proposal. I have a couple of concerns about the current implementation:

- You are accepting all prop types from `React.HTMLAttributes<HTMLElement>` but only the `on*` ones seem to be used?
- Event handlers are attached at the initial mount, which means the user cannot add new callbacks after that. I can see that behavior potentially leads to confusion in some cases.

Does it make sense to use `createPortal` to render a `<div {...props}>` into the marker element? All prop handling will be deferred to React then.

charkour:
Using a portal sounds great! (Sorry for the late reply)

charkour:
I opted for using an event listener attached via a ref.

PR #2105, opened 2023-01-18
Fix crash in React 18 StrictMode (#1680)

NameFILIP:
Since `v5.3.x` is not working well (https://github.com/visgl/react-map-gl/issues/2103), please accept the port of the fix for React 18 into the `v5.2.x` branch.

### DIFF
--- a/src/mapbox/mapbox.js
+++ b/src/mapbox/mapbox.js
@@ -365,15 +365,23 @@ export default class Mapbox {
     /* eslint-disable accessor-pairs */
     const {container} = props;
     // $FlowFixMe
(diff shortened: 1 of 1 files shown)


NameFILIP:
@Pessimistress would it be possible to merge and release this fix? 🙏

NameFILIP:
@Pessimistress it's the same commit that was merged into `5.3-releases`, but `5.3` is less stable than `5.2`, so asking to merge/release it for `5.2` as well.

NameFILIP:
Thank you! 🙏

PR #2109, opened 2023-01-24
Docs: Add types to use control example

tordans:
The page https://github.com/visgl/react-map-gl/blob/master/docs/api-reference/use-control.md?plain=1#L9 / http://visgl.github.io/react-map-gl/docs/api-reference/use-control uses the Types `DrawControlProps` which are not specified in the example / on the page.

They are specified on this page https://github.com/visgl/react-map-gl/blob/master/examples/draw-polygon/src/draw-control.ts#L6-L12 , however.

This PR add them to the example.

### DIFF
--- a/docs/api-reference/use-control.md
+++ b/docs/api-reference/use-control.md
@@ -6,6 +6,14 @@ The `useControl` hook is used to create React wrappers for custom map controls.
 import MapboxDraw from '@mapbox/mapbox-gl-draw';
 import Map, {useControl} from 'react-map-gl';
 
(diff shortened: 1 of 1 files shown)


Pessimistress (approved):

tordans:
FYI, we have very similar types in use at https://github.com/FixMyBerlin/atlas-app/blob/develop/src/components/MapInterface/Map/Calculator/CalculatorControlsDrawControl.tsx#L10-L17 so this PR might be right after all…

Issue #2117, opened 2023-02-01
[Feat] Change "defaultProps" strategy in components to "JavaScript default parameters"


luanmm:
### Target Use Case

Anticipate to deprecated functionality removal (and avoid error/warning messages in console).

### Proposal

As it is stated in React DOM warnings:

```
Warning: ScaleControl: Support for defaultProps will be removed from function components in a future major release. Use JavaScript default parameters instead.
```

Components should be using JavaScript default parameters as an alternative.

elewin:
I am getting this on Marker as well:

```
Warning: Marker: Support for defaultProps will be removed from function components in a future major release. Use JavaScript default parameters instead
``

This is using react-map-gl 7.0.25

luanmm:
Looking in v7.0.25 tag, seems that PR #2173 is not present in code. Don't know why exactly, as it appears to be merged. So it should keep giving warnings, indeed, until a release with this change.

Pessimistress:
This will be landed in v7.1z

PR #2125, opened 2023-02-21
Update how defaultprops are handled on functional components

fabriciopirini:
As per issue #2117, this PR removes the usage of defaultProps and move it to inside the functional component by merging the default parameters with the component props and using useMemo in order to avoid unnecessary recalculation.


Pessimistress:
Thanks. I feel that merging by creating a new object is unnecessary overhead. What do you think of the following pattern:

```
const defaultProps = {
  onClick: () => {}
}

function SomeComponent(props) {
  const {
    onClick = defaultProps.onClick
  } = props;

 ...
}
```

Pessimistress:
Closed in favor of #2173

PR #2126, opened 2023-02-23
fix: `onMouseEnter` & `onMouseLeave` trigger between joined layers (#1671)

### LINKED ISSUE #1671: onMouseEnter isn't triggered when moving between layers from different sources
I wanted onMouseEnter to trigger when moving between layers from different sources but it only triggers for the first layer entered.

Here is a gif showing the problem.
map-gl

And here is a codesandbox with the example.
https://codesandbox.io/s/gracious-greider-08zrq

Jungzl:
fix #1671


src/mapbox/mapbox.ts
```diff
@@ -786,14 +786,23 @@ export default class Mapbox {
     }
   };
 
+  _uniqueLayersCount(features: MapboxGeoJSONFeature[] | null) {
+    if (!features) {
+      return 0;
+    }
+    return features.filter(
+      (feature, index) => features.findIndex(f => f.layer.id === feature.layer.id) === index
```
Pessimistress, line 794: This is very inefficient. Use `new Set(ids).size`

src/mapbox/mapbox.ts
```diff
@@ -806,14 +815,14 @@ export default class Mapbox {
       } else {
         features = [];
       }
-      const isHovering = features.length > 0;
+      const hoveringLayersCount = this._uniqueLayersCount(features);
 
-      if (!isHovering && wasHovering) {
+      if (hoveringLayersCount < hoveredLayersCount) {
```
Pessimistress, line 820: Is this correct, though? What if the layer count is the same but one layer is replaced by another?

src/mapbox/mapbox.ts
Jungzl: done

src/mapbox/mapbox.ts
Jungzl: don't know if it's ok to call pointEvents twice in one callback

PR #2132, opened 2023-03-02
Inherit parent height in`mapboxgl-children` div

martpet:
The new 'mapboxgl-children' div brakes styling for those who depend on the default height of 100% of the parent div. Adding `height: inherit` to the new div solves the issue.

### DIFF
--- a/src/components/map.tsx
+++ b/src/components/map.tsx
@@ -153,7 +153,7 @@ const Map = forwardRef<MapRef, MapProps>((props, ref) => {
     <div id={props.id} ref={containerRef} style={style}>
       {mapInstance && (
         <MapContext.Provider value={contextValue}>
-          <div mapboxgl-children="">{props.children}</div>
+          <div mapboxgl-children="" style={{ height: 'inherit' }}>{props.children}</div>
         </MapContext.Provider>
       )}
     </div>


Pessimistress:
Can you give an example of when this makes a difference?

martpet:
An example is when there is a child with a height in percentage of the parent height.

Of course, one could wrap the child with a div with ‘position: absolute; height: 100%’, since  the ‘mapboxgl-children’ div has ‘position: relative’. In this case, I agree, my pull request is not needed.

For me, the layout broke, because my child has ‘height: calc(100% - (100vh - 100dvh))’

src/components/map.tsx
```diff
@@ -153,7 +153,7 @@ const Map = forwardRef<MapRef, MapProps>((props, ref) => {
     <div id={props.id} ref={containerRef} style={style}>
       {mapInstance && (
         <MapContext.Provider value={contextValue}>
-          <div mapboxgl-children="">{props.children}</div>
+          <div mapboxgl-children="" style={{ height: 'inherit' }}>{props.children}</div>
```
Pessimistress, line 156: Create a constant and reference it here (`style={CHILD_CONTAINER_STYLE}`)

Pessimistress:
Thanks. For CI to pass, you need to run `yarn lint fix` at project root to apply prettier styles.

Pessimistress (approved):

PR #2146, opened 2023-03-23
fix: add missing field performanceMetricsCollection

jaenal-CF:
Hello there!

I am working on a project that requires cookie consent policy. I notice that using mapbox, it will always trigger the performance cookies. For mapbox@2 it has a field that could disable the metric collection. Please let me know if I need more work for adding `performanceMetricsCollection` field

https://docs.mapbox.com/mapbox-gl-js/api/map/
https://www.mapbox.com/legal/cookies

A bunch of performance cookies trigger by mapbox
image

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
-  "name": "react-map-gl",
-  "description": "React components for Mapbox GL JS-compatible libraries",
-  "version": "7.0.0",
+  "name": "@s19/react-map-gl",
+  "description": "Fork of react components for Mapbox GL JS-compatible libraries",
+  "version": "7.0.1",
   "keywords": [
     "mapbox",
     "mapbox-gl",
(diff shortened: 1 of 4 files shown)


jaenal-CF:
@Pessimistress @xintongxia please review is there a way for me to pass this props to the mapbox in react-map-gl?

package.json
```diff
@@ -1,7 +1,7 @@
 {
-  "name": "react-map-gl",
-  "description": "React components for Mapbox GL JS-compatible libraries",
-  "version": "7.0.0",
+  "name": "@s19/react-map-gl",
+  "description": "Fork of react components for Mapbox GL JS-compatible libraries",
```
Pessimistress, line 3: Please remove this

Pessimistress:
This will be addressed in v7.1.

PR #2153, opened 2023-04-07
Fix crash in _setupUI workaround

bgeihsgt:
See https://github.com/visgl/react-map-gl/issues/2152 for more details


Pessimistress:
Moving discussion to the maplibre code base: https://github.com/maplibre/maplibre-gl-js/issues/2387

PR #2160, opened 2023-04-13
Force map reload upon reuse

### LINKED ISSUE #2154: [Bug] inconsistent map state upon recycle
### Description

When a map instance is recycled via the `reuseMaps` prop and an `initialViewState` is provided that differs from the final view state of the previous map, Mapbox state is being updated to reflect the new `initialViewState` but it does not load new basemap tiles or move the existing tiles to their new positions until the user pans/zooms the map.

Note: this is only an issue when `initialViewState` uses `latitude`/`longitude`. When `initialViewState` uses `bounds`, the map behaves as expected.

Likely caused by a bug [in `_updateViewState()`](https://github.com/visgl/react-map-gl/blob/074a4094b1d34c8df95774ec26c522e5c3dd6f64/src/mapbox/mapbox.ts#L651-L691) when called from here: https://github.com/visgl/react-map-gl/blob/master/src/mapbox/mapbox.ts#L491-L499

### Expected Behavior

When a map instance is recycled, the basemap tiles should immediately reflect the new `initialViewState`.

### Steps to Reproduce

Sandbox: https://codesandbox.io/s/bold-diffie-qhzcdt?file=/src/App.tsx

1. Load the sandbox
2. Drag the map around a little
3. Click the button to toggle `<Map>` components
4. Observe that the `<Marker>` is in the center of the map rather than at the coordinates it was set to
5. Drag the map again
6. Observe that the basemap tiles reload, putting the `<Marker>` where it belongs

### Environment

- Framework version: react-map-gl@7.0.21
- Map library: mapbox-gl@1.13.3
- Browser: Microsoft Edge@112.0.1722.34
- OS: Windows 10

### Logs

_No response_

Pessimistress:
Another attempt to resolve #2154

`map_update()` is called internally upon move event to reload sources and rerender.

### DIFF
--- a/src/mapbox/mapbox.ts
+++ b/src/mapbox/mapbox.ts
@@ -506,8 +506,9 @@ export default class Mapbox {
       map.once('styledata', () => map.fire('load'));
     }
 
-    // Force redraw
-    map.triggerRepaint();
+    // Force reload
+    // @ts-ignore
+    map._update();
     return that;
   }
 


michaelcgorman:
#2154 OP here. I haven't figured out how to update my Code Sandbox to this commit, but FWIW I manually applied the diff in my main project and it appears to have solved the problem.

chrisgervang (approved):

michaelcgorman:
Hey -- just checking in again about this. It merged into master a while ago, but it doesn't look like it was in yesterday's release. Was that intentional?

michaelcgorman:
Oh, awesome. Thank you!

PR #2170, opened 2023-05-07
Upgrade test harness

Pessimistress:


test/render/golden-images/basic-map.png
chrisgervang, line 1: The San Francisco label is slightly different. Is that expected?

chrisgervang (approved):

test/render/golden-images/basic-map.png
Pessimistress: Yes, the mapbox-gl dependency is upgraded

PR #2174, opened 2023-05-09
Fix reuseMaps with maplibre-gl ResizeObserver not changing observee

viliket:
Resolves https://github.com/visgl/react-map-gl/issues/2166


Pessimistress (approved):
Thank you!

PR #2175, opened 2023-05-09
Make mapbox-gl an optional dependency

Pessimistress:
#2161

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,7 @@ import Map from 'react-map-gl';
 
 function App() {
   return <Map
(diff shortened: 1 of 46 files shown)


chrisgervang (approved):
LGTM. I'm noticing this is planned for a minor release, but will break at runtime for a number of users. Might be worth adding a upgrade message in the error for a few releases.

Pessimistress:
After some debate I think it's better to stick to strict semantic versioning and land this in the next major version (8.0). Updated the upgrade guide accordingly.

PR #2178, opened 2023-05-23
Split exports into separate endpoints

Pessimistress:
For #2161


docs/api-reference/map.md
```diff
@@ -54,6 +55,8 @@ Returns the native [Map](https://docs.mapbox.com/mapbox-gl-js/api/map/) instance
 
 ## Properties
 
+Aside from the props listed below, the `Map` component supports all parameters of the [Map](https://docs.mapbox.com/mapbox-gl-js/api/map/#map-parameters) constructor. Beware that this is not an exhaustive list of all props. Different base map libraries may offer different options and default values. When in doubt, refer to your base map library's documentation.
```
chrisgervang, line 58: I think we should link to both mapbox and maplibre parameters.

docs/api-reference/map.md
```diff
@@ -496,156 +501,20 @@ function App() {
...
 #### `mapboxAccessToken`: string
 
 Token used to access the Mapbox data service. See [about map tokens](/docs/get-started/mapbox-tokens.md).
 
-#### `antialias`: boolean
-
-Default: `false`
-
-If `true` , the gl context will be created with [MSAA antialiasing](https://en.wikipedia.org/wiki/Multisample_anti-aliasing), which can be useful for antialiasing custom layers.
-This is `false` by default as a performance optimization.
```
chrisgervang, line 509: Are we relying on the basemap docs for these now? I can see how that reduces maintenance and confusion as the APIs diverge. I don't like how there are multiple places I'll need to look now, but I can't think of a better alternative.

src/components/attribution-control.ts
```diff
@@ -21,7 +21,7 @@ export type AttributionControlProps = {
 
 function AttributionControl(props: AttributionControlProps): null {
   const ctrl = useControl<MapboxAttributionControl>(
-    ({mapLib}) => new mapLib.AttributionControl(props),
+    ({mapLib}) => new mapLib.AttributionControl(props) as MapboxAttributionControl,
```
chrisgervang, line 24: Can we move to a lib-neutral type overtime on these as well? Seeing `Mapbox{type}` when using maplibre is confusing to the uninitiated.

src/components/map.tsx
```diff
@@ -16,17 +16,32 @@ import createRef, {MapRef} from '../mapbox/create-ref';
...
+type MapInitOptions<LibT extends MapLib> = Omit<
+  ConstructorParameters<LibT['Map']>[0],
+  'style' | 'container' | 'bounds' | 'fitBoundsOptions' | 'center'
+>;
+
+export type MapProps<LibT extends MapLib> = MapInitOptions<LibT> &
+  MapboxProps &
   GlobalSettings & {
-    mapLib?: any;
+    mapLib: LibT | Promise<LibT>;
```
chrisgervang, line 44: This is excellent!

src/components/marker.ts
```diff
@@ -122,7 +123,7 @@ function Marker(props: MarkerProps) {
   }, []);
 
   useEffect(() => {
-    marker.addTo(map.getMap());
+    marker.addTo(map.getMap() as any);
```
chrisgervang, line 126: Is this `unknown`, or can it be known from the context?

src/types/map.ts
```diff
@@ -0,0 +1,168 @@
...
+
+  off(type?: string | any, listener?: Listener);
+
+  once(type: string, listener: Listener);
+}
+
+/**
+ * A user-facing type that represents the minimal intersection between Mapbox.Map and Maplibre.Map
+ * User provided `mapLib.Map` is supposed to implement this interface
+ * Only losely typed for compatibility
```
chrisgervang, line 26: Should we ideally harden this type overtime?

docs/api-reference/map.md
```diff
@@ -26,20 +26,21 @@ function App() {
...
+  const mapRef = React.useRef<MapRef<mapboxgl.Map>>();
 
   const onMapLoad = React.useCallback(() => {
     mapRef.current.on('move', () => {
       // do something
     });
   }, []);
 
-  return <Map mapLib={import('mapbox-gl')} ref={mapRef} onLoad={onMapLoad} />;
+  return <Map mapLib={mapboxgl} ref={mapRef} onLoad={onMapLoad} />;
```
chrisgervang, line 43: It's unclear to me what kind of type hinting `<Map/>` provides now.. will it include `mapboxgl` parameters?

Pessimistress: Map props will be typed following the options expected by the `mapLib.Map` constructor.

chrisgervang:
I think it'd be useful to print a warning in a dev env that links to the upgrade guide if maplibre is detected on the mapbox version.

Pessimistress:
As of now if you use the wrong mapLib Typescript will throw an error. If you are using JS without types then the two endpoints are identical.

docs/api-reference/map.md
```diff
@@ -26,96 +26,99 @@ function App() {
...
 
 
 ## Properties
 
+Aside from the props listed below, the `Map` component supports all parameters of the [Map](https://docs.mapbox.com/mapbox-gl-js/api/map/#map-parameters) constructor. Beware that this is not an exhaustive list of all props. Different base map libraries may offer different options and default values. When in doubt, refer to your base map library's documentation.
+
 ### Layout options
 
-#### `id`: string {#id}
+#### `id`: string {#id}
```
chrisgervang, line 62: I'm seeing a bunch of edits like this that don't have a visible change.. just checking if its fine?

src/components/popup.ts
```diff
@@ -110,23 +79,20 @@ function Popup(props: PopupProps) {
     applyReactStyle(popup.getElement(), props.style);
   }, [props.style]);
 
+  useImperativeHandle(ref, () => popup, []);
```
chrisgervang, line 82: Is this a functional change?

src/mapbox/create-ref.ts
```diff
@@ -46,12 +48,20 @@ export default function createRef(mapInstance: Mapbox, mapLib: any): MapRef {
...
     unproject: (point: PointLike) => {
-      return mapInstance.transform.pointLocation(mapLib.Point.convert(point));
+      const tr = map.transform;
+      map.transform = mapInstance.transform;
+      const result = map.unproject(point);
+      map.transform = tr;
+      return result;
     },
-    queryTerrainElevation: (lnglat: LngLatLike, options: ElevationQueryOptions) => {
+    queryTerrainElevation: (lnglat: LngLatLike, options: any) => {
```
chrisgervang, line 64: Is there no good type for this?

src/mapbox/create-ref.ts
```diff
@@ -46,12 +48,20 @@ export default function createRef(mapInstance: Mapbox, mapLib: any): MapRef {
...
+      map.transform = tr;
+      return result;
     },
     unproject: (point: PointLike) => {
-      return mapInstance.transform.pointLocation(mapLib.Point.convert(point));
+      const tr = map.transform;
+      map.transform = mapInstance.transform;
+      const result = map.unproject(point);
+      map.transform = tr;
+      return result;
```
chrisgervang, line 62: Is this a functional change?

docs/upgrade-guide.md
```diff
@@ -1,14 +1,35 @@
...
 
-```jsx
-<Map mapLib={import('mapbox-gl')} ...>
+```tsx title="map-v7.0.tsx"
+import Map from 'react-map-gl';
+import maplibregl from 'maplibre-gl';
+
+function App() {
+  return <Map
+    mapLib={import('mapbox-gl')}
```
chrisgervang, line 13: Is this supposed to be `mapLib={maplibregl}`? A before/after

docs/api-reference/map.md
Pessimistress: The id field is auto generated

src/mapbox/create-ref.ts
Pessimistress: This is where mapbox and maplibre diverge. I prefer not to type it in a way that we need to upgrade whenever they make a new release.

chrisgervang (approved):

docs/api-reference/map.md
Pessimistress: I will make another pass on all the documentation.

PR #2180, opened 2023-05-25
Migrate website to Docusaurus

Pessimistress:
Staged at: https://pessimistress.github.io/react-map-gl/

#### Todo
- [ ] Search


docs/upgrade-guide.md
```diff
@@ -62,7 +62,7 @@ v7 is a complete rewrite of the library. It is redesigned to be fast, lightweigh
 
 - `MapContext` is now an official API. The experimental `_MapContext` export will be removed in a future release.
 - `react-virtualized-auto-sizer` is no longer a dependency.
-- Inertia has been enabled by default on the map controller. To revert to the behavior in previous versions, set the [interaction options](/docs/api-reference/interactive-map.md#interaction-options):
+- Inertia has been enabled by default on the map controller. To revert to the behavior in previous versions, set the [interaction options](https://github.com/visgl/react-map-gl/tree/6.1-release/docs/api-reference/interactive-map.md#interaction-options):
```
chrisgervang, line 65: Is the idea that older deprecated pages are available as markdown, but no longer needed on the website?

website/ocular-docusaurus-plugin/index.js
```diff
@@ -0,0 +1,51 @@
...
+  context,
+  opts = {
+    resolve: {modules: [], alias: {}},
+    debug: false,
+    module: {},
+    plugins: []
+  }
+) {
+  return {
+    name: 'ocular-docusaurus-plugin',
```
chrisgervang, line 22: What's the connection to ocular? Is it involved in building?

website/src/examples/layers.mdx
```diff
@@ -0,0 +1,5 @@
+# Dynamic Styling
+
+import App from 'website-examples/layers/src/app';
```
chrisgervang, line 3: nit: rename this example folder to `styling` so its connected to the title?

chrisgervang (approved):

docs/upgrade-guide.md
Pessimistress: Docusaurus fails to build if you have invalid links. In this case, the links were written according to the directory structure of an earlier release, and the files no longer exist.

website/ocular-docusaurus-plugin/index.js
Pessimistress: The plan is to eventually extract it to a standalone module that we can share across multiple repos.

website/src/examples/layers.mdx
Pessimistress: We often link to the examples in issues/discussions and renaming will break them.

docs/upgrade-guide.md
chrisgervang: Ah, ok good to know. Sounds like we could either link to the repo in all what's new and upgrade guides going forward, or let the build tool tell us when it's time to switch it from the nicer website experience to the repo.

PR #2192, opened 2023-06-02
Update examples for v7.1

Pessimistress:


examples/get-started/basic/app.jsx
```diff
@@ -2,7 +2,6 @@
 import * as React from 'react';
 import {createRoot} from 'react-dom/client';
 import Map, {Marker} from 'react-map-gl';
```
chrisgervang, line 4: I'm thinking about maplibre users looking for examples..

It could be helpful in each of these to mention maplibre can be used instead since that most (if not all) are compatible with it. I think they'd just need to change the import and apiKey/style prop.

chrisgervang (approved):

PR #2213, opened 2023-06-23
Add className prop support to Marker

marexandre:
When having multiple types of markers on the map that have hierarchy rules to them which requires setting different `z-index` being able to pass a different `className` to the Marker component is necessary.

```css
.secondary {
  z-index: 1;
}

.primary {
  z-index: 2;
}
```

```jsx
<Map>
  <Marker className="secondary">Secondary</Marker>
  <Marker className="primary">Primary</Marker>
</Map>
```

Considering on making a PR to `mapbox-gl` as well to update the Marker to have similar className option like the Popup component, but it felt it might be easier to get this update into react-map-gl first.
Edit: Stared a PR https://github.com/mapbox/mapbox-gl-js/pull/12770 - will see what mapbox team say...
Edit2:  Started a PR https://github.com/maplibre/maplibre-gl-js/pull/2729 for `maplibre-gl-js` for consistency between libs

### DIFF
--- a/src/components/marker.ts
+++ b/src/components/marker.ts
@@ -22,6 +22,7 @@ export type MarkerProps<OptionsT, MarkerT extends MarkerInstance> = OptionsT & {
   rotation?: number;
   rotationAlignment?: string;
   popup?: any;
+  className?: string;
 
   /** CSS style override, applied to the control's container */
   style?: React.CSSProperties;
@@ -32,6 +33,11 @@ export type MarkerProps<OptionsT, MarkerT extends MarkerInstance> = OptionsT & {
   children?: React.ReactNode;
 };
 
+// Adapted from https://github.com/mapbox/mapbox-gl-js/blob/v1.13.0/src/ui/popup.js
+function getClassList(className: string) {
+  return new Set(className ? className.trim().split(/\s+/) : []);
+}
+
 /* eslint-disable complexity,max-statements */
 function Marker<MarkerOptions, MarkerT extends MarkerInstance>(
   props: MarkerProps<MarkerOptions, MarkerT>,
@@ -40,6 +46,7 @@ function Marker<MarkerOptions, MarkerT extends MarkerInstance>(
   const {map, mapLib} = useContext(MapContext);
   const thisRef = useRef({props});
   thisRef.current.props = props;
+  const classNameRef = useRef(null);
 
   const marker: MarkerT = useMemo(() => {
     let hasChildren = false;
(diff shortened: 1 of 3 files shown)


src/components/marker.ts
```diff
@@ -32,6 +33,11 @@ export type MarkerProps<OptionsT, MarkerT extends MarkerInstance> = OptionsT & {
   children?: React.ReactNode;
 };
 
+// Adapted from https://github.com/mapbox/mapbox-gl-js/blob/v1.13.0/src/ui/popup.js
+function getClassList(className: string) {
+  return new Set(className ? className.trim().split(/\s+/) : []);
+}
```
marexandre, line 39: Copied this over from Popup component - can add `src/utils/get-class-list.ts` and use that util in both Marker and Popup component. Let me know what you think is best.

Pessimistress:
If you are going to propose this upstream to mapbox-gl, then I prefer to let the API change happen there. A major design decision we made in the current release track (v7.x) is to reduce API discrepancy between the React components and the underlying map library.  `react-map-gl` aims to be compatible with a range of versions and forks, not just the latest release of mapbox-gl. While I have no strong objection about this particular change, my main concern is that the more surface area we create between the wrapper and the base library, the more likely it will break in the future.

Also FYI, in the upcoming v7.1 you will be able to create a ref on `Marker` which allows you to access the div via `ref.current.getElement()`.

marexandre:
> A major design decision we made in the current release track (v7.x) is to reduce API discrepancy between the React components and the underlying map library

Think this was the best decision and one big reason we are looking to potentially switching to use `react-map-gl`.

> `react-map-gl` aims to be compatible with a range of versions and forks, not just the latest release of mapbox-gl. 

Wondering how this will get handled long term... both `mapbox-gl-js` and `maplibre-gl-js` are getting updates and slowly will get out of sync and in that case will `react-map-gl` be reflecting those changes as they come in, or will it mostly try to stay neutral to keep compatible with a big range of versions and in some cases not have latest features or updates?

I'm happy to respect your judgement on this PR and if you feel like it's not worth adding `className` prop, I'm happy to close this  PR as while `ref.current.getElement()` is not as simple compared to passing one prop it allows us to do what we need to.

Pessimistress:
Regarding the divergence of `mapbox-gl` and `maplibre-gl`, v7.1 will split the import endpoint (`react-map-gl/maplibre`) and allow for independent component implementations. In the long term, we may spin out maplibre support to a separate library when the differences become too significant.

I encourage you to open a PR on mapbox-gl, and meanwhile try the ref approach. We can keep this PR open in case that doesn't work out for you.

marexandre:
Sounds good and makes sense. Will try out `v7.1`.

PR to `mapbox-gl-js` has been open
- https://github.com/mapbox/mapbox-gl-js/pull/12770

PR to `maplibre-gl-js` has been made and merged
- https://github.com/maplibre/maplibre-gl-js/pull/2729

Issue #2228, opened 2023-07-13
[Bug] Layers don't render on certain map styles from v7.1


HelloTiago:
### Description

First of all, just wanted to say thanks for your work on this, it's been very helpful!

After updating to react-map-gl 7.1.2 I noticed that layers added using the `<Layer>` component stopped showing up on the map when using certain map styles.

For example with `mapbox://styles/mapbox/streets-v9` everything loads perfectly but when I use `mapbox://styles/mapbox/streets-v12` or `mapbox://styles/mapbox/light-v11` the layers don't show up anymore.

v7.0.21 and earlier versions don't have this problem.

### Expected Behavior

The layers should show up on every map style.

### Steps to Reproduce

- Open https://codesandbox.io/s/busy-cherry-p9jtr9?file=/src/index.js
- Set the access token
- Switch between `streets-v12` and `streets-v9` and notice the "Mapbox DC" / "Mapbox SF" labels on the map.

### Environment

- Framework version: react-map-gl@7.1.2
- Map library: mapbox-gl@2.15.0
- Browser: Brave Version 1.52.130 Chromium: 114.0.5735.198 (arm64)
- OS: macOS Ventura 13.4.1

Pessimistress:
v7.1 removed default props on `<Map>` and always use the base library's defaults.
streets-v12 uses globe projection, see https://docs.mapbox.com/mapbox-gl-js/example/simple-map/
There is a standing bug where symbols are not rendered correctly in globe mode at certain zoom levels.

You can work around it by explicitly setting `projection="mercator"` on the map.

HelloTiago:
Setting the projection worked perfectly. Thank you for the fast response!

Pessimistress:
Fixed in v7.1.3.

Issue #2136, opened 2023-03-05
[Bug] Circle layers have issues with globe projection


sh54:
### Description

When the map projection is set to `globe` then I am experiencing two issues.

This occurs with a layer of `type=circle` with `paint.circle-radius = 20`.

Firstly when zooming in and out the layer pops in and out of visibility.

Secondly I am not able to find the feature via `queryRenderedFeatures` with a point argument. But without the point argument it gets found.

I guess this is probably an upstream issue but I have an example ready to go here with React.

### Visibility during zoom issue:

https://user-images.githubusercontent.com/906180/222965151-8c4afa38-82dd-479a-abf7-e7f5254185e0.mp4

### queryRenderedFeatures issue:

https://user-images.githubusercontent.com/906180/222965218-58eda007-2306-4c6d-a1ed-5cf04df38e40.mp4

### Expected Behavior

If I am in the default projection then neither of these issues happen. The circle remains visible. The circle is always findable with ~queryRenderedFeatures~.

### Steps to Reproduce

See [example](https://stackblitz.com/edit/react-ts-x1fxmh?file=MapView.tsx). Need to use your own mapbox api token.

### Environment

- Framework version: react-map-gl@7.0.21
- Map library: mapbox-gl@2.12.1
- Browser: Chrome
- OS: macOS

dcyoung:
+1, we have encountered this issue and hoping for a fix or workaround

IliyanID:
+ 1, we've also stumbled onto this issue. A fix would be appreciated

jehie:
This issue seems to relate to the  this._renderTransform and the map.transform. For some reason the map instance transform has still globe set as the projection while the shadowTransform has the correct mercator projection (when zoomed in enough for the switch to happen).  So somehow the states are not synced properly, and when querying the features there is a mismatch between the point projection (globe) and the sourceCaches projection (mercator), while they should be both mercator at this zoom level.

Any ideas @Pessimistress how to solve this?

jehie:
@Pessimistress, thanks for the fix. I was able to verify on our side that the fix solves the queryRenderedFeatures problem and also circles are not disappearing anymore.

PR #2229, opened 2023-07-13
Fix globe/mercator projection transition

### LINKED ISSUE #1930: [Bug] Layers not visible in globe projection with initial zoom > 6
### Description

Initial zoom greater than 6 does not show layers unless we zoom out and then back in when using the globe projection

### Expected Behavior

Layers rendered

### Steps to Reproduce

https://codesandbox.io/s/eloquent-curran-369y7k

### Environment

- Framework version:"7.0.18"
- Map library:"^2.9.2"
- Browser:Chrome
- OS:macOS 12.3.1

### Logs

_No response_

Pessimistress:
Attempt to fix #1930, #2136,  #2228

### DIFF
--- a/src/mapbox/mapbox.ts
+++ b/src/mapbox/mapbox.ts
@@ -1,4 +1,9 @@
-import {transformToViewState, applyViewStateToTransform, cloneTransform} from '../utils/transform';
+import {
+  transformToViewState,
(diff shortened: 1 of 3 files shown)


chrisgervang (approved):
This is more of a rubber stamp since I'm not familiar with the projection system.

PR #2235, opened 2023-07-18
Docusaurus: Fix `editUrl`

tordans:
This was still pointing to deck.gl, so the edit links on for example https://visgl.github.io/react-map-gl/docs/get-started did not work

`https://github.com/visgl/deck.gl/tree/master/docs/get-started/get-started.md`

I did not test this new URL, but I think it should work …


Pessimistress (approved):
Thank you!

PR #2238, opened 2023-07-18
Docs: Start updating the types docs after 7.1

tordans:
This is a start to update the docs after the 7.1 release. 

This could be merged as is (after review…), but it is not finished, yet. Other parts of those pages need to be updated. 

However, I was hoping someone else could jump in and help.
And also wanted to wait for a green light that this is going into the right direction.

Closes https://github.com/visgl/react-map-gl/issues/2237

### DIFF
--- a/docs/api-reference/types.md
+++ b/docs/api-reference/types.md
@@ -1,16 +1,20 @@
 # Types
 
-The following types can be imported from `react-map-gl` when using TypeScript.
+The following types can be imported from `react-map-gl` | `react-map-gl/maplibre` when using TypeScript.
 
 ## Components
 
 #### MapboxMap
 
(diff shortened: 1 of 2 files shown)


docs/api-reference/types.md
```diff
@@ -263,18 +267,22 @@ A source containing vector tiles in [Mapbox Vector Tile format](https://docs.map
 
 #### MapboxEvent
 
+_Deprecated as of v7.1_ — Use [MapEvent](#mapevent) instead.
```
tordans, line 270: My idea is, to keep the `#mapboxevent` hash url because cool URLs don't break.

Pessimistress:
Closing in favor of #2263

Issue #1798, opened 2022-03-08
[Bug] Incorrect types causing build to fail while using Vite


DennisMartinez:
### Description

Hello! I'm experiencing an issue using react-map-gl with a Vite generated project. When I build my project, I'm presented with the following error:

```
node_modules/react-map-gl/dist/esm/components/use-map.d.ts:13:5 - error TS2411: Property 'current' of type 'MapRef | undefined' is not assignable to 'string' index type 'MapRef'.

13     current?: MapRef;
````

This doesn't seem to happen using CRA and I can remedy this by swapping `skipLibCheck` to false in the tsconfig, but that doesn't seem correct. I'm unsure if this is a TypeScript issue, Vite issue, or a bug related to this library.

I've created a test repo here: https://github.com/DennisMartinez/vite-react-map-gl-bug

### Expected Behavior

To build without type errors.

### Steps to Reproduce

Pull the test repo
Create a .env.local file and add your mapbox api token
```
VITE_APP_MAPBOX_TOKEN=
```
Install the dependencies with `yarn install`
Build the project with `yarn build`
Observe the error in the console

### Environment

- Framework version: react@17.0.2, react-dom@17.0.2
- Map library: react-map-gl@7.0.9, mapbox-gl@2.7.0
- Browser: N/A
- OS: macOS Monterey 12.0.1 (21A559)

The app was generated using vite react-ts.

### Logs

Terminal error while building

```
node_modules/react-map-gl/dist/esm/components/use-map.d.ts:13:5 - error TS2411: Property 'current' of type 'MapRef | undefined' is not assignable to 'string' index type 'MapRef'.

13     current?: MapRef;
```

**Comment by Pessimistress** [comment]
Seems that the string indexed fields must be more generic than the specific ones: https://www.typescriptlang.org/docs/handbook/interfaces.html#indexable-types

**Comment by Pessimistress** [comment]
Published 7.0.10

**Comment by trumbitta** [comment]
Hi 👋 

I'm having the same issue in two places:

```
node_modules/react-map-gl/dist/esm/components/use-map.d.ts:17:5 - error TS2411: Property 'current' of type 'MapRef<MapT> \| undefined' is not assignable to 'string' index type 'MapRef<MapT>'.
--
  |  
  | 17     current?: MapRef<MapT>;
  | ~~~~~~~
  |  
[...]

PR #2250, opened 2023-08-09
Fix typechecking (#1798)

trumbitta:
Fixes a resurgence of the bug found in #1798

It might be a breaking change because it forces code like this:

```ts
  const draw = useControl<MapboxGlDraw>(
    () => new MapboxGlDraw(rest),
    ({ map }: MapContextValue<MapInstance>) => {
      map.on("draw.create", onCreate);
    },
    ({ map }: MapContextValue<MapInstance>) => {
      map.off("draw.create", onCreate);
    },
    {
      position,
    },
  );
```

to become:

```ts
  const draw = useControl<MapboxGlDraw>(
    () => new MapboxGlDraw(rest),
    ({ map }: MapContextValue<MapInstance>) => {
      map?.on("draw.create", onCreate);
    },
    ({ map }: MapContextValue<MapInstance>) => {
      map?.off("draw.create", onCreate);
    },
    {
      position,
    },
  );
```

### DIFF
--- a/src/components/use-map.tsx
+++ b/src/components/use-map.tsx
@@ -52,7 +52,10 @@ export const MapProvider: React.FC<{children?: React.ReactNode}> = props => {
   );
 };
 
-export function useMap<MapT extends MapInstance>() {
+export function useMap<MapT extends MapInstance>(): {
+  [id: string]: MapRef<MapT> | undefined;
+  current?: MapRef<MapT>;
+} {
   const maps = useContext(MountedMapsContext)?.maps;
   const currentMap = useContext(MapContext);
 


Pessimistress:
That does not look like the right solution. If you need to make something optional, you add `| undefined` to where you use it, not in its type definition.

Can you create a project that reproduces your issue?

trumbitta:
> That does not look like the right solution. If you need to make something optional, you add `| undefined` to where you use it, not in its type definition.
> 
> Can you create a project that reproduces your issue?

It's the same issue as #1798, although happening with plain `tsc` instead of Vite, and it's also the same fix as https://github.com/visgl/react-map-gl/pull/1799

I'll try to put together a repro 🙏

trumbitta:
@Pessimistress here's the repro!

https://codesandbox.io/p/sandbox/wispy-thunder-tx58mq

Run `yarn tsc` or `yarn start` if it doesn't start automatically.

Pessimistress:
> it's also the same fix as https://github.com/visgl/react-map-gl/pull/1799

No, it's not the same fix. You need to change it where MapRef is **used**, not where it is **defined**.

Looks like TypeScript compiler is not inferring the return type of `useMap` correctly. You can enforce it like this:

```patch
--- a/src/components/use-map.tsx
+++ b/src/components/use-map.tsx
@@ -52,7 +52,10 @@ export const MapProvider: React.FC<{children?: React.ReactNode}> = props => {
   );
 };

-export function useMap<MapT extends MapInstance>() {
+export function useMap<MapT extends MapInstance>(): {
+  [id: string]: MapRef<MapT> | undefined;
+  current?: MapRef<MapT>;
+} {
   const maps = useContext(MountedMapsContext)?.maps;
   const currentMap = useContext(MapContext);
```

trumbitta:
@Pessimistress It kind of makes sense, but I don't think I know how to test this kind of change in this repo.
My change I could test locally by editing the definition files in `node_modules`, but this seems to require first compiling, then other steps I'm not familiar with.

Would you mind fixing this yourself the way you suggested? 🙏

Pessimistress (approved):

jscheid:
@Pessimistress thanks for the merge! Any chance for a new release with this fix?

PR #2268, opened 2023-08-31
fix: add missing exports for subpath module

khmm12:
Subpath exports should be defined via package.json exports field for ESM.

 - https://webpack.js.org/configuration/module/#resolvefullyspecified
 - https://www.typescriptlang.org/docs/handbook/esm-node.html
 - https://nodejs.org/api/packages.html#subpath-exports

It's still not ESM compliant, because inside CJS packages ESM modules should have extension `.mjs` and all their imports should be fully specified `import './events.mjs'`, not `import './events'`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,28 @@
   "types": "dist/esm/index.d.ts",
   "main": "dist/es5/index.js",
   "module": "dist/esm/index.js",
+  "exports": {
+    ".": {
+      "import": {
+        "types": "./dist/esm/index.d.ts",
+        "default": "./dist/esm/index.js"
+      },
+      "require": {
+        "types": "./dist/es5/index.d.ts",
+        "default": "./dist/es5/index.js"
+      }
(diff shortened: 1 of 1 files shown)


tordans:
@khmm12 I just created https://github.com/visgl/react-map-gl/issues/2272 to give this a bit more visibility and document my failed attempts at workarounds. Did you get this working for you?

khmm12:
@tordans Yes, it works well with bundlers like webpack. But as I mentioned in the description the package is still not ESM compliant, therefore it's not compatible with strict environments like node.js. The next step should be ESM compliant package.

tordans:
@khmm12 do you have experience with how to best do this? I am just reading about https://deno.com/blog/publish-esm-cjs-module-dnt which sounds nice but is based on Deno as an entry point…

Pessimistress:
Closing in favor of #2281

PR #2270, opened 2023-09-03
fix: updated navcontrol documentation to include style import, otherw…

goharShoukat:
…ise controls do not show

### DIFF
--- a/docs/api-reference/navigation-control.md
+++ b/docs/api-reference/navigation-control.md
@@ -12,6 +12,7 @@ import TabItem from '@theme/TabItem';
 ```tsx
 import * as React from 'react';
 import Map, {NavigationControl} from 'react-map-gl';
+import 'mapbox-gl/dist/mapbox-gl.css';
 
 function App() {
   return <Map


Pessimistress:
There are multiple ways to include the CSS stylesheet, see https://visgl.github.io/react-map-gl/docs/get-started#styling. Importing CSS in JavaScript is not by default supported by all build tools.

With that said, I'm supportive of adding some reminder in the docs, along the lines of

```md
Note that this component requires the inclusion of the [CSS stylesheet](https://visgl.github.io/react-map-gl/docs/get-started#styling) to work.
```

If we are going to add this, all components (Map, *Control) should get the same update.

goharShoukat:
> There are multiple ways to include the CSS stylesheet, see https://visgl.github.io/react-map-gl/docs/get-started#styling. Importing CSS in JavaScript is not by default supported by all build tools.
> 
> With that said, I'm supportive of adding some reminder in the docs, along the lines of
> 
> ```
> Note that this component requires the inclusion of the [CSS stylesheet](https://visgl.github.io/react-map-gl/docs/get-started#styling) to work.
> ```
> 
> If we are going to add this, all components (Map, *Control) should get the same update.

Just to confirm, do you want me to add that to all the relevant components? I do think this reminder is needed because is pent quite some time fiddling around with the code only to figure out this particular line does the trick.

Pessimistress:
Yep go ahead.

Issue #2196, opened 2023-06-06
[Bug] When 3D terrain is enabled, the clickable area of a line layer is not aligned its visual representation


dstantwo:
### Description

I have a small demo to reproduce this issue where I'm displaying the Half Dome hiking trail as a geojson source/layer on a map with 3d terrain enabled. I've made the layer clickable with the `interactiveLayerIds` prop.

With `react-map-gl`, the line layer renders, but the clickable area of the line is not aligned with the visual representation of the layer. You can see in the video below that the cursor turns into a pointer some distance away from the line.
🔗  Here's a code sandbox link to this `react-map-gl` demo (you'll have to add your own Mapbox token in `App.js`) https://codesandbox.io/s/react-map-gl-terrain-layer-bug-t5p3vw?file=/src/App.js

![642D09F0-7123-4951-AF83-5E5D48376586](https://github.com/visgl/react-map-gl/assets/14078943/39728010-1260-41f1-8122-39d869aba07a)

For comparison, a nearly identical setup that uses `mapbox-gl` directly (without using `react-map-gl`), works as expected so that the clickable area of the line aligns with the visual representation. 
![A8F1E6A3-7C4E-47F6-9AC6-6432AEEA6701](https://github.com/visgl/react-map-gl/assets/14078943/86d64c08-c21b-45ab-8a89-49413cc186ac)
🔗  Here's a code sandbox link to this `mapbox-gl` only demo (again, you'll have to add your own Mapbox token in `App.js`) https://codesandbox.io/s/mapbox-gl-js-terrain-layer-working-5f8vm2?file=/src/App.js

### Expected Behavior

I would expect the clickable area of a geojson layer to align with the layer's visual representation in `react-map-gl` as it does in plain `mapbox-gl`

### Steps to Reproduce

* A `react-map-gl` demo (you'll have to add your own Mapbox token in `App.js`), which shows the broken behavior https://codesandbox.io/s/react-map-gl-terrain-layer-bug-t5p3vw?file=/src/App.js
* A nearly identical setup using only `mapbox-gl` that works as expected https://codesandbox.io/s/mapbox-gl-js-terrain-layer-working-5f8vm2?file=/src/App.js

### Environment

- Framework version: 
    - react-map-gl@7.0.21 (in code sandbox demos)
    - Also reproduced in my full project with react-map-gl@7.0.25
- Map library: mapbox-gl@2.13.0
- Browser: Chrome
- OS: MacOS 12.6.3

[...]

PR #2280, opened 2023-09-11
Use shadow transform for queryRenderedFeatures call

Pessimistress:
For #2196


chrisgervang (approved):
Interesting edge case. Do you think this also cover's the Marker case someone reported?

Pessimistress:
> Do you think this also cover's the Marker case someone reported

It might be a similar issue, but I need to investigate more.

Issue #2272, opened 2023-09-04
[Bug] Release ESM compliant package (was: Does not build with AstroJS)


tordans:
### Description

Using ReactMapGL within an AstroJS Project (https://docs.astro.build/) does work great in Dev mode but fails on build `npm run build`.

It looks like https://github.com/visgl/react-map-gl/pull/2268 is partial solution but the `*.mjs` part that was mentioned in this PR is still required to get it working.

---

## Test 1: `import ReactMapGl, { NavigationControl } from 'react-map-gl/maplibre'`

Using the standard import does result is errors like this:

CODE: https://github.com/tordans/berlin-bikenetwork-monitoring/blob/0704fe8a517e75669154e1bf2a9e5ec0ca810da8/src/components/Map/Map.tsx

```
generating static routes 
 error   Directory import '/home/runner/work/berlin-bikenetwork-monitoring/berlin-bikenetwork-monitoring/node_modules/react-map-gl/maplibre' is not supported resolving ES modules imported from /home/runner/work/berlin-bikenetwork-monitoring/berlin-bikenetwork-monitoring/dist/chunks/pages/index.astro.40918a6a.mjs
  Did you mean to import react-map-gl/dist/es5/exports-maplibre.js?
Error [ERR_UNSUPPORTED_DIR_IMPORT]: Directory import '/home/runner/work/berlin-bikenetwork-monitoring/berlin-bikenetwork-monitoring/node_modules/react-map-gl/maplibre' is not supported resolving ES modules imported from /home/runner/work/berlin-bikenetwork-monitoring/berlin-bikenetwork-monitoring/dist/chunks/pages/index.astro.40918a6a.mjs
Did you mean to import react-map-gl/dist/es5/exports-maplibre.js?
    at new NodeError (node:internal/errors:405:5)
    at finalizeResolution (node:internal/modules/esm/resolve:317:17)
    at moduleResolve (node:internal/modules/esm/resolve:943:10)
    at defaultResolve (node:internal/modules/esm/resolve:1129:11)
    at nextResolve (node:internal/modules/esm/loader:163:28)
    at ESMLoader.resolve (node:internal/modules/esm/loader:835:30)
    at ESMLoader.getModuleJob (node:internal/modules/esm/loader:424:18)
    at ModuleWrap.<anonymous> (node:internal/modules/esm/module_job:77:40)
[...]

PR #2281, opened 2023-09-12
Switch to ESM module

Pessimistress:
For #2272
Targeting v7.2


test/src/components/map.spec.jsx
```diff
@@ -107,14 +107,13 @@ test('Map#controlled#no-update', async t => {
   });
 });
 
-test('Map#controlled#mirrow-back', async t => {
-  t.plan(5);
+test('Map#controlled#mirror-back', async t => {
```
chrisgervang, line 110: Fixing an unrelated typo?

test/src/utils/react-dom-mock.js
```diff
@@ -0,0 +1,5 @@
+import * as React from 'react';
+
+export function createPortal(element, container) {
```
chrisgervang, line 3: Hoping to learn what this does. Where is this used now? I only see the usage getting deleted.

chrisgervang (approved):

test/src/utils/react-dom-mock.js
Pessimistress: This file is used as an alias for `react-dom` during node tests, configured in ocularrc. Before we were hijacking the function with `require('react-dom').createPortal = createPortalMock`, but that no longer works in the ESM (import/export) world.

chrisgervang: Oh, ok got it. Could you comment that it's aliased in ocularrc?

PR #2308, opened 2023-10-28
Example Interaction: Add URL hash

tordans:
This make the zoom visible in https://visgl.github.io/react-map-gl/examples/interaction

### DIFF
--- a/examples/interaction/src/app.tsx
+++ b/examples/interaction/src/app.tsx
@@ -46,6 +46,7 @@ export default function App() {
         {...settings}
         mapStyle="mapbox://styles/mapbox/dark-v9"
         mapboxAccessToken={MAPBOX_TOKEN}
+        hash
       />
       <ControlPanel settings={settings} onChange={updateSettings} />
     </>


matburnham:
Is this trying to add the map's position to its page's location hash? I can't seem to get the same to work with react-map-gl@7.1.6.

According to https://github.com/visgl/react-map-gl/issues/665 this was previously done with `mapOptions={{ hash: true }}`. But the [upgrade guide for v.7.0](https://visgl.github.io/react-map-gl/docs/upgrade-guide#map) shows _"mapOptions is removed. Almost all of the options for the native Map class are exposed as props."_ That said, I can't find reference to `hash` in the current codebase, and it's not documented in the [Map API reference](https://visgl.github.io/react-map-gl/docs/api-reference/map).

Should this work?

tordans:
> That said, I can't find reference to hash in the current codebase, and it's not documented in the Map API reference.

I assume that is because of this...

> Aside from the props listed below, the Map component supports all parameters of the Map class constructor (Mapbox | Maplibre). Beware that this is not an exhaustive list of all props. Different base map libraries may offer different options and default values. When in doubt, refer to your base map library's documentation.

Meaning, the docs does not list all props that are just passed to the map library. I was confused by this myself, before.

matburnham:
And it does work 🤦. I was just being an idiot. I was running my previously compiled code with `yarn run start` rather than `yarn run dev` so nothing was ever changing.

mohasinkr:
> And it does work 🤦. I was just being an idiot. I was running my previously compiled code with `yarn run start` rather than `yarn run dev` so nothing was ever changing.

Happens to the best of us buddy!

Issue #2310, opened 2023-11-06
[Bug] `RTLTextPlugin` is eagerly loaded from mapbox.com when using MapLibre and raising privacy concerns


KiwiKilian:
### Description

The `RTLTextPlugin` prop is set to `https://api.mapbox.com/mapbox-gl-js/plugins/mapbox-gl-rtl-text/v0.2.3/mapbox-gl-rtl-text.js` per default. This is the same default for both maplibs Mapbox and MapLibre. Furthermore [the docs](https://visgl.github.io/react-map-gl/docs/api-reference/map#rtltextplugin) state:

> Setting this prop is the equivelant of calling [mapboxgl.setRTLTextPlugin](https://docs.mapbox.com/mapbox-gl-js/api/properties/#setrtltextplugin) with lazy: true.

But this doesn't align with the currently implemented behavior, [which loads it eagerly](https://github.com/visgl/react-map-gl/blob/e11b41eefe73ff773bbec1db73277e4aca1f1bef/src/utils/set-globals.ts#L54).

To conclude this results in eager loading with both Mapbox and MapLibre of the `RTLTextPlugin` from the mapbox.com domain. This seems totally counterintuitive when using MapLibre. Especially when viewing from a privacy standpoint. Currently judicial ruling in the EU requires for consent of users before sending requests to US providers from an EU based website.

Therefore I don't expect a library like `react-map-gl` to fetch from a remote domain I didn't specify myself. Especially when using the MapLibre maplib. Currently it's necessary to set `RTLTextPlugin={null}`, to prevent from loading the plugin. Which is also not straight-forward as the types do not supporting disabling the `RTLTextPlugin`.

### Expected Behavior

This might be subject to personal opinion again, but I would like to see a breaking change, which removes the `RTLTextPlugin` default url.

Regardless of this, lazy loading should be enabled or even allow further customization of the loading via the prop like:
```tsx
RTLTextPlugin={{
  url: "...",
  lazy: true,
}}
```

It should also be possible to disable `RTLTextPlugin`, if a default is kept.

### Steps to Reproduce

Check the network tab of https://codesandbox.io/s/react-map-gl-maplibre-hfpr6l?file=/src/index.js for resources from `mapbox.com`.

### Environment

- Framework version: 7.1.6
- Map library: maplibre-gl 3.5.2
- Browser: Arc 1.15.1 Chromium 119
- OS: macOS

Pessimistress:
Thanks for raising this.

[...]

PR #2312, opened 2023-11-09
Allow disabling RTLTextPlugin and load lazy

KiwiKilian:
As discussed in #2310 this is the patch which fixes the following for the `RTLTextPlugin` prop:

1. Load lazy
2. Allow the prop to be set to `false` to disable the current default load


Pessimistress (approved):

KiwiKilian:
@Pessimistress when are you planning on merging? Afterwards I would setup a the breaking change MR.

Pessimistress:
Hey sorry I've been busy. Will publish a patch today.

PR #2313, opened 2023-11-13
Update Demo in README to show importing mapbox styles.

lacymorrow:
The readme demo does not import the mapbox styles, which causes CSS errors like the mentioned issue.

Closes #2216

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,6 +29,9 @@ npm install --save react-map-gl mapbox-gl
 import * as React from 'react';
 import Map from 'react-map-gl';
 
+// MapBox styles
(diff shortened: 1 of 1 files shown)


Pessimistress:
There are multiple ways to include the required css, see https://visgl.github.io/react-map-gl/docs/get-started#styling

Importing css in JS may not be supported by all bundlers.

lacymorrow:
Makes sense

lacymorrow:
Seeing that 5 more folks have had this issue since I did, it would probably be useful to mention this in the docs.

#2216

Issue #2248, opened 2023-08-03
Pointer event lat lng position doesn't factor in terrain


jonshutt:
### Description

In the example shown below, when clicking the map to add points, the markers appear correct until you move the map. It appears that however the mouse screen X Y position is converted to a lat lng coordinates is not factoring in terrain.

A similar issue happens with draggable markers, or clicking on the map to show features at that position.  

```
<Map
          initialViewState={{
            latitude: 32.6141,
            longitude: -114.34411,
            zoom: 14,
            bearing: 80,
            pitch: 80
          }}
          maxPitch={85}
          mapStyle="mapbox://styles/mapbox/satellite-v9"
          mapboxAccessToken={TOKEN}
          terrain={{source: 'mapbox-dem', exaggeration: 3}}
          onClick={event => {
            setPoints([...points, event.lngLat]);
          }}
        >
          <Source
            id="mapbox-dem"
            type="raster-dem"
            url="mapbox://mapbox.mapbox-terrain-dem-v1"
            tileSize={512}
            maxzoom={14}
          />
          <Layer {...skyLayer} />
          {points.map(point => (
            <Marker longitude={point.lng} latitude={point.lat} />
          ))}
        </Map>
```

### Expected Behavior

When moving the map, the position on the marker remains in the same place. Eg, the lat lng position is correct

### Steps to Reproduce

Using the terrain example, add a click event and markers

based on `terrain\src\app.tsx` in the repo

```
import React, {useEffect, useRef, useState} from 'react';

import {createRoot} from 'react-dom/client';
import Map, {Source, Layer, Marker} from 'react-map-gl';

import ControlPanel from './control-panel';

import type {SkyLayer} from 'react-map-gl';

const TOKEN = ''; // Set your mapbox token here

const skyLayer: SkyLayer = {
  id: 'sky',
  type: 'sky',
  paint: {
    'sky-type': 'atmosphere',
    'sky-atmosphere-sun': [0.0, 0.0],
    'sky-atmosphere-sun-intensity': 15
  }
};

export default function App() {
  const [points, setPoints] = useState([]);

  return (
    <>
        <Map
          initialViewState={{
            latitude: 32.6141,
            longitude: -114.34411,
            zoom: 14,
            bearing: 80,
            pitch: 80
          }}
          maxPitch={85}
          mapStyle="mapbox://styles/mapbox/satellite-v9"
          mapboxAccessToken={TOKEN}
[...]

PR #2329, opened 2023-12-19
Use shadow transform for all pointer events

jonshutt:
Mouse events seem to be returning locations without elevation. This handled for _queryRenderedFeatures, but other pointer events return the wrong location.

#2248

### DIFF
--- a/src/mapbox/mapbox.ts
+++ b/src/mapbox/mapbox.ts
@@ -625,18 +625,25 @@ export default class Mapbox<
   }
 
   _onPointerEvent = (e: MapMouseEvent<MapT> | MapMouseEvent<MapT>) => {
-    if (e.type === 'mousemove' || e.type === 'mouseout') {
-      this._updateHover(e);
-    }
-
-    // @ts-ignore
-    const cb = this.props[pointerEvents[e.type]];
-    if (cb) {
-      if (this.props.interactiveLayerIds && e.type !== 'mouseover' && e.type !== 'mouseout') {
-        e.features = this._hoveredFeatures || this._queryRenderedFeatures(e.point);
+    const map = this._map;
+    const tr = map.transform;
+    try {
+      map.transform = this._renderTransform;
+      if (e.type === 'mousemove' || e.type === 'mouseout') {
+        this._updateHover(e);
       }
-      cb(e);
-      delete e.features;
+
+      // @ts-ignore
+      const cb = this.props[pointerEvents[e.type]];
+      if (cb) {
+        if (this.props.interactiveLayerIds && e.type !== 'mouseover' && e.type !== 'mouseout') {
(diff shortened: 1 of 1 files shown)


kaij:
I also see this problem and tried this pull request against 7.2.0-beta.1 (using maplibre). It seems to work better with mouse events. However, markers (and 3d overlay layers as well) still jump around erratically. I think this makes react-map-gl currently unusable when having terrain and markers. Which is a pity since there is this nice plugin react-three-map.

Using maplibre 4.0.1 standalone (without react-map-gl), none of the problems occur: movement in heavy terrain is stable, markers and 3d overlay layers as well. 

It's possibly a hard-to-find, small fix in the transform sync (and maybe maplibre behaves differently than mapbox). Any idea?

jonshutt:
Hi Kai

I agree, the issue makes this library unusable with terrain and markers/clicking.

I'll have another look at this issue myself. I'm also wondering it `this._renderTransform` just isn't being updated when the terrain is added - so basically the library thinks the transform is flat, and mapbox knows it has terrain.

jonshutt:
I've been looking at this again, and I'm very sure what's happening is that after terrain source data is loaded, the mapbox transformation info changes to include this, and react-map-gl isn't updating it's shadow copy.

I've tried adding the following into the `_initialize` function
` map.on('sourcedata', (event) => {
      if (event.sourceId === 'mapbox-dem') {
        map.transform = this._renderTransform;
      }
    });`  
    
this does run multiple times as mapbox does keep loading more terrain data, but it does seem to fix most the problems. However, I noticed that camera jumps up and down when after moving over areas with terrain.


I tracked that down to `_updateViewState`, and commented out the lines that sets the camera elevation
eg, ` tr.cameraElevationReference = 'sea';`. I suspect this was to fix issues with the camera not being aware of the terrain due to having the wrong transform set

There's also various that seem to switch out the transform (`map.transform = `) while doing something, and set it back afterwards. I've commented all these out, I think they're all there as workarounds, but the cause is that transform is wrong.

chrisgervang:
@Pessimistress has published an early alpha of this new wrapper at https://github.com/visgl/react-maplibre 
We're still working out how we'll publish this new work before it comes out of alpha, but it would be great to hear if this helps the issue you're experiencing in maplibre.

nthh:
@jonshutt Have you been able to get the map working with terrain and markers? Specifically, I'm trying to implement draggable markers (similar to the example [here](https://visgl.github.io/react-map-gl/examples/mapbox/draggable-markers)) but have been running into this issue. I'm curious if a workaround would be possible with something like mapbox-gl-draw.

Pessimistress:
Closing this due to restructure of the code base.

A new proposal can be found in #2513

PR #2346, opened 2024-01-18
Docs: Fix maplibre link on attribution-control.md

tordans:
This is the right deep like. Ideally, we would also link to https://maplibre.org/maplibre-gl-js/docs/API/types/maplibregl.AttributionOptions/ which goes into details on the options. But the top level link might have to be enough…?


Pessimistress (approved):
Thank you!

chrisgervang:
Looks like the link changed again and was fixed by #2392

PR #2356, opened 2024-02-15
Add getCanvasContainer to types

georgemoon:
Both MapLibre GL and Mapbox GL have a function 'getCanvasContainer':

- https://github.com/mapbox/mapbox-gl-js/blob/bb51e3ba7cceef993628f6c85eef69435d902e4c/src/ui/map.js#L788
- https://github.com/maplibre/maplibre-gl-js/blob/3bea9a10acf741caf0913eb3792ef0a268f6fbe3/src/ui/map.ts#L2837

I would like to suggest that it be added to the types if possible. The attached is a simple pull request adding this type.

Thanks!

### DIFF
--- a/src/types/lib.ts
+++ b/src/types/lib.ts
@@ -77,6 +77,8 @@ export interface MapInstance extends Evented {
 
   getCanvas(): HTMLCanvasElement;
 
+  getCanvasContainer(): HTMLElement;
+
   remove(): void;
 
   triggerRepaint(): void;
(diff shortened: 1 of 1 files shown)


Pessimistress (changes requested):
This type is only used for internal type check. Map methods that are exposed to the user come from `@types/mapbox-gl` or `maplibre-gl`.

georgemoon:
The type was being exposed in the context of using "useControl" from "react-map-gl/maplibre".
I can see if I can do a minimal replication of the bug if that would be useful?

Pessimistress:
I see. Yes, that would be a bug.

Pessimistress (approved):

PR #2360, opened 2024-03-21
Fix changing zoom and center at once

tomhicks:
## What this does

Previously, if you wanted to zoom in and pan the map in the same viewport operation when you were very zoomed out, it would get "stuck" at latitudes near the equator.

If you're zoomed right out such that the equator is at the center of your viewport, and want to zoom in to a point at a much higher zoom level away from the equator, because the center change is handled first, the center change is constrained, so the latitude doesn't change. Then the zoom happens, and you end up at the wrong place.

By changing the order to zoom first, we ensure that the constraining that happens during the center change happens with the correct available center range.

## Testing

The unit tests still pass, but it's hard to exercise the exact scenario here because it's dependent on the viewport constraining behaviour which is mocked out in this repo.

### DIFF
--- a/src/utils/transform.ts
+++ b/src/utils/transform.ts
@@ -57,16 +57,16 @@ export function applyViewStateToTransform(tr: Transform, props: MapboxProps): bo
   const v: Partial<ViewState> = props.viewState || props;
   let changed = false;
 
+  if ('zoom' in v) {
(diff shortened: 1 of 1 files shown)


Pessimistress (approved):

Pessimistress:
Fixed by #2391

PR #2363, opened 2024-03-28
Update fullscreen-control source code link

Yvad60:
It appears that the link to the source code of `fullscreen-control` was updated to have `.tsx` extension instead of `.ts` currently the live will break


chrisgervang (approved):

chrisgervang:
Thank you!

PR #2367, opened 2024-04-05
Update marker.md

thorwebdev:
Value of type 'typeof Popup' is not callable. Did you mean to include 'new'?


thorwebdev:
Hm, no idea what happened to the headlines here. I only changed line 195 🤔

Pessimistress (approved):

PR #2374, opened 2024-04-10
chore(ci): Update to Node.js v22 in CI

donmccurdy:
PR #2369 fails in CI with error — 

> error puppeteer-core@22.6.3: The engine "node" is incompatible with this module. Expected version ">=18". Got "16.20.2"
error Found incompatible module.

This PR attempts to upgrade to Node v22, but if CI still fails I'll try downgrading to v20 or v18.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -21,9 +21,9 @@ jobs:
         run: git fetch --tags -f
 
       - name: Use Node.js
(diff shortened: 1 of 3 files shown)


DomeQdev (changes requested):
I'm requesting this change.

Pessimistress:
Already covered in monorepo refactor

PR #2386, opened 2024-05-10
Examples: Update to maplibre

tordans:
This updates the filter example to maplibre and removes the notes on mapbox.
I was actually testing something else and this was a by product.
Let me know if this would be useful and what would have to change.

The example itself is not as good as before because it always only highlights one state. But moving to the maplibre state example dataset was the quickest way to get this working again. It would be more ideal to have the mapbox dataset as open data as an example but I did not check that.


Pessimistress:
@tordans Do you mind rebase this to target `examples/maplibre/filter` ?

PR #2422, opened 2024-08-09
Update state-management.md

The-Lone-Druid:
The previous code example was misleading and it should be `viewState` instead of `newCenter`


chrisgervang (approved):
Nice catch. Thanks!

PR #2425, opened 2024-08-16
Docs: Include `properties` in FeatureCollection example

Johngeorgesample:
I tried copy/pasting the example for [Native Mapbox Layers](https://visgl.github.io/react-map-gl/docs/get-started/adding-custom-data#native-mapbox-layers), but encountered a TypeScript error.

```
[tsserver 2741] [E] Property 'properties' is missing in type '{ type: "Feature"; geometry: { type: "Point"; coordinates: number[]; }; }' but required in type 'Feature<Geometry, GeoJsonProperties>'.
```

This PR addresses that error by adding `properties` to the object.

### DIFF
--- a/docs/get-started/adding-custom-data.md
+++ b/docs/get-started/adding-custom-data.md
@@ -40,8 +40,7 @@ function App() {
         longitude: -122.45,
         latitude: 37.78,
         zoom: 14
-      }}
-    >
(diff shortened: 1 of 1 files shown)


chrisgervang (approved):
Thank you!

PR #2452, opened 2024-12-19
Fix text wrap issue for mobile devices in React Map GL

### LINKED ISSUE #2419: [Enhancement] Improv - Website hero text breaks on mobile
### Target Use Case

This would improve the experience of the React Map GL website for mobile users, and it should be a relatively easy fix.

Here is an example of what I am seeing:

Screenshot 2024-07-31 at 6 37 03 AM

### Proposal

N/A

RoseLV:
Closes #2419

This PR fixes the text wrapping issue on mobile devices like Samsung Galaxy S8+ 
by setting `word-wrap: normal` for the title element. This ensures proper rendering
of "React Map GL" without breaking the word "React".

### DIFF
--- a/website/src/components/home/styled.js
+++ b/website/src/components/home/styled.js
@@ -53,6 +53,7 @@ export const ProjectName = styled.h1`
   font-weight: 700;
   margin: 0;
   margin-bottom: 16px;
+  word-wrap: normal;
 `;
 
 export const GetStartedLink = styled.a`


Pessimistress:
Thank you for proposing the fix! I've opened https://github.com/visgl/dev-tools/pull/8 to update the shared component for all vis.gl websites.

PR #2472, opened 2025-01-24
[v8] Website

Pessimistress:
Restore website


website/docusaurus.config.js
```diff
@@ -1,199 +1,39 @@
...
...undation</a> and vis.gl contributors. All rights reserved. The <a href="https://openjsf.org">OpenJS Foundation</a> has registered trademarks and uses trademarks.  For a list of trademarks of the <a href="https://openjsf.org">OpenJS Foundation</a>, please see our <a href="https://trademark-policy.openjsf.org">Trademark Policy</a> and <a href="https://trademark-list.openjsf.org">Trademark List</a>.  Trademarks and logos not indicated on the <a href="https://trademark-list.openjsf.org">list of OpenJS Foundation trademarks</a> are trademarks&trade; or registered&reg; trademarks of their respective holders. Use of them does not imply any affiliation with or endorsement by them.</p><p><a href="https://openjsf.org">The OpenJS Foundation</a> | <a href="https://terms-of-use.openjsf.org">Terms of Use</a> | <a href="https://privacy-policy.openjsf.org">Privacy Policy</a> | <a href="https://bylaws.openjsf.org">Bylaws</a> | <a href="https://code-of-conduct.openjsf.org">Code of Conduct</a> | <a href="https://trademark-policy.openjsf.org">Trademark Policy</a> | <a href="https://trademark-list.openjsf.org">Trademark List</a> | <a href="https://www.linuxfoundation.org/cookies">Cookie Policy</a></p>'
```
chrisgervang, line 190: Is this included by default now?

website/package.json
```diff
@@ -1,38 +1,40 @@
...
+    "d3-scale": "^4.0.2",
+    "immutable": "^4.0.0",
+    "mapbox-gl": "^3.9.0",
+    "maplibre-gl": "^5.0.0",
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0"
   },
-  "volta": {
-    "node": "18.20.5",
-    "yarn": "1.22.22"
```
chrisgervang, line 36: Will this be the same as the root now? Volta was super helpful when switching between yarn 1 and 4, but even better if it can all be one version of yarn now

website/src/examples/mapbox/clusters.mdx
```diff
@@ -0,0 +1,5 @@
+# Clusters
+
+import App from 'mapbox-examples/clusters/src/app';
+
+<App />
```
chrisgervang, line 5: Is this compiled into to a running example?

website/src/mapbox-gl.css
chrisgervang, line 1: What do you think the process of keeping these up to date would look?

website/write-heading-ids.ts
chrisgervang, line 1: I'm assuming this is built-in now?

chrisgervang (approved):

website/write-heading-ids.ts
Pessimistress: Yes.

website/src/examples/mapbox/clusters.mdx
Pessimistress: Yes, this is the example page

website/docusaurus.config.js
Pessimistress: Good catch, no. I will update ocular to allow customization.

website/src/mapbox-gl.css
Pessimistress: When we bump the map libraries we will need to update them. I'm not too worried because they don't tend to change much within a major release...

chrisgervang: Gotcha, I think we'll thank ourselves later by including some contributing notes around how this can get updated

PR #2474, opened 2025-01-25
[v8] Align type names with base map libraries

Pessimistress:
See upgrade-guide.md

Docs will be updated in a separate PR.

### DIFF
--- a/modules/main/src/mapbox-legacy/types/style-spec.ts
+++ b/modules/main/src/mapbox-legacy/types/style-spec.ts
@@ -1,77 +1,34 @@
 /*
  * Mapbox Style Specification types
  */
-// Layers
-import type {
-  BackgroundLayer,
-  SkyLayer,
-  CircleLayer,
-  FillLayer,
-  FillExtrusionLayer,
-  HeatmapLayer,
-  HillshadeLayer,
-  LineLayer,
-  RasterLayer,
-  SymbolLayer
-} from 'mapbox-gl';
-
-export type AnyLayer =
-  | BackgroundLayer
-  | CircleLayer
-  | FillExtrusionLayer
-  | FillLayer
-  | HeatmapLayer
-  | HillshadeLayer
-  | LineLayer
-  | RasterLayer
-  | SymbolLayer
-  | SkyLayer;
-
 export type {
-  BackgroundLayer,
-  SkyLayer,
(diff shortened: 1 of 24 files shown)


modules/main/src/mapbox-legacy/types/style-spec.ts
```diff
@@ -1,77 +1,34 @@
...
-  | GeoJSONSource
-  | VideoSource
-  | ImageSource
-  | CanvasSource
-  | VectorSource
-  | RasterSource
-  | RasterDemSource;
+  // Sources
+  AnySourceData as SourceSpecification,
+  CanvasSourceRaw as CanvasSourceSpecification,
```
chrisgervang, line 20: Can you help me understand why we're aliasing some of these specifications?

modules/react-mapbox/src/types/internal.ts
```diff
@@ -1,34 +1,15 @@
...
+import type {Map} from 'mapbox-gl';
+
+export type Transform = Map['transform'];
+
+export type {
   GeoJSONSource as GeoJSONSourceImplementation,
-  ImageSource as ImageSourceImplemtation,
-  CanvasSource as CanvasSourceImplemtation,
+  ImageSource as ImageSourceImplementation,
+  CanvasSource as CanvasSourceImplementation,
```
chrisgervang, line 9: And then are these "Implementation" types the official name? 

Are they private? (I don't see them mentioned in the upgrade guide)

chrisgervang: Oh, I see these are in the internal type file

chrisgervang (approved):
Ok, I think I get it. This is aliasing mapbox-legacy to the current mapbox.

Is it also mapping maplibre spec to current mapbox spec, or are they aligning types eventually?

modules/main/src/mapbox-legacy/types/style-spec.ts
Pessimistress: This will match type names from `@types/mapbox-gl` (unofficial types) with the latest `mapbox-gl` official type names.

Pessimistress:
Mapbox v3 and Maplibre v5 type naming conventions are mostly the same, aside from actual API difference, and some niche exceptions.

PR #2477, opened 2025-01-28
Update example configs

Pessimistress:
Testing v8 alpha with examples.

### DIFF
--- a/examples/mapbox/clusters/tsconfig.json
+++ b/examples/mapbox/clusters/tsconfig.json
@@ -4,7 +4,7 @@
     "jsx": "react",
     "allowSyntheticDefaultImports": true,
     "resolveJsonModule": true,
-    "moduleResolution": "node",
+    "moduleResolution": "bundler",
     "module": "ES2020",
     "sourceMap": true
   }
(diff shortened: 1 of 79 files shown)


examples/mapbox/clusters/tsconfig.json
```diff
@@ -4,7 +4,7 @@
     "jsx": "react",
     "allowSyntheticDefaultImports": true,
     "resolveJsonModule": true,
-    "moduleResolution": "node",
+    "moduleResolution": "bundler",
```
Pessimistress, line 7: @chrisgervang this is a fairly unfamiliar issue for me - 

We re-export everything from submodules as follows:

```ts
// modules/main/src/mapbox.ts
export * from '@vis.gl/react-mapbox';
```

and

```json
// modules/main/package.json
"exports": {
    "./mapbox": {
      "types": "./dist/mapbox.d.ts",
      "require": "./dist/mapbox.cjs",
      "import": "./dist/mapbox.js"
    },
```

In the application tsconfig,

```json
`"moduleResolution": "node"`
```

leads to

```ts
import {Map} from '@vis.gl/react-mapbox' // no error
//--- but ---
import {Map} from 'react-map-gl/mapbox' // fails to resolve
```

chrisgervang: Weird, I see this is in an example.. is this something that end users would be concerned with?

PR #2480, opened 2025-01-30
feat: remove `RTLTextPlugin` default for MapLibre

KiwiKilian:
As discussed previously, I've removed the default value for the `RTLTextPlugin` which loaded from mapbox.com. This aligns more closely to the default behavior of the maplib, forcing the user to manually set such url. The prop now also accepts an object, to allow full configuration of the `setRTLTextPlugin` call.

This implements the proper fix towards https://github.com/visgl/react-map-gl/issues/2310

BREAKING CHANGE: `RTLTextPlugin` must now be set manually as prop

### DIFF
--- a/docs/upgrade-guide.md
+++ b/docs/upgrade-guide.md
@@ -22,7 +22,7 @@
 
 #### Removed default for `RTLTextPlugin`
 
-The default `RTLTextPlugin` loaded from mapbox.com has been removed for MapBox.
+The default `RTLTextPlugin` loaded from mapbox.com has been removed to align with the default behavior of MapLibre.
 To keep the previous behavior, specify the `pluginUrl` which was previously used or supply the plugin from any other source:
 
 ```tsx
--- a/docs/upgrade-guide.md
+++ b/docs/upgrade-guide.md
@@ -26,12 +26,7 @@ The default `RTLTextPlugin` loaded from mapbox.com has been removed to align wit
 To keep the previous behavior, specify the `pluginUrl` which was previously used or supply the plugin from any other source:
 
 ```tsx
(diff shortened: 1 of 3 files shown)


docs/upgrade-guide.md
```diff
@@ -18,6 +18,22 @@
   | `*Layer` | `*LayerSpecification` |
   | `*SourceRaw` | `*SourceSpecification` |
 
+### MapLibre
+
+#### Removed default for `RTLTextPlugin`
+
+The default `RTLTextPlugin` loaded from mapbox.com has been removed for MapBox.
```
Pessimistress, line 25:
```suggestion
The default `RTLTextPlugin` loaded from mapbox.com has been removed to align with the default behavior of MapLibre.
```

modules/react-maplibre/src/utils/set-globals.ts
```diff
@@ -4,7 +4,7 @@ export type GlobalSettings = {
    */
   maxParallelImageRequests?: number;
   /** The map's RTL text plugin. Necessary for supporting the Arabic and Hebrew languages, which are written right-to-left.  */
-  RTLTextPlugin?: string | false;
+  RTLTextPlugin?: {pluginUrl: string; lazy?: boolean};
```
Pessimistress, line 7: Not a blocker, but it shouldn't be hard to allow a string value as well?

Pessimistress (approved):

modules/react-maplibre/src/utils/set-globals.ts
KiwiKilian: You are right, reimplemented!

Issue #2506, opened 2025-03-18
[Bug] v8.x does not display Mapbox road labels when using Standard Styles


newguy123-creator:
### Description

When using any of the new Standard Mapbox mapstyles or varients of it since Mapbox 3.x, then the road labels do not display or render with React-Map-Gl. This was also true for V7.0
The OLD classic styles display fine and road labels etc work fine. Its the new Standard styles specifically I'm talking about. This also affects any new styles you create in Mapbox Studio based on the new Standard Styles. In Mapbox Studio I can clearly see everything as you expect, but then in the web app itself, no road labels

### Expected Behavior

Expect the road labels to render in all maps, not just when using classic maps, but also standard styles.

### Steps to Reproduce

Simply use the Standard Style from Mapbox 3.x, not the older classic styles like streets v12 etc.

### Environment

- Framework version: react-map-gl@8.0.1
- Map library: mapbox-gl 3.10.0
- Browser: Any (Chrome, Firefox, etc)
- OS: Any (Windows 10, MacOS)

### Logs

No errors, street labels just dont render when using standard mapbox style

Pessimistress:
This should be working with 8.1.0-alpha.1

Issue #2194, opened 2023-06-02
[Bug] `map.flyTo` centers on the incorrect spot when terrain is enabled.


erichcarson:
### Description

I recently tried to upgrade our version of `react-map-gl` to `v7.0.25`.  During the upgrade I encountered a bug where `map.flyTo` goes to the incorrect location when `map.setTerrain()` has been called.  Once at the incorrect location, if I manually move the map around, the `Marker` at that location stutters between the actual location and several incorrect locations.

Video with terrain and incorrect behavior:
https://github.com/visgl/react-map-gl/assets/127768277/fea86023-6558-457c-898a-eff71924333d

Video without terrain and correct behavior:
https://github.com/visgl/react-map-gl/assets/127768277/9f7d02b3-537a-49b7-a613-50fcfa17d97b

### Expected Behavior

With terrain enabled, I would expect the same behavior as without terrain enabled.  `flyTo` should center the map view on the actual location of the `Marker` taking into account the terrain information.  This was working when using `react-map-gl` `v6.1.10`.

### Steps to Reproduce

Behavior is reproducible at this [sandbox](https://codesandbox.io/s/flyto-terrain-example-dhlcq9?file=/src/index.js), although you will need your own Mapbox API token.

Comment out the following line (line 21) to toggle the behavior:
```
map.setTerrain({ source: "mapbox-dem", exaggeration: 1.5 });
```

### Environment

- Framework version: react-map-gl@7.0.25
- Map library: mapbox-gl@2.1.1
- Browser: Chrome Version 112.0.5615.137 
- OS: MacOS

### Logs

No error logs present in the console.

loganbrunner:
Is there any update on this? I am also experiencing a similar issue. When the terrain is enabled, my popup is not rendering where expected (both on a hillside as well as on a flat area).

See this screenshot, where I would expect the popup to only appear when my cursor is over the blue feature, yet it appears even when my cursor is just outside of the polygon. This does not occur if I don't include terrain on my map.
![Screenshot 2023-08-03 at 3 43 18 PM](https://github.com/visgl/react-map-gl/assets/14346184/abd389c7-eaa4-45d0-806a-a44353cb7075)

I _think_ this has something to do with the order in which the source layer is added and the terrain source is set. Mapbox GL JS seems to add the source layer and then set the terrain (code taken from [this Mapbox example](https://docs.mapbox.com/mapbox-gl-js/example/add-terrain/)):
```
map.on('style.load', () => {
[...]

Issue #2088, opened 2022-12-18
[Bug] When 3D Terrain is enabled, updating state in React causes Markers to render in their original locations, instead of on top of the terrain.


joshkautz:
### Description

When I add Markers to a Map and have 3D Terrain enabled, I noticed that changing state in React causes the markers to render in their original location on the map, rather than their modified 3D Terrain location. I'm not sure the technical words to use for this.

I've created a Code Sandbox as basic as possible to show this behavior. It sounds like this is the preferred way of sharing/showing potential bugs.

I couldn't find any issues that were related to this, but perhaps [this closed issue](https://github.com/visgl/react-map-gl/issues/1619) is along the same vein? 

I'd be happy to help in however I can, but I'll wait to hear back! Thank you!

### Expected Behavior

Changing state in React will still keep Markers in their existing position on the 3D Terrain.

### Steps to Reproduce

[https://codesandbox.io/s/vibrant-khayyam-n9lo2d](https://codesandbox.io/s/vibrant-khayyam-n9lo2d)

I didn't use the boiler plate template, because it's configured for a very old version of React. I just used the most up-to-date _Create React App_ template.

### Environment

- Framework version: react-map-gl@7.0.20
- Map library: mapbox-gl@2.11.1
- Browser: Version 108.0.5359.125 (Official Build) (64-bit)
- OS: Windows 10 Pro

joshkautz:
This does **not** happen when I use the default `<Marker>`!

If I replace ...
```
<Marker
   key={coordinate[0] + coordinate[1]}
   longitude={coordinate[0]}
   latitude={coordinate[1]}
   anchor="center"
>
   <img
      alt=""
      src="https://hikers.social/icons/mapbox-marker-icon-20px-red.png"
   />
</Marker>
```
... with ...
```
<Marker
   key={coordinate[0] + coordinate[1]}
   longitude={coordinate[0]}
   latitude={coordinate[1]}
   anchor="center"
/>
```
... then I don't have this issue. Maybe that will be helpful in understanding what's going on under the hood. 🤷

joshkautz:
Upon further investigation, I'm almost certain that this issue lies within mapbox-gl itself, so [I've reported the bug](https://github.com/mapbox/mapbox-gl-js/issues/12483) with greater detail to the Mapbox team. I noticed that using the "globe" projection type was introducing a lot of inconsistency with Layers of Symbols, as well as with Markers; these issues weren't happening when using the default projection type.

[...]

PR #2514, opened 2025-03-28
feat(mapbox): Replace shadow transform with proxied approach

Pessimistress:
This is a pretty big change to how we handle controlled Mapbox instance.

## Background 

The basis of our approach is that we keep two Transform instances: one that Map can mutate freely, and one that is controlled by React props.

In 7.x and 8.0, we created a `renderTransform` as the controlled instance. The mutable instance is the permanent source of truth of the map state. Its changes are merged with React props and synchronized back to `renderTransform`. During 1. render, 2. query and 3. event callbacks, `renderTransform` is swapped in so the map reads the controlled matrices.

We have made a lot of hacks and patches around this approach, and there still are a handful of long-standing issues, mostly around terrain and non-mercator projection. The source of the grievance is that
1. Mapbox can mutate `map.transform` on various lifecycle stages and some of them do not trigger an event.
2. Transform cannot be deterministically reproduced by its public methods and from a ViewState object. Terrain mutates unexposed private fields that affect matrix calculation (#2194, #2088, #2248). New features can also fail to be synchronized (#2506).

## Changes

This PR proposes a (hopefully) cleaner approach. We replace `map.transform` with a `ProxyTransform`, which uses [Proxy](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Proxy) to trap get and set operations. Under the hood, it manages the two Transform instances (freely mutable `proposedTransform` and React-controlled `controlledTransform`). Of the two, `controlledTransform` is the permanent source of truth, and `proposedTransform` is only temporarily populated IF the map is controlled AND during a move session. This results in much better consistency in map behavior:
1. In the uncontrolled use case (using `initialViewState`) there is only one transform instance, which should minimize the perf/bugs impact from our wrapper
2. No more instance swapping, which should eliminate the discrepancy in behavior between GL renderer, DOM controls and query.

I am also trying to document inline extensively to improve code readability.

## Release plan

Targeting a v8.1 release.
I am leaning towards NOT back porting this to `mapbox-legacy`, which is mainly used for mapbox-gl@1 and will be eventually deprecated. The old approach works fairly well for the basic (no terrain or alt projection) use case. The new approach relies heavily on the internals of the `Transform` class and the official types. I did a quick audit and most of the private methods have existed since v2.0. However we do not test with non-latest mapbox versions and it could become a maintenance hazard.

### DIFF
--- a/modules/react-mapbox/src/types/common.ts
+++ b/modules/react-mapbox/src/types/common.ts
@@ -27,6 +27,8 @@ export type ViewState = {
   pitch: number;
   /** Dimensions in pixels applied on each side of the viewport for shifting the vanishing point. */
   padding: PaddingOptions;
+  /** Center elevation on terrain */
+  altitude?: number;
 };
 
 export interface ImmutableLike<T> {
--- a/modules/react-mapbox/src/types/common.ts
+++ b/modules/react-mapbox/src/types/common.ts
@@ -28,7 +28,7 @@ export type ViewState = {
   /** Dimensions in pixels applied on each side of the viewport for shifting the vanishing point. */
   padding: PaddingOptions;
   /** Center elevation on terrain */
-  altitude?: number;
+  elevation?: number;
 };
 
 export interface ImmutableLike<T> {
--- a/docs/api-reference/mapbox/types.md
+++ b/docs/api-reference/mapbox/types.md
@@ -168,6 +168,7 @@ An object with the following fields:
 - `zoom`: number - The zoom level.
 - `pitch`: number - The pitch (tilt) of the map, in degrees.
 - `bearing`: number - The bearing (rotation) of the map, in degrees.
+- `elevation`: number|undefined - The map center elevation from sea leavel on terrain surface, if any
 
 
 ## Events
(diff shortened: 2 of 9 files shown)


modules/react-mapbox/src/types/common.ts
```diff
@@ -27,6 +27,8 @@ export type ViewState = {
   pitch: number;
   /** Dimensions in pixels applied on each side of the viewport for shifting the vanishing point. */
   padding: PaddingOptions;
+  /** Center elevation on terrain */
+  altitude?: number;
```
chrisgervang, line 31: Could we call this elevation since it represents the ground level? Altitude can be reserved for camera position up in the air.

chrisgervang (approved):
Thanks for tackling this! I think mapbox users will be happy to have terrain supported. The proxy approach looks a lot more maintainable as well.. much better than needing to reserve engineer all of the ways the camera gets transformed!

Could we populate `elevation` in maplibre's viewState as well?

Also, do you want to add `elevation` to the docs?

Issue #2520, opened 2025-04-02
[Bug] When passing multiple Sources in, each with their own Layer, only one additional Layer is attached each time the Map updates


carlton-l-cassedy:
### Description

I have an array of data objects, which I call `map()` to JSX inside my `<Map />` component. The JSX consists of some custom components which wrap `react-map-gl` Source and Layer components. For the sake of reproducing the error, I've simplified my actual use case down, and so it may seem odd that I would choose to use this approach rather than Markers, but it is just for the discussion.

When I pass in an array to the Map of these <Source><Layer /></Source> based components, only the 0th index actually appears on the map. Using React Dev tools extension, I can see that all of the Sources are mounted, but only the Layer for the first one is present in the hierarchy. 

The App component has a state value count, which can be incremented with a button. Of course when the state updates, the app (and therefore the Map component and data components) re-render. If I cause a re-render by incrementing the counter, the second Layer becomes present, and visible on the map. Continuing to increment the counter adds an additional Layer each time.

If I put the array inside the App render body as a useMemo with empty dependency array, the same behavior occurs, except that incrementing the counter has no effect. However, if I add count as a dependency, the original behavior is fully matched. This is despite the fact that count should clearly not be a dependency for the components being displayed on the map (and my linter correctly recognizes it as an unnecessary dependency).

Things are slightly different in the CodeSandbox I'm providing. I included another button to swap to use Markers instead of the custom Source+Layer, and these have the correct behavior, where they are all present immediately, regardless of the counter state or memoization. Additionally, in the CodeSandbox, instead of adding 1 layer each render, it adds 2. I am not sure why this difference occurs. I was not able to make it only do 1 in CodeSandbox, or do 2 in my local environment. But of course there is still an issue here regardless.

### Expected Behavior

I expect to see all of the map items for the data array on the map on the first render, as happens when using Marker.

### Steps to Reproduce

You will need to fork the sandbox and put your mapbox token in. I started from the sandbox boilerplate which does not involve a token due to being an older version of the libraries. However, the issue did not occur on those versions. Nor does it occur (at least with my testing so far) with the latest maplibre version of react-map-gl. So it is necessary to use mapbox and a token to test the issue.

https://codesandbox.io/p/sandbox/react-map-gl-issue-2520-xtdkt2

### Environment

- Framework version: 8.0.2
- Map library: mapbox-gl@3.11.0
- Browser: Chrome, Edge
- OS: Windows 11

### Logs

[...]

PR #2493, opened 2025-02-21
Fix style loading check on Source layers (#2377)

Inpacchi:
Closes #2377 

Fixed the `Style is not done loading` error by using a map.isStyleLoaded() function check in place of the primitive values.

PR #2524, opened 2025-04-08
Fix addSource condition

Pessimistress:
For #2520

Fallout from #2493 - see inline comments

### DIFF
--- a/modules/react-mapbox/src/components/source.ts
+++ b/modules/react-mapbox/src/components/source.ts
@@ -1,5 +1,5 @@
 import * as React from 'react';
-import {useContext, useEffect, useMemo, useState, useRef, cloneElement} from 'react';
+import {useContext, useEffect, useMemo, useCallback, useState, useRef, cloneElement} from 'react';
 import {MapContext} from './map';
 import assert from '../utils/assert';
 import {deepEqual} from '../utils/deep-equal';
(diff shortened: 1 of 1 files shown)


modules/react-mapbox/src/components/source.ts
```diff
@@ -125,6 +124,20 @@ export function Source(props: SourceProps) {
...
   propsRef.current = props;
 
+  useEffect(() => {
+    if (!source) {
+      // on `styledata` event, `map.isStyleLoaded()` still returns false.
+      // `load` and `style.load` only fire once and not when `isStyleLoaded` changes from true to false to true.
+      // `sourcedata` potentially suggests that `isStyleLoaded` has changed. But it fires on every tile load.
+      // Unsubscribe once source is added.
+      map.on('sourcedata', forceUpdate);
+      return () => {
```
chrisgervang, line 134: These three events are not very intuitive. Not much we can do

modules/react-mapbox/src/components/source.ts
```diff
@@ -83,17 +83,16 @@ export function Source(props: SourceProps) {
...
 
   useEffect(() => {
     if (map) {
-      /* global setTimeout */
-      const forceUpdate = () => setTimeout(() => setStyleLoaded(version => version + 1), 0);
-      map.on('load', forceUpdate);
+      // Fired on initial load signaling the map is ready to add custom sources
+      // Subsequently fired on style changes
       map.on('styledata', forceUpdate);
       forceUpdate();
```
chrisgervang, line 94: Safe to remove since styledata fires at the same time?

chrisgervang (approved):

modules/react-mapbox/src/components/source.ts
Pessimistress: `styledata` is the "correct" event to observe for map style changes. `sourcedata` fires much more frequently (e.g. on every new tile load) and is excessive once source is added.

In this PR, we subscribe to both `styledata` and `sourcedata` while the map is loading - either event could lead to `map.isStyleLoaded()` become true - once map is loaded we add the source, and unsubscribe from `sourcedata`. If the user swaps out the map stylesheet, `styledata` will fire and `sourcedata` is added again.

Issue #2523, opened 2025-04-08
[Bug] Possible regression in v8.0.2 — Layers and sources not rendering consistently


MartinSandstrom:
### Description

Hi there,

Thank you for maintaining this great library :) 

After upgrading from `react-map-gl@8.0.1 to 8.0.2`, we’ve noticed some unexpected behavior: Certain sources and layers are not rendering reliably. In some cases, layers don’t appear, and toggling visibility on one layer can cause others (previously missing) to appear.

So far, the behavior seems inconsistent and difficult to pin down. We’ve reviewed the patch notes and commits but haven’t found any obvious changes that would explain this issue. We've also tried reproducing the problem in a minimal sandbox environment, but haven’t had any luck, possibly because we’re not yet sure what specifically is triggering it.

We have tried to lo the layers that exist in the map with `map?.getStyle()?.layers;` and the number of layers is different for more or less each render. 

**What we’ve observed:**
- Some layers fail to render initially
- Toggling one layer may cause others to appear or disappear (different renders have different layers)
- Reverting to 8.0.1 restores the expected behavior

Do you have any clue as to what might be causing this behavior in the latest release?

### Expected Behavior

The layers should be rendered at all times and stay consistent between re-renders. 

### Steps to Reproduce

At the point of writing the only way I have found to reproduce this is to upgrade the version in one of our private repositories. Sorry for this.

### Environment

- Framework version: 8.0.2
- Map library: 3.9.4
- Browser: Chrome
- OS: macOS

marcus-pervanovo:
Thanks! I just had a similar issue (one layer did not render initially, unless the map was zoomed out below level 6).
Downgrading to v8.0.1 solved the issue.

Pessimistress:
Should be fixed by 8.0.3

MartinSandstrom:
Yes, the ID is from our system. The user can navigate backward and forward in time, and each day corresponds to a specific layer with a unique ID, like the one mentioned above. Each time the user changes the day, we render another raster. The errors are inconsistent with the time changes, sometimes there is no error, and sometimes there are multiple.

In one of our other applications, we’ve seen related errors being logged, such as:
`Error: Layer with id "0193f7f9-c327-7eb3-8412-526ffb44624c" does not exist on this map.`

And sometimes also:
`Error: Source "jsx-source-58" cannot be removed while layer "0193f7f9-c327-7eb3-8412-526ffb44624c" is using it.`

[...]

Issue #2377, opened 2024-04-15
[Feat] Add support for latest Mapbox rendering release (V3)


trumbitta:
### Target Use Case

The simplest basic case with Mapbox v3 works, but as soon as you add a `Source` and `Layer` we get the `Style is not done loading` error.

Here's a repro: https://codesandbox.io/p/sandbox/frosty-minsky-pdqrlk?file=%2Fsrc%2FApp.js

zsloan112:
Glad I checked here today. I was getting this same error and had been fighting it for hours. 
Same use case and reproduction steps for me.

iostat:
As a workaround, you can wrap everything inside your `<Map...>` with a component that listens for the `style.load` and only populates children of the `<Map>` once that event has been fired.

It can get a little hairy if you start doing stuff like changing the map style dynamically (totally doable, but hairy!), but something like the following can get you started.
```tsx
import {useEffect, useState} from 'react';
import {default as ReactMapGL, useMap} from 'react-map-gl';

type StyleLoadedGuardProps = {
  // this state has to come from outside the StyleLoadedGuard component as otherwise
  // it'll get cleared if/when the map changes. And the guard has to be it's own component
  // as it seems like that's the only way to get a ref to the map via useMap()...
  guardState: [boolean, React.Dispatch<React.SetStateAction<boolean>>]; // useState(false)
  children?: React.ReactNode;
};
const StyleLoadedGuard: React.FC<StyleLoadedGuardProps> = props => {
  const mapRef = useMap();
  const [styleLoadedAtLeastOnce, setStyleLoadedAtLeastOnce] = props.guardState;
  useEffect(() => {
    if (mapRef.current) {
      const map = mapRef.current;

      const onStyleLoad = () => {
        setStyleLoadedAtLeastOnce(true);
      };
      map.on('style.load', onStyleLoad);
      if (map.isStyleLoaded()) {
        onStyleLoad();
      }
      return () => {
        map.off('style.load', onStyleLoad);
      };
    } else {
      return undefined;
    }
  }, [mapRef]);

  return styleLoadedAtLeastOnce && props.children;
}

export const MyMap: React.FC<ReactMapGL.MapProps> = (props) => {
  const styleLoadedGuardState = useState(false);
  const mapProps = {...props, children: undefined}; // to be safe
  return <ReactMapGL.Map {...props}>
    <StyleLoadedGuard guardState={styleLoadedGuardState}>
      {props.children}
[...]

PR #2526, opened 2025-04-14
Revert changes to add source condition

Pessimistress:
For #2523

Reverts the collection of commits downstream from #2493 until we are able to reproduce #2377

### DIFF
--- a/examples/vite.config.local.js
+++ b/examples/vite.config.local.js
@@ -9,8 +9,8 @@ export default defineConfig(async () => {
     resolve: {
       alias: {
         // Use root dependencies
-        'react-map-gl/mapbox': join(rootDir, './modules/react-mapbox/src'),
-        'react-map-gl/maplibre': join(rootDir, './modules/react-maplibre/src'),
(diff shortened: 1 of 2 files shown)


examples/vite.config.local.js
```diff
@@ -9,8 +9,8 @@ export default defineConfig(async () => {
     resolve: {
       alias: {
         // Use root dependencies
-        'react-map-gl/mapbox': join(rootDir, './modules/react-mapbox/src'),
-        'react-map-gl/maplibre': join(rootDir, './modules/react-maplibre/src'),
+        'react-map-gl/mapbox': join(rootDir, './modules/main/src/mapbox.ts'),
+        'react-map-gl/maplibre': join(rootDir, './modules/main/src/maplibre.ts'),
```
chrisgervang, line 13: Meant to be included?

chrisgervang (approved):
Thanks for continuing to iterate on this issue!

examples/vite.config.local.js
Pessimistress: Good catch

PR #2535, opened 2025-05-20
fix: max depth exceeded error when dynamically changing map settings

hwandrew:
stems from an issue found in https://github.com/visgl/react-map-gl/issues/2194#issuecomment-2878158442

When we check for if settings changed a new proxy transform is instantiated, but a previously instantiated proxy was already created and assigned to the argument. This call just created a proxy of the proxy (and kept doing so, hence the stack error) - I think we could just keep the original proxy object without reinstantiating.

Added a spec to dynamically update map settings after initial render to catch this.

Before:

https://github.com/user-attachments/assets/14a76a5d-7b3d-44e3-b4de-f4e695c47521 

After:

https://github.com/user-attachments/assets/996fe4a8-1dcb-44e2-b256-5f19502a9345

### DIFF
--- a/modules/react-mapbox/src/mapbox/mapbox.ts
+++ b/modules/react-mapbox/src/mapbox/mapbox.ts
@@ -216,9 +216,6 @@ export default class Mapbox {
     this.props = props;
 
     const settingsChanged = this._updateSettings(props, oldProps);
-    if (settingsChanged) {
-      this._createProxyTransform(this._map);
-    }
     const sizeChanged = this._updateSize(props);
     const viewStateChanged = this._updateViewState(props, true);
     this._updateStyle(props, oldProps);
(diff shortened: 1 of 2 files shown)


modules/react-mapbox/src/mapbox/mapbox.ts
```diff
@@ -216,9 +216,6 @@ export default class Mapbox {
     this.props = props;
 
     const settingsChanged = this._updateSettings(props, oldProps);
-    if (settingsChanged) {
-      this._createProxyTransform(this._map);
```
chrisgervang: @Pessimistress do you know why we had this?

hwandrew:
thanks for looking! @Pessimistress looks like I don't have permission to merge, so feel free to merge. Lmk if the copilot suggestions need to be addressed, but it clones the existing patterns used in that spec.

PR #2560, opened 2025-09-30
Invoke setter with default when prop missing

felixpalmer:
Closes https://github.com/visgl/react-map-gl/issues/2559

### DIFF
--- a/modules/react-maplibre/src/maplibre/maplibre.ts
+++ b/modules/react-maplibre/src/maplibre/maplibre.ts
@@ -80,6 +80,16 @@ export type MaplibreProps = Partial<ViewState> &
 
 const DEFAULT_STYLE = {version: 8, sources: {}, layers: []} as StyleSpecification;
 
+const DEFAULT_SETTINGS = {
+  minZoom: 0,
+  maxZoom: 22,
+  minPitch: 0,
+  maxPitch: 85,
+  maxBounds: [-180, -85.051129, 180, 85.051129],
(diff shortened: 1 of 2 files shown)


modules/react-maplibre/src/maplibre/maplibre.ts
```diff
@@ -80,6 +80,16 @@ export type MaplibreProps = Partial<ViewState> &
 
 const DEFAULT_STYLE = {version: 8, sources: {}, layers: []} as StyleSpecification;
 
+const DEFAULT_SETTINGS = {
+  minZoom: 0,
+  maxZoom: 22,
+  minPitch: 0,
+  maxPitch: 85,
+  maxBounds: [-180, -85.051129, 180, 85.051129],
+  projection: 'mercator',
```
chrisgervang, line 89: Is projection in maplibre? Might want to cross reference https://www.maplibre.org/maplibre-gl-js/docs/API/type-aliases/MapOptions/

chrisgervang (approved):

modules/react-maplibre/src/maplibre/maplibre.ts
felixpalmer: Yes: https://maplibre.org/maplibre-style-spec/projection/

chrisgervang: That's a different API I think. That's the JSON style spec whereas this is the `MapOptions` for the js library

PR #2563, opened 2025-10-09
Fix Map component's child container style

joeriddles:
https://github.com/visgl/react-map-gl/pull/2132 introduced a change where there is always a div rendered below the map component with the same height of the map. This can result in an empty div that obstructs other elements rendered below the map.

This PR makes two changes:
1. The child container div is only rendered if there are any children
2. Adds a prop to allow styling the child container


Pessimistress:
If we add `display: contents` to this container, will it fix your problem?

PR #2573, opened 2025-12-23
Call setMinZoom, setMaxZoom, setMinPitch and setMaxPitch in the right order to avoid error in Maplibre

### LINKED ISSUE #2572: [Bug] Changing minZoom and maxZoom at the same time can throw an error in Maplibre
### Description

A recent change (https://github.com/maplibre/maplibre-gl-js/issues/6766) introduced into the way that `maplibre-gl` handles `setMinZoom` and `setMaxZoom` calls to prevent stale values means that these changes are applied immediately.

As such, a change in props from `{ minZoom: 1, maxZoom: 3 }` to `{ minZoom: 5, maxZoom: 10 }` can throw an error in Maplibre depending on the order in which these are updated. If `setMinZoom` is called first the new `minZoom: 5` is greater than the current `maxZoom: 3` and an error is thrown (https://github.com/maplibre/maplibre-gl-js/blob/main/src/ui/map.ts#L1154). Maybe there's an order to call these in that always works? Or we request an upstream function eg. `setZoomConstraint({ min: x, max: y })` that changes both in one go?

### Expected Behavior

_No response_

### Steps to Reproduce

- Create a map with `{ ..., minZoom: 1, maxZoom: 3 }` props
- Change props to `{ minZoom: 5, maxZoom: 10 }`
- Note error thrown

### Environment

- Framework version: react-map-gl@8.1.0
- Map library: maplibre-gl@5.15.0
- Browser: N/A
- OS: N/A

### Logs

_No response_

Auspicus:
Adds a test to cover different cases where `minZoom` and `maxZoom` are adjusted together. Ensures that `minZoom` and `maxZoom` never cross. The same has been done for `minPitch` and `maxPitch`.

fixes: https://github.com/visgl/react-map-gl/issues/2572

see also: https://github.com/maplibre/maplibre-gl-js/pull/6926 where a bug was fixed causing pitch to also throw an error if changed in the wrong order


Pessimistress (approved):

Pessimistress:
Can you run `yarn lint fix`?

Auspicus:
@Pessimistress I've done that but I'd like to clean this up a bit before we merge if that's OK?

Auspicus:
@Pessimistress I think this is in a better state now. Let me know if you have feedback :)

PR #2580, opened 2026-02-03
fix(react-maplibre): guard against undefined map.style in _updateStyleComponents

kevinjmo:
## Summary

When using React 19's `<Activity>` component with `mode="hidden"`, the map can be in a partially initialized state when event handlers fire. This causes a `TypeError` when accessing `map.style._loaded` because `map.style` is undefined.

**Error:**
```
TypeError: undefined is not an object (evaluating 'map.style._loaded')
```

**The fix:** Add optional chaining (`map.style?._loaded`) to prevent the crash while maintaining the same logical behavior - if `map.style` is undefined, the condition is falsy and the style components won't be updated (which is the correct behavior when the map isn't fully initialized).

## Changes

- `modules/react-maplibre/src/maplibre/maplibre.ts`: Line 465 - changed `map.style._loaded` to `map.style?._loaded`

## Context

React 19's Activity component with `mode="hidden"`:
- Runs cleanup effects (like unmounting)
- Preserves React state
- When switching back to `mode="visible"`, effects run again

During this transition, internal map events (`style.load` or `sourcedata`) can fire and call `_updateStyleComponents` before the map is fully reinitialized, causing the null reference error.

### DIFF
--- a/modules/react-maplibre/src/maplibre/maplibre.ts
+++ b/modules/react-maplibre/src/maplibre/maplibre.ts
@@ -462,7 +462,7 @@ export default class Maplibre {
     const map = this._map;
     const currProps = this._styleComponents;
     // We can safely manipulate map style once it's loaded
-    if (map.style._loaded) {
+    if (map.style?._loaded) {
       if (light && !deepEqual(light, currProps.light)) {
         currProps.light = light;
         map.setLight(light);


onewby:
I’ve been having issues with this, and some testing in an emulated old Chrome/Android version appears to show this fix prevents the issue. Would love to see it get merged.

kashik0i:
Confirming this bug from a different trigger — we hit the same `TypeError: Cannot read properties of null (reading '_loaded')` in Next.js 16 via Fast Refresh / HMR, not React 19's `<Activity>`. Same root cause, worth flagging for the maintainers since it affects a much broader dev workflow.

**Environment**
- `react-map-gl@8.1.0` → `@vis.gl/react-maplibre@8.1.0`
- `maplibre-gl@5.21.1`
- `next@16.2.2`, `react@19.2.4`
- Next.js 16 dev server with Turbopack + Fast Refresh

**Trigger**: Editing any file in our `<MapGL>` component subtree while the dev server is running. Fast Refresh causes `setProps` to fire, which synchronously calls `_updateStyle` (triggers `map.setStyle()`) followed immediately by `_updateStyleComponents` — the exact same race window you described, just opened by HMR instead of `<Activity>` cleanup.

**Stack trace:**
```
Uncaught TypeError: Cannot read properties of null (reading '_loaded')
    at Maplibre._updateStyleComponents (maplibre.js:375)
    at Maplibre.setProps (maplibre.js:140)
    at MapGL (...)
```

**Workaround we've applied**: `pnpm patch @vis.gl/react-maplibre@8.1.0` with the same single-line guard (we used `map.style && map.style._loaded` rather than optional chaining, but both are equivalent — your `?.` approach is arguably cleaner). Tracking the patch for removal once this PR lands.

Would love to see this merged — happy to provide a minimal Next.js reproduction repo if it would help unblock review.

Pessimistress (approved):

PR #2594, opened 2026-06-15
Add `react-map-gl/mapbox-esm` entrypoint

stepankuzmin:
Exposes Mapbox GL JS's new `mapbox-gl/esm` entrypoint.


stepankuzmin:
Looks like I missed the `mapLib` prop that should be enough to override the GL JS

```jsx
import Map from 'react-map-gl/mapbox';
import mapboxGl from 'mapbox-gl/esm';

<Map mapLib={mapboxGl} {...props} />
```

PR #2596, opened 2026-07-14
fix: unset marker style properties that become undefined (#2595)

### LINKED ISSUE #2595: [Bug] Prop `style` does not unset properties that become `undefined`
### Description

In the following snippet:
```ts
<MapGLMarker
        latitude={location.coordinates.latitude}
        longitude={location.coordinates.longitude}
        anchor="bottom"
        style={{
          ...(condition
            ? { background: 'red', }
            : { background: undefined, }), // Note: similar problem if just passing {}
        }}
      >
```

### Expected Behavior

When condition goes from `true` to `false`, the `background` attribute is un-set on the element.

### Steps to Reproduce

Reproduction: https://codesandbox.io/p/sandbox/long-rgb-6f4mdv

<details>
<summary>Code snippet</summary>

```tsx
import * as React from "react";
import { createRoot } from "react-dom/client";
import Map, { Marker } from "react-map-gl";

import "mapbox-gl/dist/mapbox-gl.css";

function Root() {
  const [state, setState] = React.useState(false);
  return (
    <>
      <button onClick={() => setState(!state)}>
        background: {state ? "on" : "off"}
      </button>
      <Map
        initialViewState={{
          latitude: 37.8,
          longitude: -122.4,
          zoom: 14,
        }}
        style={{ width: "100vw", height: "100vh" }}
        mapStyle="https://basemaps.cartocdn.com/gl/positron-gl-style/style.json"
      >
        <Marker
          longitude={-122.4}
          latitude={37.8}
          style={{
            ...(state ? { background: "red" } : { background: undefined }), // Note: similar problem if just passing {}
          }}
        />
      </Map>
    </>
  );
}

/* global document */
createRoot(document.getElementById("app")).render(<Root />);
```

</details>

### Environment

- Framework version: react-map-gl@8.1.1 (in my project; reproduction template used `^7.1.0`)
- Map library: mapbox-gl@3.24.0
- Browser: (not relevant)
- OS: (not relevant)

### Logs

No logging produced

MohammadYusif:
## Summary

- `<Marker style={...}>` did not unset a CSS property when it changed to `undefined` or was removed from the style object (e.g. `{background: 'red'}` -> `{background: undefined}` or `{}` left `background` set).
- Root cause: `applyReactStyle` only wrote the keys present in the incoming style object and never cleared previously-applied keys; assigning `undefined` to a style property is a no-op, so stale values lingered.

## Changes

- `modules/react-mapbox/src/utils/apply-react-style.ts` / `modules/react-maplibre/src/utils/apply-react-style.ts`: track applied keys per element via a `WeakMap`, skip `undefined`/`null` values, and clear any previously-applied key that is no longer present.
- `modules/react-mapbox/test/utils/apply-react-style.spec.js` / `modules/react-maplibre/test/utils/apply-react-style.spec.js`: add regression test covering the `{background: undefined}` and `{}` cases.

Fixes #2595

---

> [!NOTE]
> **Low Risk**
> Localized DOM styling helper change with tests; no auth, data, or API surface changes.
> 
> **Overview**
> Fixes stale inline styles on map overlays (e.g. `<Marker style={...}>`) when a property is cleared via `undefined`/`null` or dropped from the style object.
> 
> `applyReactStyle` in **react-mapbox** and **react-maplibre** now remembers which keys it applied per DOM node (`WeakMap`), skips `undefined`/`null` values, and resets any previously applied key that is no longer present by setting `style[key] = ''`. Regression tests cover the `background: undefined` and empty `{}` cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bbdc9ced44b3c8bd808feee48b8debdee7590846. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


chrisgervang:
Thanks for working on this! While reviewing, we found a few additional edge cases around CSS shorthand/longhand transitions and removing the style prop. In particular, removed properties need to be cleared before applying new values; otherwise transitions such as `padding` to `paddingTop` can erase the newly applied value.

We opened #2601 as an alternative implementation. It keeps the per-element `WeakMap` approach you've written here and adds mirrored Mapbox/MapLibre coverage for `undefined`, `null`, omitted properties, whole-style removal, reapplication, shorthand transitions, and preserving unmanaged inline styles. The bot’s “null style crashes” warning also appears to be a false positive under the project’s ES2020 target, though the alternative normalizes nullish styles explicitly.

Curious to hear your thoughts on this and whether there’s anything from your implementation we should carry over?

Thanks again for getting the fix started!

PR #2600, opened 2026-07-28
ci: test MapLibre v4/v5 and export GlobeControl

### LINKED ISSUE #2592: [Bug] GlobeControl was added in v8.1.1 but not exported from index
### Description

`GlobeControl` was introduced in the latest version but can't be imported. 
We need to add GlobeControl exports to these two places:

https://github.com/visgl/react-map-gl/blob/master/modules/react-mapbox/src/index.ts
https://github.com/visgl/react-map-gl/blob/master/modules/react-maplibre/src/index.ts

### Expected Behavior

_No response_

### Steps to Reproduce

Verify `GlobeControl` can't be imported:

```js
// this errors
import { GlobeControl } from "react-map-gl/maplibre";
```

### Environment

- Framework version: v8.1.1
- Map library: "maplibre-gl": "^5.24.0",
- Browser: chrome
- OS: macos

### Logs

_No response_

chrisgervang:
Adds a non-fail-fast CI matrix for MapLibre `^4.0.0` and `^5.0.0`, with Coveralls uploading only from v5. Makes GlobeControl typing version-aware so v4 remains supported, exports GlobeControl and its props publicly, and documents its v5+ requirement. Adds an export regression test and corrects the helper so it checks each endpoint. Validated build, lint, and node tests against MapLibre 4.7.1 and 5.0.1.

Fixes #2592

---

> [!NOTE]
> **Low Risk**
> Changes are CI matrix expansion, public exports, optional typing, and test fixes—no auth, security, or core map runtime behavior beyond explicit GlobeControl version checks.
> 
> **Overview**
> CI now runs lint and tests against **MapLibre `^4.0.0` and `^5.0.0`** in a non-fail-fast matrix; Coveralls uploads only from the v5 job, and a gate job fails the workflow if any matrix leg fails.
> 
> **GlobeControl** is exported from the public MapLibre entry (component + `GlobeControlProps`), with docs noting **maplibre-gl v5+**. Types treat `GlobeControl` on `MapLib` as optional and define `GlobeControlInstance` conditionally so v4 consumers still compile; runtime throws a clear error if the control class is missing.
> 
> Export regression tests now check each endpoint against the right component list (including Terrain, Logo, and Globe for MapLibre), and the helper compares the passed module instead of always using legacy exports.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4363759ff4b1cb5bc6472cba3be3496f84babe7a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


modules/react-maplibre/src/components/globe-control.ts
```diff
@@ -12,9 +12,18 @@ export type GlobeControlProps = {
...
 
 function _GlobeControl(props: GlobeControlProps) {
-  const ctrl = useControl(({mapLib}) => new mapLib.GlobeControl(props), {
-    position: props.position
-  });
+  const ctrl = useControl(
+    ({mapLib}) => {
+      const GlobeControl = mapLib.GlobeControl;
+      if (!GlobeControl) {
+        throw new Error('GlobeControl is not supported by this version of MapLibre GL JS');
```
chrisgervang, line 19: GlobeControl exists only in MapLibre v5+

modules/react-maplibre/src/types/lib.ts
```diff
@@ -44,10 +44,17 @@ export type {
   TerrainControl as TerrainControlInstance,
   LogoControl as LogoControlInstance,
   LogoControlOptions,
-  GlobeControl as GlobeControlInstance,
```
chrisgervang, line 47: Building against v4 produced three TypeScript errors: these two GlobeControl exports don't exist, and one incompatible MapLib assignment. The fix preserves v4 support while making GlobeControl explicitly optional and unavailable at runtime on versions that don’t provide it.

Wdyt, @Pessimistress? This was previously undetected since we don't run CI on v4 anymore

test/src/exports.ts
```diff
@@ -18,10 +18,20 @@ const Components = [
...
+  'GlobeControl'
+] as const;
+
+function getMissingExports(
+  module: any,
+  components = Components as readonly string[]
+): null | string[] {
   const missingExports: string[] = [];
-  for (const key of Components) {
-    if (!legacyComponents[key]) {
```
chrisgervang, line 24: this seems to be an old typo. the `module` param was not used. Also expanded this to cover Maplibre-specific components

Pessimistress (approved):

